PR #560, opened 2011-12-30
Droppable: Modified the ui.ddmanager.prepareOffsets and droppable._drop ...

zhizhangchen:
...to get droppable's offset related to draggable's Window. Fixed #7925 - droppable prepareOffsets sets incorrect offset when droppable is in an iframe

### DIFF
--- a/ui/jquery.ui.droppable.js
+++ b/ui/jquery.ui.droppable.js
@@ -120,7 +120,7 @@ $.widget("ui.droppable", {
 				&& !inst.options.disabled
 				&& inst.options.scope == draggable.options.scope
 				&& inst.accept.call(inst.element[0], (draggable.currentItem || draggable.element))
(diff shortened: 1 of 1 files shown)


kamelkev:
Hi,

I don't think the above statement is correct, specifically the part about how the offset() function has taken scrolling into account. iFrames are logically separate entities, jQuery does not introspect the elements or attributes of an IFrame without you directly telling it to do so.

I'm working on a plugin to bridge this gap until some better tools come along, details here:
https://github.com/kamelkev/jquery-ui-droppable-iframe

I implemented your solution and ran my tests against it, and as expected the droppable offsets are not quite right, which makes sense, as those numbers do need to be included.

I've gotten pretty far, but your approach is a looooot simpler than mine is. My solution also has a lot of trouble with floating elements, which seemingly yours will not. I'd be happy to integrate your solution if you can somehow take scrolling into account. It's not clear if that's possible, but there certainly is a need for a comprehensive plugin to assist with this problem.

PR #590, opened 2012-02-02
Datepicker: option for full month names in month select

svermillion:
An option to display the full month name in the select list (when changeMonth=true), which is more user friendly in some scenarios, such as wider datepickers.


ecastillo550:
Wow... I also think this would be good.

davidpustai:
:+1:

davidpustai:
You can work it around by extending the whole datepicker.
```
$.extend($.datepicker, {
	__base__generateMonthYearHeader: $.datepicker._generateMonthYearHeader,
	_generateMonthYearHeader: function (inst, drawMonth, drawYear, minDate, maxDate, secondary, monthNames, monthNamesShort) {
		return $.datepicker.__base__generateMonthYearHeader(inst, drawMonth, drawYear, minDate, maxDate, secondary, monthNames, monthNames);
	}
});
```

PR #879, opened 2013-01-06
Datepicker: focus on next element after selecting date. Fixed #7765 - Da...

schettino72:
...tepicker: Focus not placed back into textbox after date selection


KevinGitTest:
Any update on this one regarding the working version on IE 7-10 ?

PR #960, opened 2013-04-16
Fixed resizable containment bug.

harrycheung:
Resizable wasn't behaving correctly when the element is absolutely positioned. There were several typos and un-symmetric logic (ie, not treating horizontal and vertical the same). Also, resizable didn't deal well in a scrolling environment.

### DIFF
--- a/tests/unit/resizable/resizable.html
+++ b/tests/unit/resizable/resizable.html
@@ -105,12 +105,15 @@
 	<div id="lots_o_content"></div>
 	<div id="containment2"></div>
 	<div id="resizable3"></div>
+<<<<<<< HEAD
 </div>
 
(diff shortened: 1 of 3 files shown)


scottgonzalez:
Looks like @harrycheung did rebase but never commented so we weren't notified.

@harrycheung this caused some failures and there's some leftovers from a merge conflict. Could you clean that up, rebase this again, and fix the test failures? Pushing new commits doesn't send a notification, so make sure to ping us afterward. Thanks.

harrycheung:
I rebased and got a bunch of unit test failures. I've kind of hit my upper limit on this, and I've moved on with my internal project. I'll close the pr now.

PR #968, opened 2013-04-26
Droppable event ordering issue

wesleycho:
Fixes bug #9258. Previously, when two droppable elements were nested and an item was dropped on the child (with "greedy" disabled), the order in which the callbacks were called was determined only by the order in which they were bound. This PR changes it so that deeper droppable elements always have their callbacks called first.

### DIFF
--- a/tests/unit/droppable/droppable_events.js
+++ b/tests/unit/droppable/droppable_events.js
@@ -35,8 +35,6 @@ test( "droppable destruction/recreation on drop event", function() {
 	ok( !droppable2.hasClass( "active" ), "subsequent droppable no longer active" );
 });
 
-
-
 // todo: comment the following in when ready to actually test
 /*
 test("activate", function() {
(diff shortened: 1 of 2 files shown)


tests/unit/droppable/droppable_events.js
```diff
@@ -60,4 +58,43 @@ test("drop", function() {
...
 
+test( "drop event callback order", function() {
+	expect( 4 );
+
+	$( "<div id='droppable3'></div>" ).appendTo( "body" );
+	$( "<div id='droppable4'></div>" ).appendTo( $( "#droppable3" ) );
+	$( "<div id='draggable2'></div>" ).appendTo( "body" );
+	var droppable3 = $( "#droppable3" ),
+		droppable4 = $( "#droppable4" ),
+		draggable2 = $( "#draggable2" );
```
scottgonzalez: Why not just store the references as we create the elements? They don't even need IDs then.

tests/unit/droppable/droppable_events.js
```diff
@@ -60,4 +58,43 @@ test("drop", function() {
...
+
+	droppable3.css({ position: 'absolute', width: 100, height: 100, left: 0, top: 0 });
+	droppable4.css({ position: 'absolute', width: 50, height: 50, left: 0, top: 0 });
```
scottgonzalez: Always use double quotes for strings. Might as well just chain these into the creation.

tests/unit/droppable/droppable_events.js
```diff
@@ -60,4 +58,43 @@ test("drop", function() {
...
+
+	droppable3.droppable({ drop: function() { droppable3.addClass( "dropped" ) } });
+	droppable4.droppable({ drop: function() { droppable4.addClass( "dropped" ) } });
```
scottgonzalez: Can chain these as well.

tests/unit/droppable/droppable_events.js
```diff
@@ -60,4 +58,43 @@ test("drop", function() {
...
+
+	draggable2.draggable();
+	draggable2.css({ position: 'absolute', width: 1, height: 1, left: 0, top: 0 });
+
+	draggable2.simulate( "drag", { dx: 1, dy: 1 } );
```
scottgonzalez: And all of these.

ui/jquery.ui.droppable.js
```diff
@@ -42,7 +42,8 @@ $.widget("ui.droppable", {
 	_create: function() {
 
 		var o = this.options,
-			accept = o.accept;
+			accept = o.accept,
+			tgt = this.element;
```
scottgonzalez: Use `target`.

ui/jquery.ui.droppable.js
```diff
@@ -54,9 +55,15 @@ $.widget("ui.droppable", {
 		//Store the droppable's proportions
 		this.proportions = { width: this.element[0].offsetWidth, height: this.element[0].offsetHeight };
 
+		//Store the droppable's depth in the DOM tree
```
scottgonzalez: Space after `//`.

ui/jquery.ui.droppable.js
```diff
@@ -54,9 +55,15 @@ $.widget("ui.droppable", {
...
+		this.depth = 0;
+		while ( tgt.parent().size() >= 1 ) {
+			++this.depth;
+			tgt = tgt.parent();
```
scottgonzalez: Why not just use the length for the depth and grab the last element? This loop isn't necessary.

ui/jquery.ui.droppable.js
```diff
@@ -275,6 +282,16 @@ $.ui.ddmanager = {
 		}
 
 	},
+	addDroppable: function ( scope, droppable ) {
+		$.ui.ddmanager.droppables[ scope ] = $.ui.ddmanager.droppables[ scope ] || [];
+		// TODO: Binary search would be faster
```
scottgonzalez: Blank line above comments.

scottgonzalez: Actually, just drop this. If it ever becomes a problem, someone will file a bug and we can deal with it then. Otherwise, this todo will just sit forever.

scottgonzalez:
Closing due to inactivity.

PR #983, opened 2013-05-08
Allow adding html elements as tooltip content

daniel-o:
Can now pass HTMLElement and jQuery objects to the tooltip content option.
Added tests as well. Hopefully those look right.

Pretty much just changed the conditions that are checked for the contentOption variable. I put some of the conditions on separate lines because I thought it increased readability. So if someone could look at the formatting and scold me if I did it wrong, that would be great.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ docs
 *.patch
 .DS_Store
 .settings
+*.sw?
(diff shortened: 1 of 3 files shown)


jzaefferer:
Thanks @daniel-o for the initial implementation.

PR #990, opened 2013-05-16
Sortable: move createPlaceholder call. Fixes: #9309 - Sortable: scroll jumps up if list is at the bottom of the page

KaneCohen:
Fixes: [#9309](http://bugs.jqueryui.com/ticket/9309) and [#4074](http://bugs.jqueryui.com/ticket/4074)
After fix: <http://jsfiddle.net/pDevf/>

Root of the problem is in the order of how and when stuff is being created. placeholder is being created split millisecond after we assign absolute position to the helper. That is enough for browser to catch that the overall height of the window has been changed which leads to a scroll jump.


scottgonzalez:
As far as I can tell, this bug was fixed a while ago. If this is still an issue for you, please file a new ticket with a reduced test case. Thanks.

PR #1058, opened 2013-08-12
Fixed #5025 - Resizable: Elements that have inner resizable elements do not work

TheZygar:
Added unit tests which prove that the issue exists, and is fixed by my change.


scottgonzalez:
This bug was fixed in 1.11.0.

PR #1060, opened 2013-08-16
Sortable: added checks for the helper touching the edge of the containme...

mickylad:
...nt while dragging. Fixed #5772 - Sortable: containment parent restricting replacement of first and last elements

This issue can be reproduced based on where the mouse is positioned when sorting - in the patched versions you can drag from anywhere in the item to trigger a rearrangement, where as in the unpatched versions you must drag such that the initial mouse down position inside the item will overlap with the target item.

Example fiddles:
[Horizontal](http://jsfiddle.net/morchard/aXkyz/)
[Horizontal Patched](http://jsfiddle.net/morchard/Z7dqj/)
[Vertical](http://jsfiddle.net/morchard/SRXGK/)
[Vertical Patched](http://jsfiddle.net/morchard/Lt9f9/)
[2D](http://jsfiddle.net/morchard/xjjzR/)
[2D Patched](http://jsfiddle.net/morchard/Em4aH/)

There is also an existing issue where the vertical container will be pushed down when sorting into the last index, which is not fixed in this pull request.


lukeapage:
@scottgonzalez do you know why this PR stalled? It was made by a colleague of mine. Is it worth me bringing it up to date with master?
There also seem quite a few other stalled PR's - I well understand some PR's are controversial but it would be good to know the status or what something is waiting on - or if the core team thinks it is not right to be merged then to close the PR with an explanation.

mikesherov:
Yes, this is worth bringing up to date with master. Thanks again!

Typically, a PR is stalled because it's made against a widget that is currently undergoing a rewrite or expected to be rewritten soon, or we respond and it takes a while for the patch author to respond to our comments. Some things just get lost in the shuffle.

Please update this and then ping me and I'll look into getting it merged!

Thanks again

lukeapage:
moved to https://github.com/jquery/jquery-ui/pull/1424
@mickylad can you close this?

PR #1071, opened 2013-09-04
Fixed #8710 - Position: collision: "flip" doesn't work for my: "right top", at: "right bottom" (in some situations)

meyertee:
Bug #8719

Reverted this commit: https://github.com/jquery/jquery-ui/commit/7f808b2047725cd8fde51a948cb4e5f5946c82e1

The intention of the commit is unclear.. I added unit tests to show that the code before that commit already did what it aimed to fix.

http://bugs.jqueryui.com/ticket/8710

### DIFF
--- a/tests/unit/position/position.html
+++ b/tests/unit/position/position.html
@@ -47,6 +47,9 @@ elements smaller than 20px have a line-height set on them to avoid a bug in IE6
 	<div id="bug-5280" style="height: 30px; width: 201px;">
 		<div style="width: 50px; height: 10px;"></div>
 	</div>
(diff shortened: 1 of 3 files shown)


scottgonzalez:
Well, since the original commit didn't reference a bug or include tests, I guess we'll just go with this and see if anyone reports bugs based on `within` :-P

Thanks for handling this @meyertee. Sorry it took so long for us to get to.

lazd:
Hooray!

meyertee:
Great, thank you for merging it :)

PR #1093, opened 2013-09-28
Sortable - fixed parentOffset calculation 

adamziel:
In `Sortable` plugin in the `_getParentOffset` method there is a bit of code described as "a special case where we need to modify a offset calculated on start" which adds `scrollParent.scrollTop()` to the `parentOffset.top`.

It does it if (among others) `scrollParent` is not equal to `document`. On chrome the condition was evaluating to true even if `scrollParent` was `document.body`; This pull request adds necessary check for `document.body`.

### DIFF
--- a/ui/jquery.ui.sortable.js
+++ b/ui/jquery.ui.sortable.js
@@ -954,6 +954,9 @@ $.widget("ui.sortable", $.ui.mouse, {
 		// 1. The position of the helper is absolute, so it's position is calculated based on the next positioned parent
 		// 2. The actual offset parent is a child of the scroll parent, and the scroll parent isn't the document, which means that
 		//    the scroll is included in the initial calculation of the offset of the parent, and never recalculated upon drag
+		//
+		// this.scrollParent[0] !== document.body <-- this condition was added in order to deal with the following WebKit issue:
+		//                                            https://code.google.com/p/chromium/issues/detail?id=157855
 		if(this.cssPosition === "absolute" && this.scrollParent[0] !== document && this.scrollParent[0] !== document.body && $.contains(this.scrollParent[0], this.offsetParent[0])) {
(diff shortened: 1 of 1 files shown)


Arkh1:
Just as a heads up...  This is still an issue.  I ran into the same problem and patching with AZielinski's suggested change fixed the issue.

dmethvin:
So the comment @AZielinski added in 3dbdfeec55737b95e1d87aa917e88149fd250a60 clarifies it is trying to fix [this Chrome bug](https://code.google.com/p/chromium/issues/detail?id=157855). @mikesherov has been prodding the Chrome team in that ticket to land the fix.

I have a feeling this same problem may be relevant to my interests because of https://github.com/jquery/jquery/pull/1714, where the PR proposes to add `offsetParent.scrollTop()` to the offset, but probably shouldn't if the offsetParent is `body` (Chrome) or `document` (everyone else).

mikesherov:
@Arkh1 the proper fix for this calculation is already in the draggable widget. I just haven't yet gotten around to doing all the necessary cleanup and testing for the same thing in sortable yet. There are a ton of edge cases though. :-\

Mykybo:
**Can this be fixed here please?** Or at least provide a way how to workaround it. Waiting for Chrome to fix its behavior will surely take much longer (look how long it is since this was first reported). To me it seems quite common that libraries like jQuery  includes fixes for browser long-living bugs.

http://jsbin.com/nugorawuqi/edit?html,js,output - just scroll down and try to drag any item

blueForestIcarus:
I dont understand why the initial pull request was closed. He provided a fix for the bug, and it was rejected because it didn't have a unit test? It would seem you are allowing a known bug to remain because of the possiblity that some other more obscure bug may still exist. Unit tests aren't all that helpful if you already know the software doesn't work.

PR #1103, opened 2013-10-09
slider: fix to restrict handle values when set programmatically.Fixed #3762-Slider: handles not restricted properly when set programmatically

dekajp:
http://bugs.jqueryui.com/ticket/3762

 `_cleanValues` method added at  `create` , `setOption` and `values` method. 
Test cases added for `option` and `values` method.

### DIFF
--- a/tests/unit/slider/slider_methods.js
+++ b/tests/unit/slider/slider_methods.js
@@ -90,8 +90,31 @@ test( "value", function() {
 	equal( element.slider( "option", "value" ), 1, "value method set respects max" );
 });
 
-//test( "values", function() {
-//	ok(false, "missing test - untested code is broken code." );
-//});
+test( "values", function() {
+	expect( 5 );
+	
+	var element = $( "<div></div>" ).slider({
+		range: true,
+		min: 10,
+		max: 100,
+		step: 1
+	});
+
+	equal( element.slider( "values", 0 ), 10, "get the value of handle" );
+
+	element.slider( "option", "values", [ 10, 25 ] );
+	deepEqual( element.slider( "values" ), [ 10, 25 ], "get value for all handles" );
+
+	element.slider( "values", 0, 50 );
+	equal( element.slider( "values", 0 ), 50, "set the value of handle" );
+
+	element.slider( "values", [ 15, 1 ] );
+	deepEqual( element.slider( "values" ), [ 10, 10 ], "handles restricted properly when set programmatically" );
+
+	element.slider( "option", "values", [ 11, 22, 33, 44 ] );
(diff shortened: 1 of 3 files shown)


tests/unit/slider/slider_methods.js
```diff
@@ -90,8 +90,31 @@ test( "value", function() {
...
+		step: 1
+	});
+
+	equal( element.slider( "values", 0 ), 10, "get the value of handle" );
+
+	element.slider( "option", "values", [ 10, 25 ] );
+	deepEqual( element.slider( "values" ), [ 10, 25 ], "get value for all handles" );
+
+	element.slider( "values", 0, 50 );
+	equal( element.slider( "values", 0 ), 50, "set the value of handle" );
```
scottgonzalez: If the values are `[ 10, 25 ]` and we're updating the first value to `50`, how does that succeed? Shouldn't it be adjusted to `25` because of the existing value of the second handle?

tests/unit/slider/slider_methods.js
```diff
@@ -90,8 +90,31 @@ test( "value", function() {
...
+
+	element.slider( "values", [ 15, 1 ] );
+	deepEqual( element.slider( "values" ), [ 10, 10 ], "handles restricted properly when set programmatically" );
```
scottgonzalez: I'm sure this is because I said "if greater than next handle value, set to next handle value" but this is showing something I didn't consider. Are we talking about the currently set value or the newly provided value? I was actually expecting this to result in `[ 15, 15 ]`.

tests/unit/slider/slider_methods.js
```diff
@@ -90,8 +90,31 @@ test( "value", function() {
...
+
+	element.slider( "option", "values", [ 11, 22, 33, 44 ] );
```
scottgonzalez: This is not actually a valid use of a range slider.

scottgonzalez:
So I think I would change what I said in the ticket. If you're setting all values, then during the iteration, the current handle is *not* restricted by the next handle's value. Only when setting a single handle's value would you be restricted. I also think the restriction only applies to ranges.

scottgonzalez:
I've written out detailed logic in http://bugs.jqueryui.com/ticket/3762#comment:31. I think this is sufficiently different from the implementation, and I let this PR sit long enough, that we should just close this and start over. Sorry for the huge delay and confusion. Hopefully you're still interested in fixing this.

dekajp:
np , I will send new PR soon.

scottgonzalez:
Thanks @dekajp!

PR #1214, opened 2014-03-17
Dialog: Fix resizing with iframes

Rantanen:
Make sure the element that overlays iframes during dialog resize follows
the iframe size during the resize operation.

Fixes #9919

--

Quick fix for the iframe resizing issue. Few things that bother me though:

The iframe the div is overlaying is stored using $.data. Is there a better way and do we need to clear this reference in the _unblockFrames or is it enough to just delete the div itself?

Performance-wise we shouldn't need to iterate through all the iframes in ` _resizeBlocks`. It should be enough to iterate through the iframes inside the current dialog as those are the only ones likely to change size. What would be the cleanest way to implement this? Storing them in different field in ` _blockFrames`, `$(this).children().filter( this.iframeBlocks )`, giving them a className and using `$(this).children("...")` or something else?

### DIFF
--- a/ui/dialog.js
+++ b/ui/dialog.js
@@ -551,6 +551,7 @@ return $.widget( "ui.dialog", {
 				that._trigger( "resizeStart", event, filteredUi( ui ) );
 			},
 			resize: function( event, ui ) {
+				that._resizeBlocks();
 				that._trigger( "resize", event, filteredUi( ui ) );
 			},
 			stop: function( event, ui ) {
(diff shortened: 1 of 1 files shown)


scottgonzalez:
I hate to say this, but given how long this problem has existed, I'm inclined to just leave it until we can address this with the interaction rewrite. Same concern as I expressed in https://github.com/jquery/jquery-ui/pull/1215#issuecomment-50233738.

PR #1262, opened 2014-06-08
Datepicker: allows the correct use of 0-99 years,

BrunoRB:
Fixes #7098, http://bugs.jqueryui.com/ticket/7098.
The cause of the bug is relatively simple, javascript date constructor cast 0-99 years to 1900-1999. In order to fix this I had to introduce the "_getDateObjectWithFixedYear" function and change all the "new Date(year, month, day)" calls to this new function.
Also, this fix only applies when "shortYearCutoff" is setted to 0.

### DIFF
--- a/tests/unit/datepicker/datepicker_options.js
+++ b/tests/unit/datepicker/datepicker_options.js
@@ -1120,4 +1120,45 @@ test( "Ticket 7602: Stop datepicker from appearing with beforeShow event handler
 	inp.datepicker( "destroy" );
 });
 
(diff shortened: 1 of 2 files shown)


scottgonzalez:
Thanks for the patch. Unfortunately, large changes to datepicker tend to result in unforeseen regressions. Since we have a lot of work done on the datepicker rewrite already, we'll just wait for that to fix the bug.

brunobg:
This is a known bug which has been open for years, with a valid patch that was submitted according to your guidelines and that passes your tests. It makes no sense to just sweep it under the rug :(

PR #1270, opened 2014-06-13
Selectmenu: Close on scroll

scottgonzalez:
Fixes #10082

`scroll` events don't bubble, so we need to bind directly to all ancestors. We should take the same approach for other widgets with popups.


scottgonzalez:
Closing this for now.

PR #1287, opened 2014-07-17
Accordion: fixes #9264

JacquesPerrault:
Using .show method and outerHeight property to force height calculation when
heightStyle='auto'


scottgonzalez:
@JacquesPerrault Are you still working on this?

JacquesPerrault:
There's a race condition that I was trying to address with my last commit, but it didn't do the trick.  The whole point is to force the recompute of outerHeight so that the section collapses smoothly.

PR #1324, opened 2014-08-25
Resizable: alsoResize more than one element of a jQuery selection

benmosher:
Fixes #4666.

Did not cause any (new) tests to fail. Several tests were failing on my machine when I pulled the project and set up per the readme (including two [unrelated?] resizable tests).

The selector syntax used in the implementations of `_store` and `_alsoResize` appear to be semantically identical to the removed code (sans the `o.alsoResize[0]`) and function as expected in my personal use.

### DIFF
--- a/tests/unit/resizable/resizable_options.js
+++ b/tests/unit/resizable/resizable_options.js
@@ -419,7 +419,7 @@ test( "alsoResize + multiple selection", function() {
 				width: 50,
 				height: 50
 			})
-			.appendTo( "body" )
+			.appendTo( "body"),
 		element = $( "#resizable1" ).resizable({
 			alsoResize: $.merge(other1, other2),
 			containment: "#container"
--- a/tests/unit/resizable/resizable_options.js
+++ b/tests/unit/resizable/resizable_options.js
@@ -404,4 +404,34 @@ test( "alsoResize + containment", function() {
 	equal( other.height(), 150, "alsoResize constrained height at containment edge" );
 });
 
+test( "alsoResize + multiple selection", function() {
+	expect( 6 );
+	var other1 = $( "<div>" )
+			.addClass("other")
+			.css({
+				width: 50,
+				height: 50
+			})
+			.appendTo( "body" ),
+		other2 = $( "<div>" )
+			.addClass("other")
+			.css({
+				width: 50,
+				height: 50
+			})
+			.appendTo( "body" )
+		element = $( "#resizable1" ).resizable({
+			alsoResize: $.merge(other1, other2),
+			containment: "#container"
+		});
+
+	TestHelpers.resizable.drag( ".ui-resizable-se", 400, 400 );
+	equal( element.width(), 300, "resizable constrained width at containment edge" );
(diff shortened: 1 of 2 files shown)


ui/resizable.js
```diff
@@ -1040,14 +1028,7 @@ $.ui.plugin.add("resizable", "alsoResize", {
 					el.css(style);
 				});
 			};
-
-		if (typeof(o.alsoResize) === "object" && !o.alsoResize.nodeType) {
-			$.each(o.alsoResize, function(exp, c) {
```
scottgonzalez: It looks like this was supporting an undocumented format of the option where the keys were selectors and the values were contexts to search in. I'm ok dropping this since I don't think this was *ever* documented, but we should be aware of it.

ui/resizable.js
```diff
@@ -1040,14 +1028,7 @@ $.ui.plugin.add("resizable", "alsoResize", {
...
-				_alsoResize(exp, c);
-			});
-		} else {
-			_alsoResize(o.alsoResize);
-		}
+		_alsoResize(o.alsoResize);
```
scottgonzalez: The function call is no longer needed since there's a single path. Just do the looping directly. The logic inside the loop should also be cleaned up since there are no longer any contexts provided for the searching.

ui/resizable.js
```diff
@@ -994,19 +994,7 @@ $.ui.plugin.add("resizable", "alsoResize", {
...
-				_store(o.alsoResize);
-			} else {
-				$.each(o.alsoResize, function(exp) {
-					_store(exp);
-				});
-			}
-		} else {
-			_store(o.alsoResize);
-		}
+		_store(o.alsoResize);
```
scottgonzalez: This can also be inlined instead of having a function now.

tests/unit/resizable/resizable_options.js
```diff
@@ -404,4 +404,34 @@ test( "alsoResize + containment", function() {
 	equal( other.height(), 150, "alsoResize constrained height at containment edge" );
 });
 
+test( "alsoResize + multiple selection", function() {
+	expect( 6 );
+	var other1 = $( "<div>" )
+			.addClass("other")
```
scottgonzalez: spacing

tests/unit/resizable/resizable_options.js
```diff
@@ -404,4 +404,34 @@ test( "alsoResize + containment", function() {
...
+			.css({
+				width: 50,
+				height: 50
+			})
+			.appendTo( "body" ),
+		other2 = $( "<div>" )
+			.addClass("other")
```
scottgonzalez: spacing

tests/unit/resizable/resizable_options.js
```diff
@@ -404,4 +404,34 @@ test( "alsoResize + containment", function() {
...
+			.css({
+				width: 50,
+				height: 50
+			})
+			.appendTo( "body"),
+		element = $( "#resizable1" ).resizable({
+			alsoResize: $.merge(other1, other2),
```
scottgonzalez: `other1.add( other2 )`

PR #1333, opened 2014-09-02
Button Re-write

arschmitz:
This replaces PR #1216 

This includes new widget controlgroup which replaces buttonset
Current known issues 
 - [x] Checkboxradio testfailure missing default element (@scott_gonzalez said it should not have one because it cant be sane ) this causes common tests to fail
 - [x] Backcompat: Draft this on the wiki
 - [x] Controlgroup icon only buttons 1px too short in chrome, firefox extra weird @jaspermdegroot think you could take a look?
 - [x] demos/controlgroup/default.html As above, labels should demo something
 - [x] demos/checkboxradio/default.html doesn't have the "command line" anymore, but still doesn't look like a demo. Labels should reflect some real world usecase.
 - [ ] demos/button/icons.html still looks like a visual test, not a demo


ui/checkboxradio.js
```diff
@@ -0,0 +1,287 @@
...
+		var checked = this.element.is( ":checked" );
+		this.label.toggleClass( this._classes( "ui-" + this.type + "-checked" ) +
+			" ui-state-active", checked );
+		if ( this.options.icon && this.type === "checkbox" ) {
+			this.icon
+				.toggleClass( "ui-icon-check", checked )
+				.toggleClass( "ui-icon-blank", !checked );
+		}
+		if ( this.type === "radio" ) {
+			radioGroup( this.element[0] )
```
arschmitz: It's not a private method just to avoid adding it to each instance. It could be inline I suppose. It was also external before in button. I wonder if there is a use case to over ride the logic here though? If so it should be a private method so it can also be an extension point @scottgonzalez @tjvantoll thoughts?

arschmitz:
@jaspermdegroot Turns out its actually not possible to set width auto on selectmenu buttons in a vertical controlgroup which is why I was not. Selectmenu sets an actual width on the button via JS no matter what setting auto actually breaks this and makes it tiny. It seems like there should be a way to completely stop the selectmenu from setting a width for cases like this where you want it to fill a containers width. /cc @fnagel

ui/checkboxradio.js
scottgonzalez: I'd keep it as is. Ideally this would be something like `element.radioGroup()`, but it doesn't really make sense for us to create a proper plugin and have to document it and add a dependency. The reason it doesn't exclude the current element is because it does what you'd expect: finds all radios associated with the current one. This method already existed and its scope didn't change.

arschmitz:
Closing replaced by #1415

demos/controlgroup/splitbutton.html
```diff
@@ -0,0 +1,53 @@
...
+	<script src="../../ui/button.js"></script>
+	<script src="../../ui/checkboxradio.js"></script>
+	<script src="../../ui/menu.js"></script>
+	<script src="../../ui/position.js"></script>
+	<script src="../../ui/selectmenu.js"></script>
+	<script src="../../ui/controlgroup.js"></script>
+	<link rel="stylesheet" href="../demos.css">
+	<script>
+	$(function() {
+		$( "select" ).selectmenu({
```
jzaefferer: This wasn't addressed in #1415.

arschmitz: Yes it was https://github.com/jquery/jquery-ui/pull/1415/files#diff-b643773f101aeba505f286557eb71bc5R25

ui/checkboxradio.js
jzaefferer: I still don't understand why we don't just make it a private instance method. Its used once inside the instance, so it doesn't share any of the requirements that `formResetHandler` has. Why make unnecessary one-off exceptions?

scottgonzalez: We can move it and get rid of the parameter.

arschmitz: will do

tjvantoll: +1

jaspermdegroot:
@arschmitz 

> setting auto actually breaks this and makes it tiny

I don't see that problem. The screenshot below shows the result when I set ```width: auto important!;``` for the selectmenu button in the vertical controlgroup and ```margin: 0;``` for all buttons in the vertical controlgroup. To illustrate the problem I set ```font-size: 22px;``` on the ```body```.

image

With current CSS, changing the font-size to 22px results in this:

image

Using ```!important``` to override the width that the JS sets on the button should be avoided, so I agree this should be resolved in the selectmenu widget so it doesn't set a width when the select is in a vertical controlgroup.


Ps. I think you forgot to actually close this PR :)

arschmitz:
@jaspermdegroot exactly and using `!important` is not acceptable. It makes it essentially impossible for anyone to override. I'm surprised this is not in the style guide /cc @sfrisk

jaspermdegroot:
Well, you can override it. Just need to use ```!important``` in your override as well and make sure it comes after the framework CSS. But, like I said, we should avoid using it and it might indeed be good to add a note about that to the style guide.

Will the new selectmenu widget still use JS to set a width as far as you know? I was wondering what problem had to be solved that couldn't be done with CSS only.

arschmitz:
@jaspermdegroot I'm not sure what you mean by new selectmenu widget? This is the UI widget which just landed in 1.11 ( the current stable version )

jaspermdegroot:
@arschmitz - Sorry, I had Mobile in my mind where we didn't adopt the new selectmenu yet.

PR #1350, opened 2014-09-25
Slider: More refined handling of overlapping handles

gwagroves:
In "range" mode: A click on the slider affects the handle **closest to the click**.

If the two handles share the same value, and both are nearest to the click, the last active handle is selected (or the right most slider if they have the minimum value (this is a moot point – the right-most handle will always be the last active in this case)).

**New behaviour:** If the nearest handles have the same value, and the mouse click value is _lower_ than (i.e. to the left of) the handles' value, the lower (left-most) handle is selected. If the mouse click value is _higher_, the higher (right-most) handle is selected. If the mouse click is _exactly_ on the value of the handles, the last selected will be selected.

Other smaller changes:
commit 92dca47 : refactored to avoid unnecessary `else`
commit c8823b1 + 3fd1806 :  added helper function to avoid repetition of `( this.options.values && this.options.values.length )`
commit afc6124 : add index-based classes to handles

Thanks!
Tim

### DIFF
--- a/ui/slider.js
+++ b/ui/slider.js
@@ -300,7 +300,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 			handle: this.handles[ index ],
 			value: this.value()
 		};
-		if ( this._isRange() ) {
+		if ( this._hasMultipleValues() ) {
 			uiHash.value = this.values( index );
 			uiHash.values = this.values();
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Closing due to inactivity.

PR #1364, opened 2014-10-20
Droppable: $.ui.ddmanager.current released when droppable destroys

eshcharc:
Fixes #10666


scottgonzalez:
This has the same problems as we discussed in #1366. We'd be happy to review a new PR that addresses all the of previous and current issues. Thanks.

PR #1380, opened 2014-10-31
Sortable: Append a tr with td to the placeholder of tbody elements

oemmes:
When sorting tbody elements of a table the placeholder needs to have a tr with
td elements to be visible. The appended elements are created in the same way
as for the placeholder of a tr element; the first row of the sorted tbody is
used for that.

### DIFF
--- a/tests/unit/sortable/sortable_options.js
+++ b/tests/unit/sortable/sortable_options.js
@@ -388,6 +388,42 @@ test( "{ placholder: String } tr", function() {
 	});
 });
 
+test( "{ placholder: String } tbody", function() {
+	expect( 6 );
+
+	var originalWidths,
+		element = $( "#sortable-table" ).sortable({
+			placeholder: "test",
+			start: function( event, ui ) {
+				var currentWidths = otherBody.children().map(function() {
+					return $( this ).width();
+				}).get();
+				ok( ui.placeholder.hasClass( "test" ), "placeholder has class" );
+				deepEqual( currentWidths, originalWidths, "table cells maintain size" );
+				equal( ui.placeholder.children().length, 1,
+					"placeholder has one child" );
+				equal( ui.placeholder.children( "tr" ).length, 1,
+					"placeholder's child is tr" );
+				equal( ui.placeholder.find( "> tr" ).children().length,
+					dragBody.find( "> tr:first" ).children().length,
+					"placeholder's tr has correct number of cells" );
+				equal( ui.placeholder.find( "> tr" ).children().html(),
+					$( "<span>&#160;</span>" ).html(),
+					"placeholder td has content for forced dimensions" );
+			}
+		}),
+		bodies = element.children( "tbody" ),
+		dragBody = bodies.eq( 0 ),
+		otherBody = bodies.eq( 1 );
+
+	originalWidths = otherBody.children().map(function() {
+		return $( this ).width();
+	}).get();
+	dragBody.simulate( "drag", {
+		dy: 1
+	});
+});
+
 /*
 test("{ revert: false }, default", function() {
 	ok(false, "missing test - untested code is broken code.");
--- a/ui/sortable.js
+++ b/ui/sortable.js
@@ -776,8 +776,7 @@ return $.widget("ui.sortable", $.ui.mouse, {
 	_createPlaceholder: function(that) {
 		that = that || this;
 		var className,
-			o = that.options,
-			tr;
+			o = that.options;
 
 		if(!o.placeholder || o.placeholder.constructor === String) {
 			className = o.placeholder;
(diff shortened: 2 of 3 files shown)


scottgonzalez:
@tjvantoll Were you still planning on reviewing this?

ui/sortable.js
```diff
@@ -830,6 +831,16 @@ return $.widget("ui.sortable", $.ui.mouse, {
 
 	},
 
+	_createTrPlaceholder: function(that, sourceTr, targetTr) {
+		that = that || this;
```
tjvantoll: The first parameter is unnecessary. You can change this line to `var that = this;`.

ui/sortable.js
```diff
@@ -830,6 +831,16 @@ return $.widget("ui.sortable", $.ui.mouse, {
 
 	},
 
+	_createTrPlaceholder: function(that, sourceTr, targetTr) {
```
tjvantoll: Spaces around the parenthesis—`( sourceTr, targetTr )`

ui/sortable.js
```diff
@@ -776,7 +776,8 @@ return $.widget("ui.sortable", $.ui.mouse, {
 	_createPlaceholder: function(that) {
 		that = that || this;
 		var className,
-			o = that.options;
+			o = that.options,
+			tr;
```
tjvantoll: Flip these two lines. Uninitialized variables come first.

tjvantoll: Actually I'd just remove the variable.

ui/sortable.js
```diff
@@ -788,12 +789,12 @@ return $.widget("ui.sortable", $.ui.mouse, {
...
 							.removeClass("ui-sortable-helper");
 
-					if ( nodeName === "tr" ) {
-						that.currentItem.children().each(function() {
-							$( "<td>&#160;</td>", that.document[0] )
-								.attr( "colspan", $( this ).attr( "colspan" ) || 1 )
-								.appendTo( element );
-						});
+					if ( nodeName === "tbody" ) {
+						tr = $( "<tr></tr>", that.document[0] )
```
tjvantoll: I'm curious why this is creating a `<tr>` instead of another `<tbody>` to act as a placeholder? As is, sorting `<tbody>`s is kind of weird because the injected `<tr>` used as a placeholder is not the same height as a `<tbody>`.

oemmes:
@tjvantoll: Thanks for the feedback.
I made some changes to the code according to your suggestions.

A little explanation for the `<tr>` element that is created:
The actual placeholder `element` is already a `<tbody>` which is added to the sortable. But it is not visible unless it contains any cells. That's the reason for adding the `<tr>` to it.

tjvantoll:
Ok that makes sense, and thanks for sticking with this. @scottgonzalez this looks good to me.

ui/sortable.js
```diff
@@ -788,12 +788,10 @@ return $.widget("ui.sortable", $.ui.mouse, {
...
 							.removeClass("ui-sortable-helper");
 
-					if ( nodeName === "tr" ) {
-						that.currentItem.children().each(function() {
-							$( "<td>&#160;</td>", that.document[0] )
-								.attr( "colspan", $( this ).attr( "colspan" ) || 1 )
-								.appendTo( element );
-						});
+					if ( nodeName === "tbody" ) {
+						that._createTrPlaceholder( that.currentItem.find( "tr:first" ), $( "<tr></tr>", that.document[0] ).appendTo( element ) );
```
scottgonzalez: * Avoid use of `:first` selector, use a filter method like `.eq( 0 )`.
* Empty element construction uses just the opening name: $( "<tr>" )
* Missing spaces inside brackets.
* This line exceeds 100 characters.

ui/sortable.js
```diff
@@ -830,6 +828,16 @@ return $.widget("ui.sortable", $.ui.mouse, {
 
 	},
 
+	_createTrPlaceholder: function( sourceTr, targetTr ) {
+		var that = this;
+
+		sourceTr.children().each(function() {
+			$( "<td>&#160;</td>", that.document[0] )
```
scottgonzalez: I know you just copy/pasted this, but since other code needs to be updated, might as well fix the spacing here as well.

tests/unit/sortable/sortable_options.js
```diff
@@ -388,6 +388,42 @@ test( "{ placholder: String } tr", function() {
...
+					dragBody.find( "> tr:first" ).children().length,
+					"placeholder's tr has correct number of cells" );
+				equal( ui.placeholder.find( "> tr" ).children().html(),
+					$( "<span>&#160;</span>" ).html(),
+					"placeholder td has content for forced dimensions" );
+			}
+		}),
+		bodies = element.children( "tbody" ),
+		dragBody = bodies.eq( 0 ),
+		otherBody = bodies.eq( 1 );
```
scottgonzalez: Move these up so that reading the code from top to bottom has the proper context.

scottgonzalez: Actually, feel free to ignore this since there are other tests written like this already. We can address during the interaction rewrites.

oemmes:
@scottgonzalez I adjusted the code again. Hopefully it fits the guidelines now.

scottgonzalez:
Thanks, squashed and landed in master.

PR #1384, opened 2014-11-07
Theme

jaspermdegroot:
I wasn't sure if switching to shorthand for hex color codes would cause problems with the ThemeRoller so I didn't follow our new CSS style guide.


jzaefferer:
@rxaviers can you look into this asap? We're getting closer to 1.12.

rxaviers:
Sure, it will be my priority #1 for tomorrow.

themes/base/tooltip.css
```diff
@@ -13,8 +13,8 @@
 	position: absolute;
 	z-index: 9999;
 	max-width: 300px;
-	-webkit-box-shadow: 0 0 5px #aaa;
-	box-shadow: 0 0 5px #aaa;
+	-webkit-box-shadow: 0 0 5px #ddd;
+	box-shadow: 0 0 5px #ddd;
```
rxaviers: What's the motivation of this change? I'm asking, because it doesn't feel natural to me having a structural stylesheet changed for a theme adjustment.

Perhaps the question must be should we consider moving this into theme.css? Tooltip.css is the only structural stylesheet including a color as far as I can tell.

rxaviers: In today's UI meeting, it was decided to "Update tooltip to use ui-widget-shadow class (via classes option, once that lands), move box-shodow definition into theme.css/.ui-widget-shadow, update TR to demo that instead of the weird fake shadow we have currently"

rxaviers:
Except for the change in `tooltip.css`, everything else have been made onto `theme.css` and are bound directly to a variable. It means TR can generate it just fine for 1.12 and 1.11 (and earlier versions - need to check). Note it doesn't necessarily mean user will be able to use the TR interface to come up with the above theme. Because, a few variables changed here have no corresponding UI (i.e., `bg*Url`, `bg*XPos` and `bg*YPos`).

Currently, the default theme of a custom downloaded package is UI Lightness. The theme that reflects the base theme is called Smoothness. @jzaefferer and I have discussed and brought to UI's meeting today the inconsistency of having UI Lightness as the default theme of a custom download, but using the base (aka smoothness) theme as default on the repo and demos (including the one in the api docs).

The accepted suggestion was to keep Smoothness and UI Lightness in the gallery, while the new theme will be available as “Base” and it will be the default everywhere.

jaspermdegroot:
@rxaviers 

> Note it doesn't necessarily mean user will be able to use the TR interface to come up with the above theme. Because, a few variables changed here have no corresponding UI (i.e., bg*Url, bg*XPos and bg*YPos).

You can select a background image in the ThemeRoller so I am not sure what you mean by no UI for bg Url.
Regarding bg XPos and bg YPos: The user doesn't need to be able to change the values for those variables so it's fine that there is no UI for it. Their values are hardcoded (see my [comment](https://github.com/jquery/jquery-ui/pull/1384#issuecomment-63512483)). No changes here compared to current situation.
We only need to make sure that the ```background-position``` rule is only included in the theme CSS when the user sets a background image, not when s/he doesn't (because this new default theme doesn't use bg images). If I remember correctly the current script in /lib/themeroller.js will take care of this.

rxaviers:
@jaspermdegroot you cannot select none. You must always pick a texture via TR UI.

jaspermdegroot:
@rxaviers - Do you think it's easy to add "none" to the texture control?

rxaviers:
Not hard for sure.

rxaviers:
@jaspermdegroot, @jzaefferer and @scottgonzalez feel free to review https://github.com/jquery/download.jqueryui.com/pull/248. (a) there is a picture of the ThemeRoller displaying "Base", (b) there are jQuery UI 1.9.2, 1.10.4, 1.11.2, 1.12.0-pre packages available for download using "Base", (c) there are steps to reproduce DB/TR locally using "Base". Just let me know in any questions.

themes/base/tooltip.css
jzaefferer: @jaspermdegroot could you implement this? Let me know if you have questions.

jaspermdegroot: @jzaefferer - Sure, I will make some time for this next week.

kborchers:
I thought I had seen at one point a mockup of the new theme without the rounded corners. Can we modernize this new theme a bit more by removing them rather than just slightly changing the radius?

arschmitz:
Personal prefrence i would not remove the radious entirely but i would make it a lot smaller.

kborchers:
Yeah, I guess maybe a smaller radius would be fine too. Another look does show most themes still have some radius but they are much smaller so I would be fine with that too. Maybe 3px is ok. It could be better use of shadows in other framework themes that I like more as well. I don't know, I'm not a designer and just throwing in my 2 cents so take it or leave it. :smile:

jaspermdegroot:
@kborchers @arschmitz - I am not sure what you guys are looking at. I did already reduce the border-radius to 3px in this PR.

And if someone doesn't want rounded corners at all, he/she can remove the corner class with the new classes option I suppose.

rxaviers:
Is this theme supposed to work on 1.11.x, 1.10.x or 1.9.x too without any change? Or is this only intended to work on 1.12.x and up?

themes/base/tooltip.css
scottgonzalez: @jaspermdegroot Did you get a chance to work on this?

scottgonzalez:
I'd say it only matters for 1.12+, but I would assume that it would work on all versions.

themes/base/tooltip.css
jaspermdegroot: @scottgonzalez - I didn't because @jzaefferer already created PR #1420. I think that PR, together with https://github.com/jquery/download.jqueryui.com/pull/248, covers everything that needs to be done. Or did I miss something?

rxaviers:
I'm concerned about having download builder (which serves 1.11.x, 1.10.x and 1.9.x) to distribute this theme by default giving it hasn't been tested on previous versions. So, I'm wondering if @jaspermdegroot (that created the theme) or anyone that is reviewing it could ensure it can be used just fine without any modification to the previous releases.

themes/base/tooltip.css
scottgonzalez: Ah, right. Nevermind. I was just looking over this and saw the last comment saying you were going to make a change last week.

scottgonzalez:
Isn't this exactly what ThemeRoller does? I don't see any reason this wouldn't work.

rxaviers:
I made a wrong assumption (discussed via IRC), nevermind.

jzaefferer:
I'm closing this in favour of #1420 to prevent further confusion.

PR #1394, opened 2014-12-03
Make widget factory remove classes in destroy

arschmitz:
This alters the proposed api of _addClass and _removeClass to make them auto remove all classes in destroy
it also add a garbage collection method for the object which tracks classes elements

this changes the api to:
`
this._addClass( elements, "classes" );
`

It removes the need to differentiate the "extra" classes and tracks all classes and removes all classes in destroy automatically


arschmitz:
replaced by #1411

PR #1398, opened 2014-12-16
Slider: Fix max calculation, when step is float

dekajp:
Fixes #10721

new _calculateNewMax method uses while loop, the cost should be very minimum
since this method is called one time in life of control, in create method or
when  setOption(max) is called.

below method uses math functions to avoid loop, but this was failing
unit tests of slider_options.

TODO : to avoid while loop in Max Calculation

```

_calculateNewMax: function() {
  var precision = this._precision();
  var multiplier = Math.pow( 10 , precision );
  var max = this.options.max * multiplier ;
  var min = this._valueMin() * multiplier ;
  var step = this.options.step * multiplier ;
  var remainder = ( max - min ) % step;
  this.max = ( ( max - remainder ) / multiplier ).toFixed(precision);
}

```

### DIFF
--- a/ui/slider.js
+++ b/ui/slider.js
@@ -552,8 +552,26 @@ return $.widget( "ui.slider", $.ui.mouse, {
 	},
 
 	_calculateNewMax: function() {
-		var remainder = ( this.options.max - this._valueMin() ) % this.options.step;
-		this.max = this.options.max - remainder;
+		var max = this.options.max,
+			min = this._valueMin(),
+			step = this.options.step,
+			aboveMin = Math.floor( ( max - min ) / step ) * step;
+		max = aboveMin + min;
+		this.max = parseFloat( max.toFixed( this._precision() ) );
+	},
+
+	_precision: function() {
+		var precision = this._precisionOf( this.options.step );
+		if ( this.options.min !== null ) {
+			precision = Math.max( precision, this._precisionOf( this.options.min ) );
+		}
+		return precision;
+	},
+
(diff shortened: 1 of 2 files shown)


ui/slider.js
```diff
@@ -552,8 +552,27 @@ return $.widget( "ui.slider", $.ui.mouse, {
 	},
 
 	_calculateNewMax: function() {
-		var remainder = ( this.options.max - this._valueMin() ) % this.options.step;
-		this.max = this.options.max - remainder;
+		var precision = this._precision(),
+			multiplier = ( Math.pow( 10, precision ) ),
+			max = ( this.options.max * multiplier ),
+			min = ( this._valueMin() * multiplier ),
+			step =( this.options.step * multiplier ),
```
scottgonzalez: Please remove all of the extraneous parens.

ui/slider.js
```diff
@@ -552,8 +552,27 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+			remainder = ( max - min ) % step;
+		this.max = parseFloat( ( ( max - remainder ) / multiplier ).toFixed( precision ) );
```
scottgonzalez: Couldn't this follow the same logic as spinner and avoid the modulo operation? Something along the lines of:

```js
aboveMin = Math.floor( ( max - min ) / step ) * step;
max = aboveMin + min;
this.max = parseFloat( max.toFixed( this._precision() ) );
```

PR #1407, opened 2014-12-30
Slider: extract distance calculation logic into dedicated methods

ssaki:
Fixes #10737 

~ Extracted the code responsible for translation of values into slider handle's distances into _calculateDistance() (was in _calculateDistance())
~ Extracted the code responsible for translation of slider handle's distance into numeric values into _calculateValue() (was in _normValueFromMouse())
~ Added two new callback options and modified the constructor to set the new methods as default values (if not already set)
~ cleaned up all variables that became unused due to the re-factoring
~ updated the common test case to reflect the change in options

### DIFF
--- a/tests/unit/slider/slider_common.js
+++ b/tests/unit/slider/slider_common.js
@@ -18,6 +18,9 @@ TestHelpers.commonWidgetTests( "slider", {
 		change: null,
 		slide: null,
 		start: null,
-		stop: null
+		stop: null,
+		
+		calculateDistance: null,
+		calculateValue: null
 	}
 });
--- a/ui/slider.js
+++ b/ui/slider.js
@@ -50,7 +50,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		slide: null,
 		start: null,
 		stop: null,
-		
+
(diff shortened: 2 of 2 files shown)


arschmitz:
@ssaki I agree completely with having an extension point. I was only saying we may want to wait for the new widget to add it. We generally don't add anything new to widgets that are going to be re-written, until the re-write.

jzaefferer:
I think its fine to work this out before the rewrite. This PR needs an update, as TJ described above. It also missing unit tests that cover this extension. And this should be accompanied by a [API docs PR](https://github.com/jquery/api.jqueryui.com/blob/master/entries/slider.xml) to document the new extension point(s), similar to, for example, [_isDivider](http://api.jqueryui.com/menu/#method-_isDivider) in menu.

scottgonzalez:
Since this involves public API, please make proposals for the new APIs in the ticket. Once we have consensus on the API, then you can update the PR.

scottgonzalez:
Closing due to inactivity. If you're still interested in working on this, please start by proposing the actual API in the ticket.

PR #1412, opened 2015-01-04
Spinner: Support RTL

RockQ:
Using "dir=rtl" in markup to config the spinner arrows alignment. Add "ui-rtl" class on widget element for rtl styling.

Fixes [#10657](http://bugs.jqueryui.com/ticket/10657)


scottgonzalez:
Thanks. You can ignore the failing datepicker test since it's unrelated.

We'll need to have a larger discussion around using this idea across all widgets before we can properly review this land land it, so the review may take some time.

RockQ:
Okay, look forward your comments later.

lukeapage:
my 2 cents (I'm not saying this with any level of authority so take this with a pinch of salt if you think I'm completely wrong)

I've worked on a large web application with RTL support - it is not a small undertaking. You need to decide whether you want things to look the same regardless of the direction set on the document or whether you want controls to all "reverse" or to look the same as LTR. If you want them reversed then every float/margin/padding/border/left/right needs reversing (while inline/inline-block elements get reversed based on the direction attribute already).

Either way, it makes sense to keep as much as possible in the css (although there are of course some challanges with that). In terms of the css, overriding all the ltr css with rtl css leads to double the amount of css content - it is not a good idea and it adds a big support drain. My favoured approach is to maintain a ltr css and a rtl css and to not support having a mix of both on the same page. You can then use preprocessor plugins to auto convert certain properties if you want (this worked quite well for me and more consistent than manual overriding).

The approach in this pr has a few advantages - 1 - it allows a ltr and rtl spinner to exist on the same page together. 2 - you do not need to specify seperately a single rtl setting for the whole page.
However, you will end up with querying rtl everywhere and you will incur a large amount of bloat everywhere. IMO I don't think its worth it - a simpler approach is more maintainable and might even result in something that works (or is close to working) across all of jquery-ui just by using an appropriate tool.

Regardless of the approach jquery ui takes, this will require a visual test I would have thought?

scottgonzalez:
@lukeapage We're actually going to be scheduling a call to discuss RTL support soon. Are you interested in joining that discussion?

lukeapage:
If I am available, I will - but I am based in the UK and will be unavailable during work hours next week.

RockQ:
@lukeapage  @scottgonzalez , here is visual test of rtl setting on jsFiddle, http://jsfiddle.net/nhvwL5my/

RockQ:
@lukeapage , and thanks for the comments, IMO, the first thing we need to confirm is that "rtl" config is on page level (global) or is on widget level(individual)? Probably we need both, then what's the priority?

RockQ:
Hi @scottgonzalez, if there's any conclusion of rtl discussion, please let me know:) thanks!

scottgonzalez:
I'll be sending out a doodle tomorrow to find a time for the call; I'm planning to use the first week of February as the time frame.

scottgonzalez:
This has become pretty stale and the same group is now working on #1682.

PR #1413, opened 2015-01-05
Accordion: Set `aria-expanded` when collapsing

scottgonzalez:
Fixes #10703

### DIFF
--- a/ui/accordion.js
+++ b/ui/accordion.js
@@ -488,7 +488,10 @@ return $.widget( "ui.accordion", {
 		toHide.attr({
 			"aria-hidden": "true"
 		});
-		toHide.prev().attr( "aria-selected", "false" );
+		toHide.prev().attr({
+			"aria-selected": "false",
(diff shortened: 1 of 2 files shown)


ui/accordion.js
```diff
@@ -499,7 +502,7 @@ return $.widget( "ui.accordion", {
 			});
 		} else if ( toShow.length ) {
 			this.headers.filter(function() {
-				return $( this ).attr( "tabIndex" ) === 0;
+				return $( this ).attr( "tabIndex" ) === "0";
```
mzgol: AFAIK IE8 is fine here; IE<8 have attroperties and they return numbers but you don't support IE<8 so it should be good.

dmethvin: You probably want to use `parseInt` for the best compat. I think on older versions core returns a number when it should return a string. Looks like it changed in 1.8.something  http://jsfiddle.net/qvpq4wo1/

PR #1414, opened 2015-01-06
Datepicker, allow right-alignment of widget

Nefarion:
Added an option to right-align the Datepicker with the input.


scottgonzalez:
Thanks, but we're not accepting any new options as we're in the process of a complete rewrite. Also, the rewrite uses CLDR to handle all locale data, so this would not be an explicit option anyway.

PR #1416, opened 2015-01-09
Dialog: respect minWidth option for auto-sized dialog on init

karol-wojcik:
When dialog is created with 'auto' size option, it gets 'auto' width that can be less than minWidth option. The code fixes the problem.


scottgonzalez:
`"auto"` is not a supported value for the `width` option. See the [documentation](http://api.jqueryui.com/dialog/#option-width).

PR #1419, opened 2015-01-10
Fix DatePicker Tests running in 2015

lukeapage:
If you would rather mock the current date so this doesn't happen again (or use the current date - 2008 to work out the offset) then I can refactor

### DIFF
--- a/tests/unit/datepicker/datepicker_options.js
+++ b/tests/unit/datepicker/datepicker_options.js
@@ -506,8 +506,8 @@ test("minMax", function() {
 
 	inp.datepicker("option", {
 		minDate: new Date(1900, 0, 1),
-		maxDate: "-6Y",
-		yearRange: "1900:-6"
+		maxDate: "-7Y",
+		yearRange: "1900:-7"
 	}).val( "" );
 	ok(dp.find(".ui-datepicker-next").hasClass("ui-state-disabled"), "Year Range Test - relative - next button disabled");
 	ok(!dp.find(".ui-datepicker-prev").hasClass("ui-state-disabled"), "Year Range Test - relative - prev button enabled");
--- a/tests/unit/datepicker/datepicker_options.js
+++ b/tests/unit/datepicker/datepicker_options.js
@@ -509,16 +509,16 @@ test("minMax", function() {
 		maxDate: "-6Y",
 		yearRange: "1900:-6"
 	}).val( "" );
-	ok(dp.find(".ui-datepicker-next").hasClass("ui-state-disabled"), "Year Range Test - next button disabled");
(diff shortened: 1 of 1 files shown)


tests/unit/datepicker/datepicker_options.js
```diff
@@ -506,19 +506,19 @@ test("minMax", function() {
 
 	inp.datepicker("option", {
 		minDate: new Date(1900, 0, 1),
-		maxDate: "-6Y",
+		maxDate: "-7Y",
 		yearRange: "1900:-6"
```
scottgonzalez: We should probably update this to use `-7` as well, just for consistency.

tests/unit/datepicker/datepicker_options.js
```diff
@@ -506,19 +506,19 @@ test("minMax", function() {
...
 
 	inp.datepicker("option", {
 		minDate: new Date(1900, 0, 1),
 		maxDate: "1/25/2007",
 		yearRange: "1900:2007"
 	}).val( "" );
-	ok(dp.find(".ui-datepicker-next").hasClass("ui-state-disabled"), "Year Range Test - next button disabled");
-	ok(!dp.find(".ui-datepicker-prev").hasClass("ui-state-disabled"), "Year Range Test - prev button enabled");
+	ok(dp.find(".ui-datepicker-next").hasClass("ui-state-disabled"), "Year Range Test - absolute - next button disabled");
+	ok(!dp.find(".ui-datepicker-prev").hasClass("ui-state-disabled"), "Year Range Test -absolute - prev button enabled");
```
scottgonzalez: Missing space after the first dash.

scottgonzalez:
Thanks, I think this is good enough since we're pretty far into our datepicker rewrite, which includes a refactoring of these tests which don't suffer from this problem.

Can you please sign our [CLA](http://contribute.jquery.org/CLA/)?

lukeapage:
adjusted for code review and CLA signed.

Good to know about the datepicker rewrite - I was wondering about that. Am interested in doing some contributing, so thought this would be a good way to ease in. If there is anything else small (can be difficult - I like difficult) you can point me to, that would be nice.

jzaefferer:
Thanks for the update, looks good, I merged it.

As for other tasks, you can take a look at our roadmap and see if there's something you're interested in: http://wiki.jqueryui.com/w/page/12138038/Roadmap

Not listed there are open issues for 1.12: http://bugs.jqueryui.com/report/17?V=1.12.0
This menu issue should be reasonable to address: http://bugs.jqueryui.com/ticket/10692
Probably a bit more challenging is this one, where you could start with the mentioned dialog method: http://bugs.jqueryui.com/ticket/9845

PR #1420, opened 2015-01-12
Tooltip shadow

jzaefferer:
This is supposed to be pulled into #1384, but the GitHub UI won't let me pick Jasper's fork as the base repository.

This implements what was discussed there: "Update tooltip to use ui-widget-shadow class (via classes option, once that lands), move box-shadow definition into theme.css/.ui-widget-shadow, update TR to demo that instead of the weird fake shadow we have currently".

Since the classes PR hasn't landed in master, yet, this just adds the `ui-widget-shadow` class to the tooltip element directly. We haven't discussed if we want to update ThemeRoller with new UI elements to allow configuration of `box-shadow`, replacing the current UI with its various settings. I tend to just dropping the whole thing. `box-shadow` has many properties, many of them optional, and its really easy to override in a custom CSS file. We don't have variables for every style property in `theme.css` already, so might as well skip this as well.


jzaefferer:
@jaspermdegroot @rxaviers what do you think about this? Just check the one commit I added, ignore the rest.

rxaviers:
@jzaefferer :+1:, My only concern would be making sure box-shadow is supported on every browsers jQuery UI supports. http://caniuse.com/#search=box-shadow has an observation for IE8.

jzaefferer:
> My only concern would be making sure box-shadow is supported on every browsers jQuery UI supports

I think its fine for IE8 not to display the shadow. The tooltip looks okay without it.

> Can't we keep using the same variables:
> [...]
> Maybe rename thicknessShadow to blurShadow to make clear that it sets the blur radius and not the spread radius.

I didn't manage to match the existing properties to the box-shadow properties. Makes sense to just keep them as they are. I can update this PR accordingly. As for renaming: I guess we should rename everything or nothing. This somewhat depends on ThemeRoller, since it doesn't support multiple versions. @rxaviers does it make any difference? Is it better to keep as-is or better to rename?

rxaviers:
About renaming...

Is this change going to propagate as patch fixes for older releases too (e.g., 1.9.3, 1.10.5 and 1.11.3)? If not, supporting both names simultaneously on TR (old names for 1.9.2 and 1.10.4 and 1.11.2, but new names for 1.12.0) will require some extra not trivial code there.

Also, do we want TR to handle old -> new names for people to click in the links of their saved themes and have it properly adjusted? One more extra work.

scottgonzalez:
> Is this change going to propagate as patch fixes for older releases too (e.g., 1.9.3, 1.10.5 and 1.11.3)?

No.

jzaefferer:
Updated this to keep the previous variables. The following variables are gone, though: `bgImgUrlShadow`, `bgShadowXPos`, `bgShadowYPos`, `bgShadowRepeat`, `opacityShadow`, `opacityFilterShadow` and `cornerRadiusShadow`.

jzaefferer:
@rxaviers could you test this update with ThemeRoller?

rxaviers:
https://github.com/jquery/download.jqueryui.com/pull/248 has been updated accordingly.

jzaefferer:
Replaced by #1436, which has all the details.

PR #1422, opened 2015-01-12
All: Drop support for jQuery 1.6.x

jzaefferer:
Affects core, effects, position and widget. Only position has unit tests that fail with jQuery 1.6 without the workaround.

Drops the 1.6.x copies jQuery of jQuery and removes them from the select in the testsuites.

Fixes [#10723](http://bugs.jqueryui.com/ticket/10723)


mikesherov:
I suppose `bind` to `on` is for another PR?

mikesherov:
LGTM

arschmitz:
@mikesherov i have an old branch im going to update and submit that removes our dependency on the event alias core module might make sense to do it at the same time since no in all supported versions bind is an event alias?

tjvantoll:
:+1:

scottgonzalez:
There are support comments in tests too.

jzaefferer:
Added a commit that updates core and tooltip tests. Verified both suites tests pass everywhere using browserstack-runner locally.

scottgonzalez:
Looks good.

jzaefferer:
Landed. @mikesherov @arschmitz yeah, updating bind/on should be done elsewhere. There's currently no ticket (for 1.12) for that.

PR #1423, opened 2015-01-12
Selectmenu: Better handling when there are no options

scottgonzalez:
Fixes #10662

This replaces #1370.

I think this is a better way to handle empty lists in general. The fact that we were relying on `this.menuItems` as a flag for the delayed rendering has always bothered me. This also includes a change that stops the menu from opening when there are no options.

### DIFF
--- a/ui/selectmenu.js
+++ b/ui/selectmenu.js
@@ -66,6 +66,9 @@ return $.widget( "ui.selectmenu", {
 		this._drawButton();
 		this._drawMenu();
 
+		this._rendered = false;
+		this.menuItems = $();
+
 		if ( this.options.disabled ) {
 			this.disable();
 		}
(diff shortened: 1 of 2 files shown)


arschmitz:
+1 for not opening when no options this is what native does.

ui/selectmenu.js
```diff
@@ -199,7 +202,8 @@ return $.widget( "ui.selectmenu", {
 		this._refreshMenu();
 		this.buttonItem.replaceWith(
 			this.buttonItem = this._renderButtonItem(
-				this._getSelectedItem().data( "ui-selectmenu-item" )
+				// Fall back to an empty object in case there are no options
```
jzaefferer: Missing blank line above comment

jzaefferer:
Looks good

PR #1426, opened 2015-01-15
Resizable: Fixed on calculation of width of grid.

odoku:


scottgonzalez:
Thanks, looks like @mikesherov actually just landed the same fix yesterday along with unit tests. See 9493839f23b6ff71aacb9cb8fe8fde6c8d0ecd61. Sorry this got missed.

PR #1427, opened 2015-01-16
_renderItem support for dom element label

alessandro-pezzato:
If item.label is a dom element, it's appended to, not rendered as text


scottgonzalez:
`label` is text and clearly documented.

PR #1428, opened 2015-01-17
Fix max height of the selectmenu and collision default

thg2k:
Hello my favorite dev team,

Here is a patch for some functionality I think should be in selectmenu, i.e. when I have really long selectmenus the popup height should be limited automatically to one third of the screen window height, plus if the menu is below half screen the popup should appear above, not below because it would go outside of the screen.

Let me know if i'm missing out on something. I suggest backporting to 1.11.x.

### DIFF
--- a/ui/selectmenu.js
+++ b/ui/selectmenu.js
@@ -43,7 +43,7 @@ return $.widget( "ui.selectmenu", {
 		position: {
 			my: "left top",
 			at: "left bottom",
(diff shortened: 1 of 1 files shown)


scottgonzalez:
This for sure can't be backported to 1.11.x as it's a major change. Please start proposals in tickets at http://bugs.jqueryui.com/newticket.

PR #1429, opened 2015-01-19
Menu: Use ui-state-active consistently

jzaefferer:
So far we were using ui-state-active on active parent items, ui-state-focus on active child items. The theme update highlighted the visual inconsistency. With this change, only ui-state-active is used.

Fixes [#10692](http://bugs.jqueryui.com/ticket/10692)

This also affects autocomplete and selectmenu, since they both use menu. Only selectmenu tries to mess with menu classes.

The `theme.css` change needs to be dropped and is there only to show what the menu will look like after #1384 lands. As long as we land both before 1.12, this can also land before #1384.

Tested using browserstack-runner locally, running all three affected testsuites, along with clicking around demos and visual tests manually in Chrome.


scottgonzalez:
Looks good.

jzaefferer:
Updated to drop the theme change, will merge that.

PR #1430, opened 2015-01-19
Calendar: Fix unit tests for CMD

fnagel:
* Make keyboard tests work in phantom 
* Run datepicker tests in Grunt qunit task

Remaining: spinner tests (due to Globalize update?)

### DIFF
--- a/tests/unit/calendar/calendar_core.js
+++ b/tests/unit/calendar/calendar_core.js
@@ -139,7 +139,7 @@ asyncTest( "keyboard handling", function() {
 			.focusGrid( element )
 			.simulate( "keydown", { keyCode: $.ui.keyCode.LEFT } );
 		setTimeout(function() {
-			$( ":focus" ).simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
+			$( document.activeElement ).simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
 			TestHelpers.calendar.equalsDate(
 				element.calendar( "valueAsDate" ),
 				new Date( 2013, 12 - 1, 31 ),
@@ -170,7 +170,7 @@ asyncTest( "keyboard handling", function() {
 
 		TestHelpers.calendar.focusGrid( element ).simulate( "keydown", { keyCode: $.ui.keyCode.UP } );
 		setTimeout(function() {
-			$( ":focus" ).simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
+			$( document.activeElement ).simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
 			TestHelpers.calendar.equalsDate(
 				element.calendar( "valueAsDate" ),
 				new Date( 2013, 12 - 1, 25 ),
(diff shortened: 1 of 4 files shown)


tests/unit/calendar/calendar_test_helpers.js
```diff
@@ -16,9 +16,9 @@ TestHelpers.calendar = {
 	},
 	focusGrid: function( element ) {
 		element.find( ":tabbable" ).last().simulate( "focus" );
```
jzaefferer: Should store the selected element and use it for the next three calls, including the return.

tests/unit/calendar/calendar_core.js
```diff
@@ -139,7 +139,7 @@ asyncTest( "keyboard handling", function() {
 			.focusGrid( element )
 			.simulate( "keydown", { keyCode: $.ui.keyCode.LEFT } );
 		setTimeout(function() {
-			$( ":focus" ).simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
+			$( document.activeElement ).simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
```
jzaefferer: The `focusGrid` call above returns the to-be-focused element. Could store that in a variable and reuse it here. Applies to many other places as well. Not sure if its better to avoid a temp var at the cost of accessing globals.

tests/unit/datepicker/datepicker_events.js
```diff
@@ -108,7 +108,7 @@ asyncTest( "select", function() {
 			.simulate( "focus" )
 			.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
 		setTimeout(function() {
-			$( ":focus" )
+			$( document.activeElement )
```
jzaefferer: Why isn't using this the `input` variable? Same below.

tests/unit/calendar/calendar_test_helpers.js
fnagel: Doesn't the tab move the focus so we need to retrieve it again?

tests/unit/datepicker/datepicker_events.js
fnagel: Because the focus is no longer on the input at this point.

tests/unit/calendar/calendar_core.js
fnagel: Again: is this still the case when the focus is moved by using simulated keyboard events?

tests/unit/calendar/calendar_test_helpers.js
jzaefferer: It should, so this is probably fine.

tests/unit/calendar/calendar_core.js
jzaefferer: Here the focus stays on the grid, doesn't it? Either way, a separate assertion that verifies where the focus should be could help.

fnagel: So you would like to add another test for verifying the focus is on the correct element? Like:

````
strictEqual( document.activeElement, assumedFocusedElement [ 0 ] );
````

jzaefferer: Yeah, where it makes sense. In this case the focus should stay on the grid, when triggering tab events, it shouldn't. Currently the tests aren't explicit about that.

fnagel: I guess that something for anther PR as we have a quite some more glitches to fix in the unit tests.

jzaefferer: Alright, that's fine by me. Can you make a note on the list? And go ahead and merge this.

fnagel: Done.

PR #1431, opened 2015-01-20
Datepicker: Fix visual tests

fnagel:
Basic fixes to make visual tests involving datepicker / calendar work again.


tests/visual/theme.html
```diff
@@ -52,16 +56,11 @@
 			return false;
 		});
 
-		// Datepicker
-		$('#datepicker').datepicker({
-			inline: true,
-			showWeek: true
```
jzaefferer: Why drop the `showWeek` option? We still have that, right?

fnagel: Indeed, re-added.

jzaefferer:
Haven't opened/tested anything, but diff looks good to me. Since there's so much more to address, lets land this now.

PR #1433, opened 2015-01-22
Sortable: cache original DOM position using all node types

mickylad:
Fixes [#10800 - Sortable: Cancelling a sort does not return the item to its original position with regard to non-element nodes](http://bugs.jqueryui.com/ticket/10800) by caching the original DOM position using `previousSibling` rather than `prev`, so all nodes are considered, not just elements.

Here's the original JSFiddle showing the bug:
http://jsfiddle.net/morchard/pups17xL/

And here's a JSFiddle showing the patch:
http://jsfiddle.net/morchard/z60ox3so/

### DIFF
--- a/tests/unit/sortable/sortable.html
+++ b/tests/unit/sortable/sortable.html
@@ -92,17 +92,17 @@
 	</tbody>
 </table>
 
-<ul id="sortable-with-text">
-	<li>Item 1</li>
+<div id="sortable-with-text">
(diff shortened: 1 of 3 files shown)


scottgonzalez:
Closing due to the ticket being closed. jQuery UI is not supposed to drop down to this level.

mickylad:
Ah, sorry about this, I didn't spot the existing issue.

PR #1434, opened 2015-01-22
Tests: Remove 1.6.x testing

mickylad:
Running all.html in any of the unit tests will fail because the 1.6.x jQuery versions can't be found. This removes them from the spec.


jzaefferer:
Thanks! I missed those when removing 1.6.x. I've squashed your two PRs into one and added references to the original ticket and the previous PR.

mickylad:
No problem, glad I could help :)

PR #1436, opened 2015-01-22
Theme update

jzaefferer:
This replaces #1420, which replaced #1384. Its the same branch for now, but I pushed it to this repo, so that Jasper can contribute to it directly.

I've been testing the new theme with Rafael in ThemeRoller and the custom demo that Download Builder creates. We noticed that the background of `.ui-widget-overlay` is darker than the `box-shadow` of `.ui-widget-shadow`, causing the shadow to disappear completely when on top of the overlay. In addition, the overlay doesn't need a background image, it works fine with just a color value. So our suggestion is to remove the `url()` from the overlay background, and change the color for overlay or shadow, or both, to work together.

@jaspermdegroot would be great if you could look into the above. You could push directly to this branch (let me know if you don't have access, I can fix that).


jzaefferer:
Also rebased this, so that we get the menu state update.

jaspermdegroot:
@jzaefferer 

I will look into overlay and shadow colors/opacity. Removing the background image from overlay sounds good.

I don't have access to the jquery-ui repo, but I can do a PR against the theme-update branch.

jzaefferer:
I just added you to the UI team on GitHub, that way you can just push to this branch, we don't need even more PRs. Just don't force-push to master ;-)

jzaefferer:
@jaspermdegroot thanks for the update. Since there are no notifications for commits, can you comment here when adding something?

Otherwise, what Rafael said. Bringing back the variables should make this good to land.

jaspermdegroot:
We are discussing "bringing back the comments/variables" on IRC. It will cause issues.

jzaefferer:
@jaspermdegroot do you know of anything missing here? I think this is ready to land.

jaspermdegroot:
@jzaefferer - I think it's ready too.

jzaefferer:
Squashed and merged. I've created [#10880](http://bugs.jqueryui.com/ticket/10880) to track the tooltip change, also fixed by this PR.

PR #1438, opened 2015-01-27
position of each items should be refreshed for sortable

iamdenny:
position of each items should be refreshed when div scroll up/down.
The previous code didn't refresh the position, so it calculate with wrong information of items when inner div's scroll up or down. 
I have tested it with more than 20 of 500 ~ 2000px items.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
 	"name": "jquery-ui",
 	"title": "jQuery UI",
 	"description": "A curated set of user interface interactions, effects, widgets, and themes built on top of the jQuery JavaScript Library.",
-	"version": "1.12.0-pre",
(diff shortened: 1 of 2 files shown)


scottgonzalez:
Thanks, please start with a ticket in http://bugs.jqueryui.com including a reduced test case showing the problem. We'll also want some unit tests showing that this fixes the problem and preventing regressions. You should read the [contributing guidelines](https://github.com/jquery/jquery-ui/blob/master/CONTRIBUTING.md) for full information.

iamdenny:
Ok. I'll do it after the project I'm working on is almost done.

scottgonzalez:
I'm going to close this for now, we can revisit when you have time to work on this again and file the bug report.

PR #1439, opened 2015-01-28
Dialog: Remove overflow: hidden and reposition resize handles

tjvantoll:
The `background-position: 16px 16px;` on the SE handle doesn't seem to be necessary now that the handle is small so I removed it.

I tested this on a few browsers on my Mac and some Windows VMs and it looks to be good. I still think these targets are too small, but by design this PR keeps the status quo.

### DIFF
--- a/themes/base/dialog.css
+++ b/themes/base/dialog.css
@@ -9,7 +9,6 @@
  * http://api.jqueryui.com/dialog/#theming
  */
 .ui-dialog {
-	overflow: hidden;
 	position: absolute;
 	top: 0;
 	left: 0;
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Looks good.

jzaefferer:
Will this cause problems if we try to make the handles larger, later? If not, I'm good with landing this now and creating a follow-up ticket to make the handles bigger. The latter should also take touch screens into account.

scottgonzalez:
Do you mean if we make the handles bigger in resizable or dialog? In either case, dialog has defined both dimensions and positions, so changes are always isolated. This doesn't change anything, honestly, other than the ability to remove `overflow: hidden`. The handles are the exact same size as they were before, and making them bigger after this change will have the same effect (positive or negative) as making them bigger without this change.

jzaefferer:
Alright, let's land this and create the follow-up ticket.

tjvantoll:
Done http://bugs.jqueryui.com/ticket/10897

PR #1440, opened 2015-01-29
Add css-dependency-comments based on jQuery Mobile

rxaviers:
Ref #1029 
Ref https://github.com/jquery/download.jqueryui.com/issues/178

### DIFF
--- a/ui/dialog.js
+++ b/ui/dialog.js
@@ -12,6 +12,9 @@
 //>>description: Displays customizable dialog windows.
 //>>docs: http://api.jqueryui.com/dialog/
 //>>demos: http://jqueryui.com/dialog/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/dialog.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/draggable.js
+++ b/ui/draggable.js
@@ -12,6 +12,7 @@
 //>>description: Enables dragging functionality for any element.
 //>>docs: http://api.jqueryui.com/draggable/
 //>>demos: http://jqueryui.com/draggable/
+//>>css.structure: ../themes/base/draggable.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/resizable.js
+++ b/ui/resizable.js
@@ -12,6 +12,9 @@
 //>>description: Enables resize functionality for any element.
 //>>docs: http://api.jqueryui.com/resizable/
 //>>demos: http://jqueryui.com/resizable/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/resizable.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/selectable.js
+++ b/ui/selectable.js
@@ -12,6 +12,7 @@
 //>>description: Allows groups of elements to be selected with the mouse.
 //>>docs: http://api.jqueryui.com/selectable/
 //>>demos: http://jqueryui.com/selectable/
+//>>css.structure: ../themes/base/selectable.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/sortable.js
+++ b/ui/sortable.js
@@ -12,6 +12,7 @@
 //>>description: Enables items in a list to be sorted using the mouse.
 //>>docs: http://api.jqueryui.com/sortable/
 //>>demos: http://jqueryui.com/sortable/
+//>>css.structure: ../themes/base/sortable.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/spinner.js
+++ b/ui/spinner.js
@@ -12,6 +12,9 @@
 //>>description: Displays buttons to easily input numbers via the keyboard or mouse.
 //>>docs: http://api.jqueryui.com/spinner/
 //>>demos: http://jqueryui.com/spinner/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/spinner.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/accordion.js
+++ b/ui/accordion.js
@@ -12,6 +12,9 @@
 //>>description: Displays collapsible content panels for presenting information in a limited amount of space.
 //>>docs: http://api.jqueryui.com/accordion/
 //>>demos: http://jqueryui.com/accordion/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/accordion.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/autocomplete.js
+++ b/ui/autocomplete.js
@@ -12,6 +12,9 @@
 //>>description: Lists suggested words as the user is typing.
 //>>docs: http://api.jqueryui.com/autocomplete/
 //>>demos: http://jqueryui.com/autocomplete/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/autocomplete.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/button.js
+++ b/ui/button.js
@@ -12,6 +12,9 @@
 //>>description: Enhances a form with themeable buttons.
 //>>docs: http://api.jqueryui.com/button/
 //>>demos: http://jqueryui.com/button/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/button.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/datepicker.js
+++ b/ui/datepicker.js
@@ -12,6 +12,9 @@
 //>>description: Displays a calendar from an input or inline for selecting dates.
 //>>docs: http://api.jqueryui.com/datepicker/
 //>>demos: http://jqueryui.com/datepicker/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/datepicker.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/menu.js
+++ b/ui/menu.js
@@ -12,6 +12,9 @@
 //>>description: Creates nestable menus.
 //>>docs: http://api.jqueryui.com/menu/
 //>>demos: http://jqueryui.com/menu/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/menu.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/progressbar.js
+++ b/ui/progressbar.js
@@ -12,6 +12,9 @@
 //>>description: Displays a status indicator for loading state, standard percentage, and other progress indicators.
 //>>docs: http://api.jqueryui.com/progressbar/
 //>>demos: http://jqueryui.com/progressbar/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/progressbar.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/selectmenu.js
+++ b/ui/selectmenu.js
@@ -12,6 +12,9 @@
 //>>description: Duplicates and extends the functionality of a native HTML select element, allowing it to be customizable in behavior and appearance far beyond the limitations of a native select.
 //>>docs: http://api.jqueryui.com/selectmenu/
 //>>demos: http://jqueryui.com/selectmenu/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/selectmenu.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/slider.js
+++ b/ui/slider.js
@@ -12,6 +12,9 @@
 //>>description: Displays a flexible slider with ranges and accessibility via keyboard.
 //>>docs: http://api.jqueryui.com/slider/
 //>>demos: http://jqueryui.com/slider/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/slider.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/tabs.js
+++ b/ui/tabs.js
@@ -12,6 +12,9 @@
 //>>description: Transforms a set of container elements into a tab structure.
 //>>docs: http://api.jqueryui.com/tabs/
 //>>demos: http://jqueryui.com/tabs/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/tabs.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
--- a/ui/tooltip.js
+++ b/ui/tooltip.js
@@ -12,6 +12,9 @@
 //>>description: Shows additional information for any element on hover or focus.
 //>>docs: http://api.jqueryui.com/tooltip/
 //>>demos: http://jqueryui.com/tooltip/
+//>>css.structure: ../themes/base/core.css
+//>>css.structure: ../themes/base/tooltip.css
+//>>css.theme: ../themes/base/theme.css
 
 (function( factory ) {
 	if ( typeof define === "function" && define.amd ) {


rxaviers:
Need someone to review the CSS references. Don't worry with the paths for now. It's important that all CSS files are being referenced and they are correct (each JS is correctly referencing its CSS'es).

ui/core.js
```diff
@@ -13,6 +13,9 @@
 //>>description: The core of jQuery UI, required for all interactions and widgets.
 //>>docs: http://api.jqueryui.com/category/ui-core/
 //>>demos: http://jqueryui.com/
+//>>css.structure: ../themes/base/base.css
```
jzaefferer: Why would we reference this anywhere? Its just a bunch of `@import` statements.

ui/core.js
```diff
@@ -13,6 +13,9 @@
...
+//>>css.structure: ../themes/base/core.css
```
jzaefferer: `core.js` and `core.css` are quite different. There's nothing in this JS file that actually depends on any CSS. I'd like to get input from @arschmitz on this, but as far as I can tell, `core.css` should be referenced as `css.structure` by all widgets that need it, but not here.

ui/widget.js
```diff
@@ -12,6 +12,7 @@
 //>>description: Provides a factory for creating stateful widgets with a common API.
 //>>docs: http://api.jqueryui.com/jQuery.widget/
 //>>demos: http://jqueryui.com/widget/
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Unlike `core.js`, there's actual usage for CSS classes in here, but then by itself, `$.Widget` doesn't really do anything. You can build a widget that never touches the DOM, so I don't think this reference should be here. Again, input from @arschmitz would be very welcome.

ui/dialog.js
```diff
@@ -12,6 +12,8 @@
 //>>description: Displays customizable dialog windows.
 //>>docs: http://api.jqueryui.com/dialog/
 //>>demos: http://jqueryui.com/dialog/
+//>>css.structure: ../themes/base/dialog.css
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Are CSS dependencies inherited from JS dependencies? In this case, does dialog depend on `button.css` due to the AMD dependency to `button.js`?

ui/draggable.js
```diff
@@ -12,6 +12,8 @@
 //>>description: Enables dragging functionality for any element.
 //>>docs: http://api.jqueryui.com/draggable/
 //>>demos: http://jqueryui.com/draggable/
+//>>css.structure: ../themes/base/draggable.css
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Afaik draggable doesn't have a dependency to `theme.css`.

ui/droppable.js
```diff
@@ -12,6 +12,7 @@
 //>>description: Enables drop targets for draggable elements.
 //>>docs: http://api.jqueryui.com/droppable/
 //>>demos: http://jqueryui.com/droppable/
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Same here, no actual dependency.

ui/effect-blind.js
```diff
@@ -12,6 +12,7 @@
 //>>description: Blinds the element.
 //>>docs: http://api.jqueryui.com/blind-effect/
 //>>demos: http://jqueryui.com/effect/
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Again, I don't think any of the effects have any CSS dependency at all. Going to skip the other `effect*` files.

ui/mouse.js
```diff
@@ -11,6 +11,7 @@
 //>>group: UI Core
 //>>description: Abstracts mouse-based interactions to assist in creating certain widgets.
 //>>docs: http://api.jqueryui.com/mouse/
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Also invalid

ui/position.js
```diff
@@ -14,6 +14,7 @@
 //>>description: Positions elements relative to other elements.
 //>>docs: http://api.jqueryui.com/position/
 //>>demos: http://jqueryui.com/position/
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Same here

ui/resizable.js
```diff
@@ -12,6 +12,8 @@
 //>>description: Enables resize functionality for any element.
 //>>docs: http://api.jqueryui.com/resizable/
 //>>demos: http://jqueryui.com/resizable/
+//>>css.structure: ../themes/base/resizable.css
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: I think this is the only interaction that needs the theme, since it uses icons.

ui/selectable.js
```diff
@@ -12,6 +12,8 @@
 //>>description: Allows groups of elements to be selected with the mouse.
 //>>docs: http://api.jqueryui.com/selectable/
 //>>demos: http://jqueryui.com/selectable/
+//>>css.structure: ../themes/base/selectable.css
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Nope

ui/sortable.js
```diff
@@ -12,6 +12,8 @@
 //>>description: Enables items in a list to be sorted using the mouse.
 //>>docs: http://api.jqueryui.com/sortable/
 //>>demos: http://jqueryui.com/sortable/
+//>>css.structure: ../themes/base/sortable.css
```
jzaefferer: Out of scope of this PR, but would be good to figure out: Both `draggable.css` and `sortable.css` are there simpy to define `touch-action: none;` for handles. We could move this to `core.css` as a generic class. @tjvantoll @scottgonzalez thoughts?

ui/sortable.js
```diff
@@ -12,6 +12,8 @@
...
+//>>css.theme: ../themes/base/theme.css
```
jzaefferer: Again, no actual dependency here

ui/core.js
rxaviers: True, removed by ed74d55

ui/core.js
rxaviers: Does any widget depend on `core.css`, but not on `core.js`?

ui/dialog.js
rxaviers: Yes, they are. In this case, yeap `button.css` will be included due to the AMD dependency to `button.js`.

ui/draggable.js
rxaviers: I'm waiting for confirmation...

ui/core.js
```diff
@@ -13,6 +13,8 @@
 //>>description: The core of jQuery UI, required for all interactions and widgets.
 //>>docs: http://api.jqueryui.com/category/ui-core/
 //>>demos: http://jqueryui.com/
+//>>css.structure: ../themes/base/core.css
```
jzaefferer: Comments got lost, so repeating them here. I said:

> `core.js` and `core.css` are quite different. There's nothing in this JS file that actually depends on any CSS. I'd like to get input from @arschmitz on this, but as far as I can tell, core.css should be referenced as css.structure by all widgets that need it, but not here.

Rafael asked:

> Does any widget depend on core.css, but not on core.js?

I think the answer is "no", but that still seems like its just abusing an accidental relationship. There is no direct relationship between `core.js` and `core.css`, its only through inheritance.

Hoping for input from @arschmitz here as well (again).

rxaviers:
@jzaefferer said:

> Out of scope of this PR, but would be good to figure out: Both draggable.css and sortable.css are there simpy to define touch-action: none; for handles. We could move this to core.css as a generic class. @tjvantoll @scottgonzalez thoughts?

I'm repeating his question here, because it will get lost as soon as I push an update.

ui/effect-blind.js
rxaviers: I'm waiting for a definitive confirmation in here as well.

ui/core.js
scottgonzalez: I'd say the answer will be a resounding yes, once http://bugs.jqueryui.com/ticket/9647 lands ;-)

scottgonzalez:
Regarding `touch-action: none;`, I think we'd want to wait a bit and see how it's being used in the wild, especially with Chassis.

ui/core.js
jzaefferer: That's a good point. Can we declare a dependency of `theme.css` to `core.css`? Either using a similar comment in `theme.css`, or making that implicit?

arschmitz: `core.css` should remain css dependency free in my opinion. In mobile its not actually a dependency and there is ( while not great ) use case here. Also I don't think anything in interactions depends on `core.css` but does on the `js`. We just always include core in mobile if your getting css at all. I can't seem to find where exactly this happens right this second, but ill track it down.

jzaefferer:
Wait and see is fine to me. With the (more) explicit dependencies in place, it'll be reasonable to change.

ui/widget.js
arschmitz: No this should not be here only things that actually depend on css should have a dependency on it . Individual widgets should depend on theme that actually use it its very very possible to have widgets that require no css at all ( though not common ). So this would include unnecessary files if someone was for example just trying to do a minimal build of `core.js` and `widget.js` for doing custom widget development or something.

ui/core.js
jzaefferer: Did you mean "`core.js` should remain css dependency free"? You wrote `core.css`.

Always including `core.css` when any CSS is included sounds fine to me, that's an implicit dependency, but since both structure and theme files depend on `core.css`, I don't see an issue with that.

ui/draggable.js
jzaefferer: @scottgonzalez can you confirm this, and for other interactions except resizable?

arschmitz:
So in general things that don't actually depend on a css file should not depend on it in comments here either I would say. Any other use kinda breaks the meaning of "dependency".

ui/core.js
arschmitz: @jzaefferer yes sorry `core.js` is indeed what i meant. In mobile core also actually includes most of the button css to avoid having most things also rely on button. However @tjvantoll and i discussed this a while back and both agree this should not be the case in UI and just things that depend on button css ( like new checkboxradio, selectmenu ) should include it.

jzaefferer: > selectmenu

As far as I can tell, selectmenu currently has no dependency on `button.css`. Is the button rewrite changing that?

arschmitz: Yes it is currently selectmenu basicly duplicates ( incorrectly and inconsistently ) the button css. The button re-write applies button css making them consistent with other buttons / form controls.

rxaviers: Summary,
- `core.js` must not list dependency on `core.css` or on any other CSS.
- The widgets will list its dependency on `core.css` themselves, e.g., dialog.js: `//>>css.structure: base/core.css, base/dialog.css`.
- There is no agreement on how button CSS should be defined yet.

All true?

jzaefferer: > core.js must not list dependency on core.css or on any other CSS.

Yes

> The widgets will list its dependency on core.css themselves

No, Alex said dependency on `core.css` is implicit with any other css dependency. 
But I'm also okay with specifying it explcitly.

> There is no agreement on how button CSS should be defined yet.

I don't think this is a problem. Since this looks like it can land before the button rewrite, updating the CSS dependencies can happen in the button PR after this lands.

rxaviers: > No, Alex said dependency on core.css is implicit with any other css dependency.

@arschmitz how must builder figure out `core.css` is a dependency if it's not listed in any `js` file?

ui/draggable.js
scottgonzalez: confirmed

ui/effect-blind.js
scottgonzalez: Confirmed, effects just use style manipulation directly, even highlight.

rxaviers:
Removed `theme.css` from all effects.

arschmitz:
> There is no agreement on how button CSS should be defined yet.

I don't think this is true at all actually. What I said was that we discussed, and any widget which depends on it, will include it explicitly. This is not yet an issues however as no widget does. It is only an issue once the re-write for button lands and i will update that PR once this one lands.

rxaviers:
> any widget (that) depends on (button.css) will include it explicitly ... once the re-write for button lands.

Ok.

ui/core.js
arschmitz: Ok so I looked into this more to figure out how exactly this is happening in mobile since it was non obvious. Some of the modules that make up our `core.js` actually depend on `core.css`. So in mobile we are essentially doing exactly as @rxaviers had originally suggested. However that aside I still don't think thats what we should do here. 

There are 3 ways i could see us handling this.
1) We could explicitly make any files that actually depend on it, depend on it explicitly ( list it in all the widgets under structure )
2) Since under the hood this is `r.js`, we can just shim it, to make ever other css file depend on it.
3) we could look at allowing dependency comments in the css files as well. Then add the dependency directly to the css files.

rxaviers: Ok, so based on your (1):
 - The widgets will list its dependency on `core.css` themselves, e.g., dialog.js: `//>>css.structure: base/core.css, base/dialog.css`.

On regard of having CSS files to list their CSS dependencies, we could definitely do that. But, I would postpone it (as a new feature / next step), since this is not done on mobile either and it seems as an improvement rather than a requirement.

arschmitz: > The widgets will list its dependency on core.css themselves, e.g., dialog.js: //>>css.structure: base/core.css, base/dialog.css.

For option one yes thats correct.

> On regard of having CSS files to list their CSS dependencies, we could definitely do that. But, I would postpone it (as a new feature / next step), since this is not done on mobile either and it seems as an improvement rather than a requirement.

Agreed, this would be an enhancement. I was only saying its one potential solution to handle this.

rxaviers: > Agreed, this would be an enhancement. I was only saying its one potential solution to handle this. 

:+1:

scottgonzalez: Let's just go with the explicit dependency for now.

arschmitz: +1 for option 1

I don't like option 2 becuase its too much "magic" to figure out how the file is actually being included.

1 is the simplest, easiest to understand, and easiest to implement. Seems like a clear winner to me for now.

jzaefferer: Sounds good. We can revisit that when/if we decide to specify dependencies inside CSS files.

rxaviers:
Based on the conversation we had so far, below is the CSS dependency list I understood for each `ui/*js`.

UI Core
- core.js: none
- widget.js: none
- mouse.js: none
- position.js: none

Interactions
- draggable.js: core.css, draggable.css
- resizable.js: core.css, resizable.css, theme.css
- selectable.js: core.css, selectable.css
- sortable.js: core.css, sortable.css

Widgets
- accordion.js: core.css, accordion.css, theme.css
- autocomplete.js: core.css, autocomplete.css, theme.css
- button.js: core.css, button.css, theme.css
- datepicker.js: core.css, datepicker.css
- datepicker.js: core.css, theme.css
- dialog.js: core.css, dialog.css, theme.css
- menu.js: core.css, menu.css, theme.css
- progressbar.js: core.css, progressbar.css, theme.css
- selectmenu.js: core.css, selectmenu.css, theme.css
- slider.js: core.css, slider.css, theme.css
- spinner.js: core.css, spinner.css, theme.css
- tabs.js: core.css, tabs.css, theme.css
- tooltip.js: core.css, tooltip.css, theme.css

Effects
- effect*: none

@scottgonzalez @jzaefferer @arschmitz please, can you confirm?

jzaefferer:
Droppable is missing, datepicker is listed twice.

Do interactions actually depend on `core.css`? I don't think so.

rxaviers:
> Droppable is missing

There's no droppable.css. Does it still depend on core.css? Or no CSS at all?

rxaviers:
> datepicker is listed twice.

I've updated my above comment removing the duplicate.

jzaefferer:
You've listed other components with "none", so it makes sense to list droppable as well. Afaik none of the interactions except resizable need `core.css`.

arschmitz:
@jzaefferer @rxaviers that had been my understanding on interactions as well. Only resizable needs it for the icon.

rxaviers:
Great, thank you @jzaefferer and @arschmitz. I've updated my above comment listing droppable and updating interactions.

jzaefferer:
Looks good to me

scottgonzalez:
draggable, selectable, and sortable are all missing their respective .css files.

rxaviers:
> draggable, selectable, and sortable are all missing their respective .css files.

Updated.

rxaviers:
Please, could you re-check?

scottgonzalez:
Looks correct.

rxaviers:
Great thanks

rxaviers:
Updated

rxaviers:
https://gist.github.com/rxaviers/01ab3bc43bf9f42c769f can be used to test building the CSS bundle.

ui/spinner.js
```diff
@@ -12,6 +12,8 @@
 //>>description: Displays buttons to easily input numbers via the keyboard or mouse.
 //>>docs: http://api.jqueryui.com/spinner/
 //>>demos: http://jqueryui.com/spinner/
+//>>css.structure: ../themes/base/spinner.css
```
jzaefferer: Missing core.css

rxaviers:
https://gist.github.com/rxaviers/01ab3bc43bf9f42c769f can be used to test building the CSS bundle.

jzaefferer:
Can you rebase this on master?

rxaviers:
Rebased

jzaefferer:
Looks like the structure.css file is just concatenated, that is, all the header comments are still there. Do we keep those and leave them to the minifier to remove?

rxaviers:
> Looks like the structure.css file is just concatenated, that is, all the header comments are still there. Do we keep those and leave them to the minifier to remove?

The scope of this PR is making sure CSS dependencies are being correctly defined. Although, that's a good question for https://github.com/jquery/download.jqueryui.com/issues/178 (or its PR https://github.com/jquery/download.jqueryui.com/pull/256). Currently, the individual banners are stripped out when they are concatenated. So, we can keep doing it unless we don't want to worry about it anymore.

jzaefferer:
Seems fine to ignore that. Unless someone else wants to review this as well, this is good to land.

rxaviers:
On today's meeting, the decision was to drop `theme.css` as a dependency for all JS components. Whether to include it or not will be a manual decision made by the developer/user completely independent of the used JS components. The structural CSSes are still kept (listed as dependencies for a JS component) though.

More details can be found on http://irc.jquery.org/%23jquery-meeting/default_%23jquery-meeting_20150211.log.html

rxaviers:
The new dependency list becomes:


UI Core
- core.js: none
- widget.js: none
- mouse.js: none
- position.js: none

Interactions
- draggable.js: draggable.css
- droppable: none
- resizable.js: core.css, resizable.css
- selectable.js: selectable.css
- sortable.js: sortable.css

Widgets
- accordion.js: core.css, accordion.css
- autocomplete.js: core.css, autocomplete.css
- button.js: core.css, button.css
- datepicker.js: core.css, datepicker.css
- dialog.js: core.css, dialog.css
- menu.js: core.css, menu.css
- progressbar.js: core.css, progressbar.css
- selectmenu.js: core.css, selectmenu.css
- slider.js: core.css, slider.css
- spinner.js: core.css, spinner.css
- tabs.js: core.css, tabs.css
- tooltip.js: core.css, tooltip.css

Effects
- effect*: none

jzaefferer:
> On today's meeting, the decision was to drop theme.css as a dependency for all JS components. 

I don't think that was the consensus. As I understood it: We keep the comments in the JS files. We do nothing with them, for now. We have the option to warn the user, using the DB UI, when they're selecting components that soft-depend on the theme while no theme is selected. The latter should be filed as a separate ticket.

scottgonzalez:
@jzaefferer Correct, that's what was decided.

rxaviers:
Ok, so https://github.com/jquery/jquery-ui/pull/1440#issuecomment-72333462 will be kept.

scottgonzalez:
So I think this is correct. @jzaefferer @arschmitz @rxaviers Can you all confirm?

rxaviers:
Confirmed. The changes are correct according to what we have defined.

jzaefferer:
:+1:

PR #1441, opened 2015-02-03
fixed tooltip appearing at [0,0] when the element disappears

amik6:
Sometimes with furious clicking elements with tooltips that appear on hover over master element, the element is not visible/is destroyed before the tooltip appears. Then, the tooltip appears on [0,0] and does not disappear any more. This check for visibility fixes this behavior.


scottgonzalez:
The scenario you've described is a bit confusing. Please start by reading the [contributing guidelines](https://github.com/jquery/jquery-ui/blob/master/CONTRIBUTING.md) and filing a bug report with a reduced test case.

PR #1446, opened 2015-02-08
Resizable: Check for 'auto' left/top

jrsearles:
When the left/top positions of an element are set to "auto" this ends up
evaluating to 0 which causes the element to jump to the left/top
incorrectly.

Fixes #9033

### DIFF
--- a/tests/unit/resizable/resizable.html
+++ b/tests/unit/resizable/resizable.html
@@ -44,6 +44,15 @@
 		height: 100px;
 		width: 100px;
 	}
+	#resizable3 {
+		position: absolute;
+		left: auto;
+		right: 0;
+		top: auto;
+		bottom: 0;
+		width: 400px;
+		height: 400px;
+	}
 	#container2 {
 		position: relative;
 		width: 400px;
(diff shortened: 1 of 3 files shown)


scottgonzalez:
Why not simplify the code to just always use `.position()`?

Also, please sign our [CLA](http://contribute.jquery.org/CLA/).

scottgonzalez:
Regarding my question of why we don't just always use `.position()`, looks like doing so would reintroduce a previous bug: a1b32b59d57c2171af9ba0ed6b35ccde28794e3b. Howver, that whole block was later removed: 3576ceb360eb0381a98f3c6b67d890c3834efa8a.

Since both changes include a test, we should be able to check all of the previous bugs again. I'd like someone else to chime in on this though.

jrsearles:
Thanks. I've signed the CLA.

jrsearles:
Is there anything else i need for this PR to be accepted?

scottgonzalez:
I was hoping @mikesherov would chime in since this is related to a1b32b59d57c2171af9ba0ed6b35ccde28794e3b.

scottgonzalez:

PR #1449, opened 2015-02-18
Tests: Reduce timeouts to optimize testsuite runtime

jzaefferer:
That makes several testsuite run quite a bit faster, sometimes shaving of several seconds (on my machine, menu tests took ~7s, now 4s). Mostly this was achieved by reducing timeouts. For menu I had to also reduce the timeout used within menu itself.

Having fast tests is good in general. When using browserstack-runner, things take longer, so the gains there are even bigger. Due to an [unrelated issue](https://github.com/browserstack/browserstack-runner/issues/109) I currently can't verify though, so this may need to wait a bit until I can test again.

### DIFF
--- a/tests/unit/autocomplete/autocomplete_core.js
+++ b/tests/unit/autocomplete/autocomplete_core.js
@@ -187,7 +187,7 @@ asyncTest( "past end of menu in multiline autocomplete", function() {
 		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
 		equal( element.text(), customVal );
 		start();
-	}, 50 );
+	});
 });
 
 asyncTest( "ESCAPE in multiline autocomplete", function() {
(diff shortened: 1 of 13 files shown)


scottgonzalez:
This looks fine to me *if* we get consistently passing tests in IE. I don't remember which versions of IE required the longer delays, so it's possible that we no longer need them. But I do know that the seemingly arbitrary long delays were from IE failures.

jzaefferer:
I managed to run some tests via browserstack-runner, and there are multiple failures. Will have to actually go through each suite in IEs...

jzaefferer:
Updated. Added a commit that reverts a few changes. Those are now properly tested across all supported browsers.

scottgonzalez:
If it's all passing, go ahead and merge :-)

jzaefferer:
Closed by 666f051c3a7c0313acacad0255ab8a45697777c3

PR #1452, opened 2015-02-20
Build: Use browserSets from testswarm config

Krinkle:
It's already in jQuery's Jenkins node-testswarm config
(and set to the same value) but not used yet.

Reference it to make sure it keeps working in the future.


Krinkle:
Will need to be cherry-picked to 1-9-stable, 1-10-stable, and 1-11-stable. (Like was done for 7c3b05cfeb4888708 last year.)

That way, if it ends up being changed again, one change in Jenkins will be enough.

scottgonzalez:
Cherry-picked to the mentioned branches as well.

Thanks.

PR #1455, opened 2015-02-23
Tabs: Use standard promise methods for jqXHR

scottgonzalez:
The old success(), error() and complete() methods have been deprecated for a
while and have been removed in upstream master.

### DIFF
--- a/ui/tabs.js
+++ b/ui/tabs.js
@@ -818,6 +818,18 @@ return $.widget( "ui.tabs", {
 			eventData = {
 				tab: tab,
 				panel: panel
+			},
+			complete = function( jqXHR, status ) {
+				if ( status === "abort" ) {
+					that.panels.stop( false, true );
+				}
+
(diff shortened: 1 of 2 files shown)


ui/tabs.js
```diff
@@ -843,7 +843,7 @@ return $.widget( "ui.tabs", {
 						that._trigger( "load", event, eventData );
 					}, 1 );
 				})
-				.complete(function( jqXHR, status ) {
+				.always(function( jqXHR, status ) {
```
dcherman: The arguments that the `always` function accepts is not the same as `complete`.  `jqXHR` is only the first argument in the error case;  in the success case, it's the last.

scottgonzalez: Thanks, updated.

scottgonzalez:
Ignore Travis, it keeps trying to pull the branch from the wrong repo and failing .

jzaefferer:
:+1:

PR #1053, opened 2013-08-10
2nd Reopen - Suggested fix for #8740

mziech:
Tooltip: Register event handlers before content is loaded. Fixes #8740 - Tooltip: Does not hide consistently with dynamically loaded content.

See http://bugs.jqueryui.com/ticket/8740

This pull request incorporates all suggested changes from #992 including the fixed commit message.

PR #1456, opened 2015-02-24
Tooltip: Register event handlers before content is loaded

scottgonzalez:
Fixes #8740
Closes gh-1053

This replaces gh-1053, which won't apply cleanly because of 8825d93dc877d182cf4a3fce37b6c2593cf08552. I've kept @mziech as the author since this is 90% his work. I'd like to get a review from him since it does contain some changes.

### DIFF
--- a/tests/unit/tooltip/tooltip_options.js
+++ b/tests/unit/tooltip/tooltip_options.js
@@ -77,14 +77,15 @@ asyncTest( "content: async callback loses focus before load", function() {
 
 	var element = $( "#tooltipped1" ).tooltip({
 		content: function( response ) {
-			element.trigger( "mouseleave" );
 			setTimeout(function() {
-				response( "sometext" );
+				element.trigger( "mouseleave" );
 				setTimeout(function() {
-					ok( !$( "#" + element.data( "ui-tooltip-id" ) ).is( ":visible" ),
-						"Tooltip should not display" );
-					element.tooltip( "destroy" );
-					start();
+					response( "sometext" );
+					setTimeout(function() {
+						ok( !$( "#" + element.data( "ui-tooltip-id" ) ).is( ":visible" ),
+							"Tooltip should not display" );
+						start();
+					});
 				});
 			});
 		}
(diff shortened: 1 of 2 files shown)


tests/unit/tooltip/tooltip_options.js
```diff
@@ -71,6 +71,27 @@ asyncTest( "content: sync + async callback", function() {
...
+					ok( !$( "#" + element.data( "ui-tooltip-id" ) ).is( ":visible" ),
+						"Tooltip should not display" );
+					element.tooltip( "destroy" );
+					start();
+				});
+			});
+		}
+	});
+	element.trigger( "mouseover" );
+});
```
scottgonzalez: This test was untouched.

ui/tooltip.js
```diff
@@ -331,7 +337,7 @@ return $.widget( "ui.tooltip", {
 		// tooltips will handle this in destroy.
 		if ( target[ 0 ] !== this.element[ 0 ] ) {
 			events.remove = function() {
-				this._removeTooltip( tooltip );
+				this._removeTooltip( this._find( target ).tooltip );
```
scottgonzalez: This line was changed from `this._find( target )` to `this._find( target ).tooltip` because we now store structured data, not just the tooltip element.

ui/tooltip.js
```diff
@@ -352,6 +358,12 @@ return $.widget( "ui.tooltip", {
 
 		// The tooltip may already be closed
 		if ( !tooltipData ) {
+
+			// We set ui-tooltip-open immediately upon open (in open()), but only set the
+			// additional data once there's actually content to show (in _open()). So even if the
+			// tooltip doesn't have full data, we always remove ui-tooltip-open in case we're in
+			// the period between open() and _open().
+			target.removeData( "ui-tooltip-open" );
```
scottgonzalez: This is the key thing that was needed after rebasing.

mziech:
Thank you for rebasing my changes on the current master!

Although I can hardly remember what was actually causing the problem, the changes look fine to me. However, tests are failing not only on Travis but locally as well. Maybe you want to have a look.

scottgonzalez:
Wonderful. The test passes in isolation, but not in the full suite. I'll dig in.

scottgonzalez:
The `mouseleave` event was being fired too early (I believe it was actually in the same loop as `mouseover`. Splitting that out into a different spin of the event loop fixed the failure.

PR #1458, opened 2015-02-27
Core: Drop feature detection for version check to support jQuery 1.7

scottgonzalez:
Fixes #11197

Just read the comment... :cry:

### DIFF
--- a/ui/core.js
+++ b/ui/core.js
@@ -142,7 +142,7 @@ $.extend( $.expr[ ":" ], {
 	}
 });
 
-// support: jQuery <1.8
+// support: jQuery 1.7 only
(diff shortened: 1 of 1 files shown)


ui/core.js
```diff
@@ -143,7 +143,16 @@ $.extend( $.expr[ ":" ], {
 });
 
 // support: jQuery <1.8
```
jzaefferer: Change this to "jQuery 1.7" (or whatever the right syntax is)?

scottgonzalez:
Updated support comment to "jQuery 1.7 only."

jzaefferer:
Looks good

PR #1459, opened 2015-03-02
Accordion: Handle `box-sizing: border-box` in animations

scottgonzalez:
Fixes #9264
Replaces gh-1287

I'm not sure we can write a sane unit test for this since the problem is that we're animating to the wrong size and then jumping to the correct size at the end. We could add a visual test that just uses `box-sizing: border-box` if people think that's necessary.


jzaefferer:
Seems fine as-is.

PR #1460, opened 2015-03-03
Update core.js

qingyuexiao:
Fix: error when element.nodeName is not string type.

### DIFF
--- a/ui/core.js
+++ b/ui/core.js
@@ -92,6 +92,9 @@ $.fn.extend({
 
 // selectors
 function focusable( element, isTabIndexNotNaN ) {
+	if(typeof element.nodeName != "string") {
+		return false;
+	}
 	var map, mapName, img,
 		nodeName = element.nodeName.toLowerCase();
 	if ( "area" === nodeName ) {


scottgonzalez:
When would this ever occur?

qingyuexiao:
When the element is a form and contains a form element such as input named 'nodeName', in this case element.nodeName is the input element.

arschmitz:
@qingyuexiao thats true for every property on a form element there is no way we can guard against this.

dmethvin:
There are *so many* cases like that though, so this doesn't really make it so that a property with the name `nodeName`. There are many similar failures in jQuery Core for the same reasons. See http://kangax.github.io/domlint/ .

scottgonzalez:
Indeed. Don't use names like that.

qingyuexiao:
I understand. However many people name the form elements according to the model built by backend developers, and they are not clear about those rules in dom & js. It will be difficult for them to find out why the ui doesn't work and they might turn to other ui controls. It's important to reduce user error. Just my personal opion~

PR #1463, opened 2015-03-03
Datepicker: add Tagalog locale

richguan:
Add Tagalog locale file to Datepicker


scottgonzalez:
Thanks, but we're not accepting any new locales since we're switching to CLDR.

richguan:
That's great news. Is there an eta for when the switch will be complete?

scottgonzalez:
v1.13.0, but we don't put dates on releases. You can track progress in the datepicker branch.

PR #1465, opened 2015-03-04
Slider: Modified to allow value reach max value with float step

atomiomi:
with step of 0.01 value did not reached max value

Fixes #11286

### DIFF
--- a/ui/slider.js
+++ b/ui/slider.js
@@ -558,7 +558,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		var max = this.options.max,
 			min = this._valueMin(),
 			step = this.options.step,
-				aboveMin = Math.floor( ( +( max - min ).toFixed(5) ) / step ) * step;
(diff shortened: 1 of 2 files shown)


ui/slider.js
```diff
@@ -558,7 +558,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		var max = this.options.max,
 			min = this._valueMin(),
 			step = this.options.step,
-			aboveMin = Math.floor( ( max - min ) / step ) * step;
+				aboveMin = Math.floor( ( +( max - min ).toFixed(5) ) / step ) * step;
```
scottgonzalez: Let's not use arbitrary precision. See https://github.com/jquery/jquery-ui/blob/18e301f4e29c2080e9aa9dac87c00dee137cb6c9/ui/spinner.js#L328.

scottgonzalez:
Merged with some style updates. Thanks.

PR #1466, opened 2015-03-04
Build: Don't use .min.js extension for individual source files

scottgonzalez:
We don't actually use these files for anything other than size comparisons,
but having the .min.js extension means that AMD is broken.

Note: If you're using AMD with the minified files, just run a build instead.

Fixes #10674

@rxaviers Can you just re-confirm that these file aren't used by download builder?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -79,7 +79,7 @@ var
 	component = grunt.option( "component" ) || "**";
 
 function mapMinFile( file ) {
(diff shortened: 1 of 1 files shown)


rxaviers:
@scottgonzalez correct. Download Builder generates the minified files. Those are not used.

tjvantoll:
:+1: This is much simpler than I thought it would be.

scottgonzalez:
> This is much simpler than I thought it would be.

That's because I took the simple way out ;-) The hard way would be stripping the wrappers and then saying "HA, for real don't use AMD on minified files." :smirk:

PR #1467, opened 2015-03-04
Selectmenu: Support `width: false` and default to 14em

scottgonzalez:
`width: null` still matches the width of the original element.
`width: false` prevents an inline style from being set for the width. This
makes it easy to set the width via a stylesheet and allows the use of any
unit for setting the width, such as the new default of `14em`.

Fixes #11198

### DIFF
--- a/ui/selectmenu.js
+++ b/ui/selectmenu.js
@@ -48,7 +48,7 @@ return $.widget( "ui.selectmenu", {
 			at: "left bottom",
 			collision: "none"
 		},
-		width: null,
+		width: false,
 
 		// callbacks
 		change: null,
@@ -118,7 +118,9 @@ return $.widget( "ui.selectmenu", {
 		this.buttonItem = this._renderButtonItem( item )
 			.appendTo( this.button );
 
-		this._resizeButton();
+		if ( this.options.width !== false ) {
+			this._resizeButton();
+		}
 
 		this._on( this.button, this._buttonEvents );
 		this.button.one( "focusin", function() {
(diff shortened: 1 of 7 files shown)


scottgonzalez:
PR for API change: https://github.com/jquery/api.jqueryui.com/pull/255

sandygettings:
Scott: The *width* property set null matches the width of the original element. The original issue (#11198) is that the default width is insufficient, because the triangle icon take up part of the width. I don't think this fixes #11198.

scottgonzalez:
> The original issue (#11198) is that the default width is insufficient

Right, so we changed the default.

sandygettings:
Ah ha, I didn’t understand that. Thanks!

ui/selectmenu.js
```diff
@@ -210,7 +212,7 @@ return $.widget( "ui.selectmenu", {
 				this._getSelectedItem().data( "ui-selectmenu-item" ) || {}
 			)
 		);
-		if ( !this.options.width ) {
+		if ( this.options.width === null ) {
```
jzaefferer: Shouldn't this check be consistent with the one above, which uses `!== false`?

scottgonzalez: The check above is inside `_create()`; the logic is: if we have any size to set, then do so.

This check is inside `refresh()`. If there is no inline style (`false`) or there's an explicit width (anything else beside `null`), then there's no action to take because the state will already be correct. If the value is `null` and we're copying the size of the original element, then we need to resize the button because that's something external that can change the behavior.

tjvantoll:
:+1:

jzaefferer:
Looks good

PR #1468, opened 2015-03-06
Fix missing wrappers

jzaefferer:
Noticed these while working on other style fixes. Esformatter won't address these, so might as well fix them now.


scottgonzalez:
Looks good.

PR #1469, opened 2015-03-06
Build: Style fixes

jzaefferer:
Kind of a preview for style fixes to come. Most spacing changes reflect the updated style guide. The `monthNames` array change looks correct to me (one level of indent for multi-var declaration, one for multi-line array literal). The `testswarm.createClient` indent changes are also correctly indenting the nested calls.

### DIFF
--- a/build/release.js
+++ b/build/release.js
@@ -9,7 +9,7 @@ function replaceAtVersion() {
 	var matches = [];
 
 	function recurse( folder ) {
-		fs.readdirSync( folder ).forEach(function( fileName ) {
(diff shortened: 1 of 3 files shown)


scottgonzalez:
I didn't check if there are any changes that should have occurred that didn't, but every change that did occur looks correct to me.

jzaefferer:
I'm using an updated jscs version to check for regressions (or missed updates) as well. Since we can't really land that update without fixing everything, I'm not including the jscs update here...

PR #1478, opened 2015-03-09
Core: Add methods to work around IE active element bugs

scottgonzalez:
Discussed with @tjvantoll and @arschmitz in IRC. These methods will not be documented and will not be supported; they're only for internal use.

https://github.com/jquery/jquery-mobile/pull/7927 should be updated if/when this lands.

### DIFF
--- a/ui/core.js
+++ b/ui/core.js
@@ -51,6 +51,7 @@ $.extend( $.ui, {
 		UP: 38
 	},
 
+	// Internal use only
 	safeActiveElement: function( document ) {
 		var activeElement;
 
@@ -65,6 +66,7 @@ $.extend( $.ui, {
 		return activeElement;
 	},
 
+	// Internal use only
 	safeBlur: function( element ) {
 
 		// Support: IE9 - 10 only
--- a/ui/core.js
+++ b/ui/core.js
@@ -49,6 +49,29 @@ $.extend( $.ui, {
 		SPACE: 32,
 		TAB: 9,
 		UP: 38
+	},
+
+	safeActiveElement: function( document ) {
+		var activeElement;
+
+		// Support: IE 9 only
+		// IE9 throws an "Unspecified error" accessing document.activeElement from an <iframe>
+		try {
+			activeElement = document.activeElement;
+		} catch ( error ) {
+			activeElement = document.body;
+		}
+
+		return activeElement;
+	},
+
+	safeBlur: function( element ) {
+
+		// Support: IE9 - 10 only
+		// If the <body> is blurred, IE will switch windows, see #9420
+		if ( element && element.nodeName.toLowerCase() !== "body" ) {
+			$( element ).blur();
+		}
 	}
 });
 
--- a/ui/autocomplete.js
+++ b/ui/autocomplete.js
@@ -286,7 +286,7 @@ $.widget( "ui.autocomplete", {
 					previous = this.previous;
 
 				// only trigger when focus was lost (click on menu)
-				if ( this.element[ 0 ] !== this.document[ 0 ].activeElement ) {
+				if ( this.element[ 0 ] !== $.ui.safeActiveElement( this.document[ 0 ] ) ) {
 					this.element.focus();
 					this.previous = previous;
 					// #6109 - IE triggers two focus events and the second
--- a/ui/dialog.js
+++ b/ui/dialog.js
@@ -191,8 +191,7 @@ return $.widget( "ui.dialog", {
 	enable: $.noop,
 
 	close: function( event ) {
-		var activeElement,
-			that = this;
+		var that = this;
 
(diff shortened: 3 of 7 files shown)


ui/core.js
```diff
@@ -49,6 +49,29 @@ $.extend( $.ui, {
...
+		}
+
+		return activeElement;
+	},
+
+	safeBlur: function( element ) {
+
+		// Support: IE9 - 10 only
+		// If the <body> is blurred, IE will switch windows, see #9420
+		if ( element && element.nodeName.toLowerCase() !== "body" ) {
```
gabrielschulhof: In mobile we also make sure that we avoid blurring the window. In fact, the code for retrieving the window and the document from the element is copied from the widget factory :)
https://github.com/jquery/jquery-mobile/pull/7927/files#diff-103715cf0e3db43f218c7235fd2a18fbR213

scottgonzalez: I saw that, it seemed really whacky. When are you ever passing a window?

jzaefferer:
You always pass `this.document[ 0 ]` to `$.ui.safeActiveElement`, that leads to two potential optimizations: Pass `this.document` and unwrap inside the method. Or add this method to `$.Widget.prototype`, removing the argument altogether, since you can just use `this.document[ 0 ]` directly.

> These methods will not be documented and will not be supported; they're only for internal use.

Let's add source comments for that.

scottgonzalez:
> You always pass this.document[ 0 ] to $.ui.safeActiveElement, that leads to two potential optimizations: Pass this.document and unwrap inside the method.

I wanted these methods to be as close to the intended APIs as possible. In the real APIs, we're always using DOM elements.

Or add this method to $.Widget.prototype, removing the argument altogether, since you can just use this.document[ 0 ] directly.

This will be used outside of widgets in jQuery Mobile. I should note that there are references to `document.activeElement` inside Effects Core, but I didn't want to add a dependency on UI Core and we've never had a bug report, so I figured I'd just leave them for now.

scottgonzalez:
> Let's add source comments for that.

Done.

jzaefferer:
I just looked at the jQuery Mobile PR, but I didn't see any calls to the equivalent helpers outside of widgets. Both `pagecontainer` and `popup` are widgets. Are there other uses not covered by the PR? If so, this should be good to land as-is.

scottgonzalez:
I'm not sure, I was going by what @arschmitz said. I can say, though, that if we had already split UI Core into multiple files, I would have used this in Effects Core.

arschmitz:
Sorry the page container call used to be in navigation it was moved I forgot. The effects use seems valid though.

tjvantoll:
We should've done this a long time ago. :+1:

ui/core.js
gabrielschulhof: I believe the reason for that is that if you ever pass ```event.target``` or the context of an event handler you may end up passing in the window. I don't believe ```document.activeElement``` can ever be set to the window. I guess the basic assumption was that ```$( window ).blur()``` is unsafe in the same way as ```$( "body" ).blur()``` in that it results in the window losing focus. However, I've now tested ```$( window ).blur()``` as far back as IE6 and it does not seem to do anything, so I guess attempting to blur the window is safe.

I used the following test:

```JS
<!doctype html>
<html>
<head>
	<script src="http://code.jquery.com/jquery-1.11.1.js"></script>
	<script>
setTimeout( function() {
	$( "#console" ).text( "blurring window\n" );
	$( window ).blur();
}, 5000 );
	</script>
</head>
<body>
	<pre><code id="console"></code></pre>
	<input>
</body>
</html>
```

and I tried both with focus initially set to default and with focus immediately set on the input by my clicking on it.

gabrielschulhof:
Actually, it might be simpler to override core's ```$.event.special.focus``` and ```$.event.special.blur```. That would remove the need for introducing new API, and it would automagically work everywhere.

gabrielschulhof:
Sorry, I just wanted to throw this in there, because I don't have the time right now to actually look in more detail, but I noticed that focus and blur are implemented in core as .trigger( "focus" ) and .trigger( "blur" ) and those events, in turn, are handled in $.event.special, so they are exposed to being overridden.

Also, core already has a private method safeActiveElement() which does the try {} catch(){} thing, so we need not duplicate that if we go the $.event.special route.

gabrielschulhof:
We could override https://github.com/jquery/jquery/blob/master/src/event.js#L558-L576 and add the body-check (*facepalm*) into blur.

scottgonzalez:
Overriding `$.event.special.focus` in no way helps us with getting the active element. No matter what we do, we have to create a `safeActiveElement()` method.

Overriding `$.event.special.blur` is not an acceptable solution. Just because we don't want to trigger a blur on the document doesn't mean we should make it impossible to do so for all users.

gabrielschulhof:
@scottgonzalez That makes sense.

PR #1479, opened 2015-03-10
Slider: Fix to restrict handle values when set programmatically.

dekajp:
Fixes #3762 - Slider: handles not restricted properly when set programmatically

### DIFF
--- a/tests/unit/slider/methods.js
+++ b/tests/unit/slider/methods.js
@@ -161,4 +161,53 @@ test( "values, multi step", function() {
 	element.slider( "destroy" );
 } );
 
+test( "values", function() {
+	expect( 12 );
+
+	var element = $( "<div></div>" ).slider( {
+		range: true,
+		min: 10,
+		max: 100,
+		step: 1
+	} );
+
+	element.slider( "values", [ 20, 90 ] );
+	deepEqual( element.slider( "values" ), [ 20, 90 ], "Values (array) - get value for all handles" );
+
+	element.slider( "values", [ 0, 200 ] );
+	deepEqual( element.slider( "values" ), [ 10, 100 ], "Values (array) - restricted min and max" );
+
+	element.slider( "values", 0, 50 );
+	equal( element.slider( "values", 0 ), 50, "Values (index,value) - set first handle" );
+
+	element.slider( "values", 1, 75 );
+	equal( element.slider( "values", 1 ), 75, "Values (index,value) - set second handle" );
+
+	element.slider( "values", 0, 5 );
+	equal( element.slider( "values", 0 ), 10, "Values (index,value) - 1st handle restricted properly, against min" );
+
+	element.slider( "values", 0, 90 );
+	equal( element.slider( "values", 0 ), element.slider( "values", 1 ), "Values (index,value) - 1st handle restricted properly, against 2nd handle" );
+
+	element.slider( "values", 1, 110 );
+	equal( element.slider( "values", 1 ), 100, "Values (index,value) - 2nd handle restricted properly, against max" );
+
+	element.slider( "values", 1, 40 );
+	equal( element.slider( "values", 1 ), element.slider( "values", 0 ), "Values (index,value) - 2nd handle restricted properly, against 1st handle" );
+
+	element.slider( "values", [ 50, 75 ] );
+
+	element.slider( "values", [ 0, 75 ] );
+	deepEqual( element.slider( "values" ), [ 10, 75 ], "Values (array) - Setting both handles - against  min" );
+
+	element.slider( "values", [ 0, 90 ] );
+	deepEqual( element.slider( "values" ), [ 10, 90 ], "Values (array) - 1st handle restricted properly, against 2nd handle" );
+
+	element.slider( "values", 1, 110 );
+	deepEqual( element.slider( "values" ), [ 10, 100 ], "Values (array) - 2nd handle restricted properly, against max" );
+
+	element.slider( "values", [ 75, 50 ] );
+	deepEqual( element.slider( "values" ), [ 75, 75 ], "Values (array) - Setting both handles - 2nd Handle resets to 1st" );
+} );
+
 } );
--- a/tests/unit/slider/options.js
+++ b/tests/unit/slider/options.js
@@ -238,18 +238,34 @@ test( "step", function() {
 //});
 
 test( "values", function() {
-	expect( 2 );
+	expect( 6 );
 
 	// Testing multiple ranges on the same page, the object reference to the values
 	// property is preserved via multiple range elements, so updating options.values
 	// of 1 slider updates options.values of all the others
+
 	var ranges = $( [
-		document.createElement( "div" ),
-		document.createElement( "div" )
-	] ).slider( {
-		range: true,
-		values: [ 25, 75 ]
-	} );
+				document.createElement( "div" ),
+				document.createElement( "div" )
+			] ).slider( {
+				range: true,
+				values: [ 25, 75 ]
+			} ),
+			element = $( "<div></div>" ).slider( {
+				range: true,
+				min: 10,
+				max: 100,
+				step: 1,
+				values: [ 15, 1 ]
+			} );
+
+	ranges = $( [
+			document.createElement( "div" ),
+			document.createElement( "div" )
+		] ).slider( {
+			range: true,
+			values: [ 25, 75 ]
+		} );
 
 	notStrictEqual(
 		ranges.eq( 0 ).slider( "instance" ).options.values,
@@ -264,11 +280,24 @@ test( "values", function() {
 		ranges.eq( 1 ).slider( "values", 0 ),
 		"the values for multiple sliders should be different"
 	);
+
+	deepEqual( element.slider( "values" ), [ 15, 15 ], "handles restricted properly when set programmatically - 2nd handle reset" );
+
+	element.slider( "option", "values", [ 0, 200 ] );
+	deepEqual( element.slider( "values" ), [ 10, 100 ], "option values - restricted min and max" );
+
+	element.slider( "option", "values", [ 15, 1 ] );
+	deepEqual( element.slider( "values" ), [ 15, 15 ], "option values - restricted 2nd Handle" );
+
+	element.slider( "option", "values", [ 15, 15 ] );
+	element.slider( "option", "values", [ 25, 15 ] );
+	deepEqual( element.slider( "values" ), [ 25, 25 ], "option values - restricted 2nd Handle" );
+
 } );
 
 test( "range", function( assert ) {
 	expect( 32 );
-	var range;
+	var element,range;
 
 	// Min
 	element = $( "<div></div>" ).slider( {
--- a/ui/widgets/slider.js
+++ b/ui/widgets/slider.js
@@ -76,6 +76,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		this._detectOrientation();
 		this._mouseInit();
 		this._calculateNewMax();
+		this._cleanValues();
 
 		this._addClass( "ui-slider ui-slider-" + this.orientation,
 			"ui-widget ui-widget-content" );
@@ -157,6 +158,41 @@ return $.widget( "ui.slider", $.ui.mouse, {
 				this.range.remove();
 			}
 			this.range = null;
+		}
+	},
+
+	_cleanValues: function( index, newValue ) {
+		var values;
+
+		if ( this.options.range === true ) {
+			if ( arguments.length === 2 ) {
+				values = this.options.values;
+				if ( index === 0 && newValue > values[ 1 ] ) {
+					newValue = values[ 1 ];
+				}else if ( index === 1 && newValue < values[ 0 ] ) {
+					newValue = values[ 0 ];
+				}
+				return newValue;
+			}
+
+			values = arguments[ 0 ] || this.options.values;
+			if ( values ) {
+				if ( values[ 0 ] > values[ 1 ] ) {
+					values[ 1 ] = values[ 0 ];
+				}
+
+				if ( !arguments.length ) {
+					this.options.values = values;
+				}
+
+				return values;
+			}
+		} else {
+			if ( $.isArray( arguments[ 0 ] ) ) {
+				return arguments[ 0 ];
+			} else {
+				return newValue;
+			}
 		}
 	},
 
@@ -385,6 +421,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 
 		if ( arguments.length > 1 ) {
 			this.options.values[ index ] = this._trimAlignValue( newValue );
+			this.options.values[ index ] = this._cleanValues( index, newValue );
 			this._refreshValue();
 			this._change( null, index );
 			return;
@@ -393,7 +430,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		if ( arguments.length ) {
 			if ( $.isArray( arguments[ 0 ] ) ) {
 				vals = this.options.values;
-				newValues = arguments[ 0 ];
+				newValues = this._cleanValues( arguments[ 0 ] );
 				for ( i = 0; i < vals.length; i += 1 ) {
 					vals[ i ] = this._trimAlignValue( newValues[ i ] );
 					this._change( null, i );
@@ -452,6 +489,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 				break;
 			case "values":
 				this._animateOff = true;
+				this._cleanValues();
 				this._refreshValue();
 
 				// Start from the last handle to prevent unreachable handles (#9046)


ui/slider.js
```diff
@@ -157,6 +158,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		}
 	},
 
+	_cleanValues: function( index, newValue ) {
+		var i = 0, values ;
```
scottgonzalez: Uninitialized variables go at the top:

```js
var values,
    i = 0;
```

ui/slider.js
```diff
@@ -157,6 +158,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+			if ( index === 0 && newValue > values[ index + 1 ] ) {
+				return values[ index + 1 ];
+			}
+			if ( index === 1 && newValue < values[ index - 1 ] ) {
+				return values[ index - 1 ];
+			}
+			return newValue;
+		}
+
+		if ( arguments.length ) {
```
scottgonzalez: Am I missing something here? There are multiple calls to this method with no parameters, but we do nothing in that case?

ui/slider.js
```diff
@@ -157,6 +158,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+
+		if ( arguments.length > 1 ) {
+			values = this.options.values;
+			if ( index === 0 && newValue < this._valueMin() ){
+				newValue = this._valueMin();
+			}
+			if ( index === 1 && newValue > this._valueMax() ){
+				newValue = this._valueMax();
+			}
+			if ( index === 0 && newValue > values[ index + 1 ] ) {
```
scottgonzalez: This logic should only apply with ranges.

scottgonzalez:
There are some spacing issues and I didn't do a full review because it looks like this still needs some changes to how it works.

ui/slider.js
dekajp: My bad , I missed them.

jzaefferer:
@dekajp please comment on your PRs after you push an update, since the push itself has no notification.

@scottgonzalez there's a new commit since yesterday

ui/slider.js
```diff
@@ -422,6 +465,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 				this._refreshValue();
 			} else {
 				if ( this.options.values && this.options.values.length ) {
+					this.options.values = this._cleanValues( this.options.values );
```
scottgonzalez: Why are the values being cleaned during a get?

ui/slider.js
```diff
@@ -157,6 +157,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		}
 	},
 
+	_cleanValues: function( index, newValue ) {
+		var values,
+			i = 0 ;
+
+		if ( $.isArray( arguments[ 0 ] ) ) {
+			values = arguments[ 0 ] || this.options.values;
```
scottgonzalez: How would this ever be `this.options.values` since `arguments[ 0 ]` is guaranteed to be an array?

ui/slider.js
```diff
@@ -157,6 +157,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		}
 	},
 
+	_cleanValues: function( index, newValue ) {
+		var values,
+			i = 0 ;
```
scottgonzalez: extraneous space

ui/slider.js
```diff
@@ -157,6 +157,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+
+		if ( $.isArray( arguments[ 0 ] ) ) {
+			values = arguments[ 0 ] || this.options.values;
+			if ( values ) {
```
scottgonzalez: Why does this check exist?

ui/slider.js
```diff
@@ -157,6 +157,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+				for ( ; i < values.length; i++ ){
```
scottgonzalez: Missing space before the brace.

ui/slider.js
```diff
@@ -157,6 +157,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+					if ( values[ i ] < this._valueMin() ) {
+						values[ i ] = this._valueMin();
+					}
+
+					if ( values[ i ]  > this._valueMax() ) {
+						values[ i ] = this._valueMax();
+					}
+				}
+				if ( values[ 0 ] > values[ 1 ] ) {
```
scottgonzalez: This is wrong for non-range sliders.

ui/slider.js
```diff
@@ -157,6 +157,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+			}
+		}
+
+		if ( arguments.length > 1 ) {
+			values = this.options.values;
+			if ( index === 0 && newValue < this._valueMin() ){
+				newValue = this._valueMin();
+			}else if ( index === 1 && newValue > this._valueMax() ){
+				newValue = this._valueMax();
+			}
```
scottgonzalez: Why is this only half of the min/max logic?

ui/slider.js
```diff
@@ -157,6 +157,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+
+			if ( this.options.range === true && values.length === 2 ) {
```
scottgonzalez: The length check isn't necessary. Range sliders must have exactly two values.

ui/slider.js
```diff
@@ -157,6 +157,48 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+				if ( index === 0 && newValue > values[ index + 1 ] ) {
+					newValue = values[ index + 1 ];
+				}else if ( index === 1 && newValue < values[ index - 1 ] ) {
+					newValue = values[ index - 1 ];
+				}
```
scottgonzalez: Why the `index +/- 1` logic? You already have the indices hard-coded anyway out of necessity.

scottgonzalez:
Maybe we should start with just a set of tests so that we can agree the correct behavior is being tested.

ui/slider.js
dekajp: I will change this and move `_cleanValues` to create

dekajp:
do you want me to add/modify the unit test cases ?

ui/slider.js
dekajp: If `range = true`   `min <= X0 <= X1` and  `X0 <= X1<= Max`

dekajp:
From http://bugs.jqueryui.com/ticket/3762#comment:31
1. Restricting the value of a handle based on other `handles only applies with range: true` and therefore also only applies when exactly two handles exist. 
2. When setting both values at the same time, either via .values( array ), .option( "values", array ), or initialization,
       1.  the first handle can be any value between min and max and
       2.  the second handle must be between the first handle and max. 
3. When setting a single handle, it is restricted by the min, max, and existing value of the other handle.

dekajp:
@scottgonzalez  , i will send a new PR  on this.

tests/unit/slider/slider_methods.js
```diff
@@ -100,8 +100,90 @@ test( "value", function() {
...
+	element.slider( "values", [ 0, 200 ] );
+	deepEqual( element.slider( "values" ), [ 10, 100 ], "Values (array) - restricted min and max" );
+
+	element.slider( "values", 0, 50 );
+	equal( element.slider( "values", 0 ), 50, "Values (index,value) - set first handle" );
+
+	element.slider( "values", 1, 75 );
+	equal( element.slider( "values", 1 ), 75, "Values (index,value) - set second handle" );
+
+	element.slider( "values", [ 50, 75 ] );
```
scottgonzalez: This is effectively a no-op. Why is it here?

tests/unit/slider/slider_methods.js
```diff
@@ -100,8 +100,90 @@ test( "value", function() {
...
+	element.slider( "values", 0, 5 );
+	equal( element.slider( "values", 0 ), 10, "Values (index,value) - 1st handle restricted properly, against min" );
+
+	element.slider( "values", 1, 5 );
+	equal( element.slider( "values", 0 ), 10, "Values (index,value) - 2nd handle restricted properly, against min" );
+
+	element.slider( "values", [ 50, 75 ] );
+	element.slider( "values", 0, 110 );
+	equal( element.slider( "values", 0 ), 75, "Values (index,value) - 1st handle restricted properly, against max" );
```
scottgonzalez: It's restricted by the second handle, not the max.

tests/unit/slider/slider_methods.js
```diff
@@ -100,8 +100,90 @@ test( "value", function() {
...
+
+	element.slider( "values", 1, 110 );
+	equal( element.slider( "values", 1 ), 100, "Values (index,value) - 2nd handle restricted properly, against max" );
+
+	element.slider( "values", [ 50, 75 ] );
+	element.slider( "values", 0, 90 );
+	equal( element.slider( "values", 0 ), 75, "Values (index,value) - 1st handle restricted properly, against 2nd handle" );
```
scottgonzalez: This is already tested in L134.

tests/unit/slider/slider_methods.js
```diff
@@ -100,8 +100,90 @@ test( "value", function() {
...
+	element.slider( "values", [1.5, 1.3] );
+	deepEqual( element.slider( "values" ), [ 1.5, 1.5 ], "Values (array) - Floating Point - uneven step - 2nd Handle restricted" );
+
+	element.slider( "values", [1.5, 1.5] );
+	element.slider( "values", 1, 1.3 );
+	deepEqual( element.slider( "values" ), [ 1.5, 1.5 ], "Values (index,value) - Floating Point - uneven step - 2nd Handle restricted" );
+
+	element.slider( "values", [1.5, 1.5] );
+	element.slider( "values", 0, 1.6 );
+	deepEqual( element.slider( "values" ), [ 1.5, 1.5 ], "Values (index,value) - Floating Point - uneven step - 1st Handle restricted" );
```
scottgonzalez: Why are we doing a whole slew of floating point tests for boundary restrictions?

scottgonzalez:
I think the tests would be easier to follow if they were grouped by which handle is being tested. For example, testing the individual handle settings can be simplified to:

* Set the first handle below min, verify it goes to min.
* Set the first handle above the second handle, verify it goes to the second handle.
* Set the second handle above the max, verify it goes to the max.
* Set the second handle below the first handle, verify it goes to the first handle.

There's no need to reset the values in between because every individual assertion can be properly performed with a single value change. You can test the multi-value setter in a similar grouping.

tests/unit/slider/slider_methods.js
dekajp: element.slider( "values", 0, 110 );  ` Max = 100 `

element.slider( "values", 0, 90 ); ` Less than max but greater than 2nd handle`

scottgonzalez: But in both cases, the handle has to be restricted by the second handle.

dekajp: Alright

tests/unit/slider/slider_methods.js
dekajp: Well just making sure floating points are covered

scottgonzalez: That's not necessary.

dekajp:
@scottgonzalez  , I rebased my files. Now github is not showing here , asking me to send PR
Also `grunt` is giving lint error in html in master branch . 

https://github.com/dekajp/jquery-ui/commit/ec53b6d0ed7aaa5412fa3475f77a04ab1de0a4c7

scottgonzalez:
It looks like the branch names are different (the casing on the v2).

scottgonzalez:
@dekajp What's the status of this?

dekajp:
@scottgonzalez   last 3 weeks i was busy with relocation. i think this weekend i will be able to get back on this

scottgonzalez:
Thanks. Hope the move went well.

dekajp:
@scottgonzalez  yeah move went all , everyone happy in family :) with more space .  I rebased commits .

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+
+	element.slider( "values", 1, 75 );
+	equal( element.slider( "values", 1 ), 75, "Values (index,value) - set second handle" );
+
+	
+	element.slider( "values", 0, 5 );
+	equal( element.slider( "values", 0 ), 10, "Values (index,value) - 1st handle restricted properly, against min" );
+
+	element.slider( "values", 0, 90 );
+	equal( element.slider( "values", 0 ), element.slider( "values", 1 ), "Values (index,value) - 1st handle restricted properly, against 2nd handle" );
```
scottgonzalez: I understand that you're trying to show that the handles are matching, but I'd prefer an explicit value for the expected value.

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+
+	element.slider( "values", 1, 110 );
+	equal( element.slider( "values", 1 ), 100, "Values (index,value) - 2nd handle restricted properly, against max" );
+
+	element.slider( "values", 1, 40 );
+	equal( element.slider( "values", 1 ), element.slider( "values", 0 ), "Values (index,value) - 2nd handle restricted properly, against 1st handle" );
```
scottgonzalez: Same change here, use an explicit value.

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+	
+	element.slider( "values", [ 50, 75 ] );
```
scottgonzalez: This is still effectively a no-op and should be removed.

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+
+	element.slider( "values", [ 0, 75 ] );
+	deepEqual( element.slider( "values" ), [ 10, 75 ], "Values (array) - Setting both handles - against  min" );
```
scottgonzalez: You've already asserted this on L125.

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+
+	element.slider( "values", [ 0, 90 ] );
+	deepEqual( element.slider( "values" ), [ 10, 90 ], "Values (array) - 1st handle restricted properly, against 2nd handle" );
```
scottgonzalez: This is exactly the same as the previous assertion, and the message is wrong.

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+
+	element.slider( "values", 1, 110 );
+	deepEqual( element.slider( "values" ), [ 10, 100 ], "Values (array) - 2nd handle restricted properly, against max" );
```
scottgonzalez: You've already asserted this on L125.

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+	
+	element.slider( "values", [ 75, 50 ] );
+	deepEqual( element.slider( "values" ), [ 75, 75 ], "Values (array) - Setting both handles - 2nd Handle resets to 1st" );
```
scottgonzalez: I wouldn't use the word "resets" here; it's being restricted.

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+});
```
scottgonzalez: Let's add some assertions for non-range sliders as well, to verify the handle restrictions aren't applied.

tests/unit/slider/methods.js
```diff
@@ -109,8 +109,54 @@ test( "value", function() {
...
+	expect( 12 );
+	
+	var element = $( "<div></div>" ).slider({
+		range: true,
+		min: 10,
+		max: 100,
+		step: 1
+	});
+
+	element.slider( "values", [ 20, 90 ] );
```
scottgonzalez: Set the value in the initialization.

tests/unit/slider/options.js
```diff
@@ -264,6 +265,26 @@ test( "values", function() {
...
+	});
+	deepEqual( element.slider( "values" ), [ 15, 15 ], "handles restricted properly when set programmatically - 2nd handle reset" );
+
+	element.slider( "option", "values", [ 0, 200 ] );
+	deepEqual( element.slider( "values" ), [ 10, 100 ], "option values - restricted min and max" );
+
+	element.slider( "option", "values", [ 15, 1 ] );
+	deepEqual( element.slider( "values" ), [ 15, 15 ], "option values - restricted 2nd Handle" );
+
+	element.slider( "option", "values", [ 15, 15 ] );
```
scottgonzalez: This is effectively a no-op and should be removed.

ui/slider.js
```diff
@@ -160,6 +161,41 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+
+				if ( !arguments.length ){
+					this.options.values = values;
+				}
+
+				return values;
+			}
+		} else {
+			if ( $.isArray( arguments[ 0 ] ) ) {
+				return arguments[0];
```
scottgonzalez: Bad spacing.

ui/slider.js
```diff
@@ -160,6 +161,41 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
 
+	_cleanValues: function( index, newValue ) {
+		var values;
+
+		if ( this.options.range === true ){
+			if ( arguments.length === 2 ) {
+				values = this.options.values;
+				if ( index === 0 && newValue > values[ 1 ] ) {
+					newValue = values[ 1 ];
+				}else if ( index === 1 && newValue < values[ 0 ] ) {
```
scottgonzalez: bad spacing

ui/slider.js
```diff
@@ -160,6 +161,41 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+					newValue = values[ 0 ];
+				}
+				return newValue;
+			}
+
+			values = arguments[ 0 ] || this.options.values;
+			if ( values ) {
```
scottgonzalez: Why does this conditional exist?

ui/slider.js
```diff
@@ -160,6 +161,41 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+				if ( values[ 0 ] > values[ 1 ] ) {
+					values[ 1 ] = values[ 0 ];
+				}
+
+				if ( !arguments.length ){
```
scottgonzalez: Bad spacing.

ui/slider.js
```diff
@@ -160,6 +161,41 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
+					this.options.values = values;
```
scottgonzalez: It seems odd that this method is setting internal state in this one case.

scottgonzalez: I think it's just the way this method is organized that makes it seem odd.

tests/unit/slider/methods.js
dekajp: It is hard to follow so many test cases so i reset values , for easy readability.

scottgonzalez: But the next line of code resets values as well. The code isn't any easier to follow with this additional step.

tests/unit/slider/methods.js
jzaefferer: All of these were addressed by #1573, which landed in 4c5cd2018d88392871cfc93d3fb6f71a0bc510ff

jzaefferer:
@dekajp would you be available to rebase this and update the implementation?

dekajp:
@jzaefferer i rebased by commit. I will look into update the implementation

sameer05k21a0440:
I have dynamic multiple input and dynamic slider handle so when on change input text value need to set slider value but I can't set slider handle check below code .

```
 $(this).parents("tr").find('input[name="assesmentScore"]').change( function (event) {
	   					 var textValue = $(this).parents("tr").find('input[name="assesmentScore"]').val();
$("#bonusScore", $(this)).slider('value',textValue);    //here need to add dynamic slide how ? 
	   					handle.text(textValue); //Can't set corresponding text
});
```

Note :Please let me if needed full code.
Thanks

dmethvin:
Closing as a stale branch.

PR #1480, opened 2015-03-10
Tests: Use jquery-git1 instead of jquery-git

scottgonzalez:
We only test against jQuery 1.x, but we still have `jquery-git.js` as our WIP target. This changes our WIP target to `jquery-git1.js` so we're testing the right code.


jzaefferer:
Diff looks good, haven't checked for anything you may have missed.

scottgonzalez:
Merged and updated the Jenkins job to use `git1` instead of `git`.

PR #1482, opened 2015-03-10
Accordion: Fix tests in IE 11 with jQuery 1.7-1.8

scottgonzalez:
Hooray for async focus handling.


jzaefferer:
Diff looks "good", as good as addressing IE focus issues can get...

PR #1483, opened 2015-03-11
Component: Show the correct precision in spinner when the number of digi...

YiyangLi:
...ts are smaller; Example: https://jsfiddle.net/yli119/FNys7/108/ when you try to increment from 0.09 to 0.1, the spinner shows 0.1 but not 0.10. Fix: The parseFloat smash the .0 or .00, you might parse float firstly and then run toFixed right before returning. Thanks. http://yiyang.li

### DIFF
--- a/ui/spinner.js
+++ b/ui/spinner.js
@@ -352,17 +352,20 @@ return $.widget( "ui.spinner", {
 		// - rounding is based on 0, so adjust back to our base
 		value = base + aboveMin;
 
-		// fix precision from bad JS floating point math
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Please stop doing that.

YiyangLi:
@scottgonzalez I apologize if my frequent operations bother you. I am new to github and jQuery. We just encounter a UI problem on our product, which could be fixed by adjusting the jQuery UI. I want to contribute my effort to it, so I create the pull request. Sorry again for the spamming.

PR #1484, opened 2015-03-11
Component: Show the correct precision in spinner when the number of digi...

YiyangLi:
...ts are smaller; Example: https://jsfiddle.net/yli119/FNys7/108/ when you try to increment from 0.09 to 0.1, the spinner shows 0.1 but not 0.10. Fix: The parseFloat smash the .0 or .00, you might parse float firstly and then run toFixed right before returning. Thanks. http://yiyang.li

### DIFF
--- a/ui/spinner.js
+++ b/ui/spinner.js
@@ -352,17 +352,20 @@ return $.widget( "ui.spinner", {
 		// - rounding is based on 0, so adjust back to our base
 		value = base + aboveMin;
 
-		// fix precision from bad JS floating point math
-		value = parseFloat( value.toFixed( this._precision() ) );
+		// ensure it is a float
+		value = this._parse( value );
 
 		// clamp the value
 		if ( options.max !== null && value > options.max) {
-			return options.max;
+			value = options.max;
 		}
 		if ( options.min !== null && value < options.min ) {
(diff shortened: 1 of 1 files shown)


scottgonzalez:
This method is only meant for numeric values. You probably want to change the formatting, which currently runs through Globalize, if necessary. We will be adding an option to provide your own formatter, but this is not a proper change.

YiyangLi:
ah, I get your point. our product is still using jQuery 1.10, I think we need to upgrade to have the _parse and _format functions in spinner.

scottgonzalez:
They have existed since the first version.

YiyangLi:
http://jsfiddle.net/yli119/aawLLhy7/1/

did I set it correctly? Based on the api, http://api.jqueryui.com/spinner/

I set the numberFormat as C, meaning currency. But when I change it from 0.99 to 1, it shows 1, instead of 1.00

scottgonzalez:
You would need to include Globalize (0.x) as well. For something simple, you can just override the `_format()` method on the instance.

YiyangLi:
You are right, I made it by including Globalize, and using numberFormat of N2. Thanks.

PR #1485, opened 2015-03-12
Slider: Test to check reaching max with min that less than step

atomiomi:
Ref #10078


scottgonzalez:
Thanks. In the future, you can just do an interactive rebase (`git rebase -i`) to squash your commits together and change the commit message, then force push (`git push -f`) to the existing branch to update the PR rather than opening a new PR.

PR #1490, opened 2015-03-13
Autocomplete: Simplify _create() method

jzaefferer:
Uses logical expressions in place of conditional expressions.

Thanks for @rxaviers for noticing this in the first place.


scottgonzalez:
I personally find this logic harder to follow, but I won't object to the change.

arschmitz:
I prefer the the new logic. To me it is easier to follow.

PR #1494, opened 2015-03-13
Tabs: Fix style issues

jzaefferer:
Powered by esformatter! Will do more of these with individual files.

### DIFF
--- a/ui/tabs.js
+++ b/ui/tabs.js
@@ -16,11 +16,11 @@
 //>>css.structure: ../themes/base/tabs.css
 //>>css.theme: ../themes/base/theme.css
 
-(function( factory ) {
+( function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
 
 		// AMD. Register as an anonymous module.
-		define([
+		define( [
 			"jquery",
 			"./core",
 			"./widget"
@@ -30,7 +30,7 @@
 		// Browser globals
 		factory( jQuery );
 	}
-}(function( $ ) {
+}( function( $ ) {
 
 return $.widget( "ui.tabs", {
 	version: "@VERSION",
@@ -56,7 +56,7 @@ return $.widget( "ui.tabs", {
 		load: null
 	},
 
-	_isLocal: (function() {
+	_isLocal: ( function() {
 		var rhash = /#.*$/;
 
 		return function( anchor ) {
@@ -75,7 +75,7 @@ return $.widget( "ui.tabs", {
 
 			return anchor.hash.length > 1 && anchorUrl === locationUrl;
 		};
-	})(),
+	} )(),
 
 	_create: function() {
 		var that = this,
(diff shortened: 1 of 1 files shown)


ui/tabs.js
```diff
@@ -410,27 +410,27 @@ return $.widget( "ui.tabs", {
...
+			} );
 		this._addClass( this.tabs, "ui-tab", "ui-state-default" );
 
-		this.anchors = this.tabs.map(function() {
-				return $( "a", this )[ 0 ];
-			})
-			.attr({
+		this.anchors = this.tabs.map( function() {
+			return $( "a", this )[ 0 ];
+		} )
```
scottgonzalez: I think we either need to clarify this in our style guide or change how this is formatted. The current rule is:

> When a chain of method calls is too long to fit on one line, there must be one call per line, with the first call on a separate line from the object the methods are called on. If the method changes the context, an extra level of indentation must be used.

So, presumably this should be:

```js
this.anchors = this.tabs
    .map( function() {
        return $( "a", this )[ 0 ];
    } )
        .attr( {
            ...
        } );
```

scottgonzalez: There are other instances of this below.

jzaefferer:
Updated with a second commit. Addresses several cases where line breaks were missing. If these changes are correct, I can look into update my esformatter-jquery-chain plugin to address those line breaks as well.

scottgonzalez:
I don't remember where we had the conversation, but I thought we decided that filtering as the first action was allowed on the first line.

/cc @mikesherov

jzaefferer:
That's pretty random - both `find` and `filter` change context.

Can I merge the first commit and open a new PR for the other?

scottgonzalez:
> That's pretty random - both find and filter change context.

And `.not()` and `.children()` and lots of other methods. I thought this existing block was valid:

```js
this.tabs.not( this.active ).attr( {
	"aria-selected": "false",
	"aria-expanded": "false",
	tabIndex: -1
} );
```

Or maybe:

```js
this.tabs.not( this.active )
	.attr( {
		"aria-selected": "false",
		"aria-expanded": "false",
		tabIndex: -1
	} );
```

But I didn't expect it to be:

```js
this.tabs
	.not( this.active )
		.attr( {
			"aria-selected": "false",
			"aria-expanded": "false",
			tabIndex: -1
		} );
```

Though I understand that's the rule as written.

dmethvin:
The first looks best to me as a reader and is still very clear. I'm not sure how the style guide should balance human readability vs absolute rules that can be enforced by a formatter, but it seems sad to lose the former in search of the latter.

jzaefferer:
My second commit was created by hand, no esformatter involved. Both esformatter and my esformatter-jquery-chain plugin tolerate line breaks as-is. The plugin only fixes indent.

jzaefferer:
I merged my first commit, created a new PR for the other, #1508

PR #1499, opened 2015-03-14
Position: Fix style issues

jzaefferer:
For this one I had to manually remove the additional wrapper, then use esformatter to format the file, then manually add back the extra wrappers. Not great, but hard to avoid since the indent of most of the file is technically invalid.


scottgonzalez:
Looks good.

I'd love to get rid of all the rounding logic, but I haven't gotten around to checking if it's safe.

PR #1500, opened 2015-03-14
Widget: Fix style issues

jzaefferer:
Powered by esformatter! Will do more of these with individual files.

### DIFF
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -13,7 +13,7 @@
 //>>docs: http://api.jqueryui.com/jQuery.widget/
 //>>demos: http://jqueryui.com/widget/
 
-(function( factory ) {
+( function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
 
 		// AMD. Register as an anonymous module.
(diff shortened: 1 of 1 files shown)


ui/widget.js
```diff
@@ -514,7 +514,7 @@ $.Widget.prototype = {
 				// - disabled class as method for disabling individual parts
 				if ( !suppressDisabledCheck &&
 						( instance.options.disabled === true ||
-							$( this ).hasClass( "ui-state-disabled" ) ) ) {
+						$( this ).hasClass( "ui-state-disabled" ) ) ) {
```
scottgonzalez: This is an interesting one. I guess we don't have any rules about multi-line groups, so this is correct according to the style guide. I've started splitting closing parens on to their own lines in my projects because of this.

jzaefferer: Esformatter tends to leave line breaks as-is. I can always update settings to enforce them one way or the other. I'll going with merging this PR for now, without further changes.

PR #1502, opened 2015-03-16
Slider: Setting values or range slider to max, doesn't lock slider

atomiomi:
When both values are set to the maximum change them in descending order

Fixes [#9046](http://bugs.jqueryui.com/ticket/9046)

### DIFF
--- a/ui/slider.js
+++ b/ui/slider.js
@@ -410,16 +410,9 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		}
 	},
 
-	_changeAllValues: function( valuesLength ) {
-		var i;
-
-		for ( i = valuesLength - 1; i >= 0; i -= 1 ) {
-			this._change( null, i );
-		}
-	},
-
 	_setOption: function( key, value ) {
-		var valsLength = 0;
+		var i,
+			valsLength = 0;
 
 		if ( key === "range" && this.options.range === true ) {
 			if ( value === "min" ) {
@@ -460,7 +453,12 @@ return $.widget( "ui.slider", $.ui.mouse, {
 			case "values":
 				this._animateOff = true;
 				this._refreshValue();
-				this._changeAllValues( valsLength );
+
+				// use reverse loop to change values, then last changed handle will be the first handle
+				// it need when user set both values to max ( with normal loop slider locks )
+				for ( i = valsLength - 1; i >= 0; i -= 1 ) {
+					this._change( null, i );
+				}
 				this._animateOff = false;
 				break;
 			case "step":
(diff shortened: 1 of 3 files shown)


ui/slider.js
```diff
@@ -410,9 +410,24 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		}
 	},
 
-	_setOption: function( key, value ) {
+	_changeAllValues: function( valuesLength ) {
 		var i,
-			valsLength = 0;
+			rangeValues = [ this.values( 0 ), this.values( 1 ), this._valueMax() ];
+
+		if ( this.options.range === true && $.unique( rangeValues ).length === 1 ) {
```
jzaefferer: Is this check actually necessary? Might be enough to just always do the reverse loop. I tested that locally with the new visual test page and the unit tests, works for both.

jzaefferer:
We generally avoid adding more visual test pages, whenever its possible to cover the same behaviour with unit tests instead. In this case, we already have [one test](https://github.com/jquery/jquery-ui/blob/581bfb55bf4f4f614a6e0c0ac545cf1aa7d29d0a/tests/unit/slider/slider_events.js#L101-L150) that deals with overlapping handles - can you look into extending that test?

ui/slider.js
atomiomi: Yes, I've also tested that, it works well. I was thinking that setting values with reverse loop to the minimum will lock slider, but didn't check it.

atomiomi:
Yes, after writing unit test, visual one will be unuseful I think it should be deleted

jzaefferer:
Alright. So you could update the unit test first, once that passes with the patch in place and fails without, refactor the patch as discussed and delete the visual test.

atomiomi:
If everything is ok then I will remove visual test

ui/slider.js
```diff
@@ -453,9 +460,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 			case "values":
 				this._animateOff = true;
 				this._refreshValue();
-				for ( i = 0; i < valsLength; i += 1 ) {
```
scottgonzalez: Why not just modify this one line?

jzaefferer: :+1: 

Now that the other changes are gone, the separate method isn't needed anymore.

atomiomi: Ok, I'll fix it.

eugsa:
For me it seems that there is still an issue, please have a look at [this fiddle](https://jsfiddle.net/u7Lmokt0/).

I encountered the problem with a slider which has two handle, while setting its values to the maximum. It happens when both values are set at the same time, or when I set the value one by one with the handle [0] first. And strangely, if you set the handle [1] first, then it works perfectly fine.

scottgonzalez:
@eugsa It looks like the problem is fixed for standard multi-thumb sliders, but not for range sliders. Would you mind filing an issue at https://bugs.jqueryui.com/newticket and include a link to your test case? Thanks!

PR #1503, opened 2015-03-16
Merge master into interactions and introduce PEP

bethge:
I hope this merge goes in the direction of what you had in mind. arschmitz gave me a heads about the recent changes to the master branch, but I first wanted to make sure this PR is not totally off the mark.

The merge is strongly based on the master branch using the interactions branch rather in spirit than in code, due to considerable changes in the master branch.

`connectToSortable` causes issues, because it uses overloaded mouse functions of `sortable.js`. Its tests have been deactivated for now, along with resizable tests in draggable.

As a _temporary workaround_ for overloaded mouse functions I use temporary flags on the widget. It is not very pretty to look at and only intended as an intermediate solution.

Credit for `interaction.js` goes to the interactions-polymer branch.

I noticed that draggable extends and directly calls `_mouseUp`, thus widgets have the ability to stop a mouse interaction by directly calling `_mouseUp`. If I am not mistaken, the new interaction interface does not give such possibilities atm, since the `stopHandler` is only directly attached to the event listener. I currently mimic the behaviour by triggering an actual pointerup event. This does not seem ideal though.

The polyfill is (naturally) the latest build from https://github.com/jquery/PEP .

Lastly: While the unit tests in the browser are mostly green, phantomjs rather feels like wreaking havoc. I am not sure why that is.


jzaefferer:
I wonder if we'd be better off starting a new branch and cherry- or handpicking changes from elsewhere...

bethge:
Yes, I agree. To be honest, I totally overlooked that there are also other interaction interfaces introduced in the interactions branch. As such, merging master becomes a whole lot more intricate.

PR #1507, opened 2015-03-18
Build: Increase htmllint coverage by testing all files, ignoring errors

jzaefferer:
Also formats `jscsBad` array and sorts it alphabetically.


scottgonzalez:
Looks good.

PR #1508, opened 2015-03-18
Tabs: Improve styling, adding line breaks on multiple chains

jzaefferer:
Stricter adherence to style guide, not necessarily in improvement in readability though.

### DIFF
--- a/ui/tabs.js
+++ b/ui/tabs.js
@@ -352,16 +352,19 @@ return $.widget( "ui.tabs", {
 		this._setupEvents( this.options.event );
 		this._setupHeightStyle( this.options.heightStyle );
 
-		this.tabs.not( this.active ).attr( {
-			"aria-selected": "false",
-			"aria-expanded": "false",
-			tabIndex: -1
-		} );
-		this.panels.not( this._getPanelForTab( this.active ) )
-			.hide()
-			.attr( {
-				"aria-hidden": "true"
-			} );
+		this.tabs
+			.not( this.active )
+				.attr( {
+					"aria-selected": "false",
+					"aria-expanded": "false",
+					tabIndex: -1
+				} );
+		this.panels
+			.not( this._getPanelForTab( this.active ) )
+				.hide()
+				.attr( {
+					"aria-hidden": "true"
+				} );
 
 		// Make sure one tab is in the tab order
 		if ( !this.active.length ) {
@@ -412,16 +415,18 @@ return $.widget( "ui.tabs", {
 				}
 			} );
 
-		this.tabs = this.tablist.find( "> li:has(a[href])" )
-			.attr( {
-				role: "tab",
-				tabIndex: -1
-			} );
+		this.tabs = this.tablist
+			.find( "> li:has(a[href])" )
+				.attr( {
+					role: "tab",
+					tabIndex: -1
+				} );
 		this._addClass( this.tabs, "ui-tab", "ui-state-default" );
 
-		this.anchors = this.tabs.map( function() {
-			return $( "a", this )[ 0 ];
-		} )
+		this.anchors = this.tabs
+			.map( function() {
+				return $( "a", this )[ 0 ];
+			} )
 			.attr( {
 				role: "presentation",
 				tabIndex: -1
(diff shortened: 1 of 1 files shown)


jzaefferer:
From #1494, Scott wrote:

---

>That's pretty random - both find and filter change context.

And .not() and .children() and lots of other methods. I thought this existing block was valid:
```js
this.tabs.not( this.active ).attr( {
    "aria-selected": "false",
    "aria-expanded": "false",
    tabIndex: -1
} );
```
Or maybe:
```js
this.tabs.not( this.active )
    .attr( {
        "aria-selected": "false",
        "aria-expanded": "false",
        tabIndex: -1
    } );
```
But I didn't expect it to be:
```js
this.tabs
    .not( this.active )
        .attr( {
            "aria-selected": "false",
            "aria-expanded": "false",
            tabIndex: -1
        } );
```
Though I understand that's the rule as written.

---

Dave commented: The first looks best to me as a reader and is still very clear. I'm not sure how the style guide should balance human readability vs absolute rules that can be enforced by a formatter, but it seems sad to lose the former in search of the latter.

---

I added: My second commit was created by hand, no esformatter involved. Both esformatter and my esformatter-jquery-chain plugin tolerate line breaks as-is. The plugin only fixes indent.

dmethvin:
I think it definitely makes sense to line-break, indent, and match stacking operations when you are going to `.end()` them further down the chain. Extending that rule to cases like these seems excessive to me.

ui/tabs.js
```diff
@@ -664,24 +669,27 @@ return $.widget( "ui.tabs", {
...
 		} else if ( toShow.length ) {
-			this.tabs.filter( function() {
-				return $( this ).attr( "tabIndex" ) === 0;
-			} )
-				.attr( "tabIndex", -1 );
+			this.tabs
+				.filter( function() {
+					return $( this ).attr( "tabIndex" ) === 0;
+				} )
+					.attr( "tabIndex", -1 );
```
dmethvin: I can see how the rules get us here, but ... ick.  I'd prefer something like
```js
this.tabs.filter( function() {
  return $( this ).attr( "tabIndex" ) === 0;
} ).attr( "tabIndex", -1 );
```

scottgonzalez: This is effectively the same case as what I mentioned earlier. Since the first operation is a context change, we should be able to keep it on one line, which gets us to:

```js
this.tabs.filter( function() {
    return $( this ).attr( "tabIndex" ) === 0;
} )
    .attr( "tabIndex", -1 );
```

scottgonzalez: Or does it end up being:

```js
this.tabs.filter( function() {
        return $( this ).attr( "tabIndex" ) === 0;
    } )
    .attr( "tabIndex", -1 );
```

dmethvin: I'd prefer the second. Here's the way I'd look at it: With jQuery chaining we have two types of patterns. 

1) Sequential code, which is represented here. Results may be filtered but only those filtered results are used later in the chain. Indenting doesn't do anything for us there, and is distracting.

2) Stacking/unstacking code, which occurs when there is some filtering operation and `.end()` is used further down the chain. Indenting is important there to mark the stack/unstack points.

scottgonzalez:
So what's the verdict on this?

jzaefferer:
By itself, this is good, right?
```js
this.tabs.filter( function() {
    return $( this ).attr( "tabIndex" ) === 0;
} );
```

So adding another call shouldn't change those lines except for removing the semicolon:

```js
this.tabs.filter( function() {
    return $( this ).attr( "tabIndex" ) === 0;
} )
    .attr( "tabIndex", -1 );
```

Also having to indent the function body and closing braces makes this rule worse, in that it forces even more changes on unrelated lines.

jzaefferer:
Based on the above, I suggest closing this PR, leaving things as they are. We can make the rules for chained method calls a bit more lenient, something like "Context changing method calls at the start of the chain are acceptable on the first line.", along with an example, e.g. the filter/attr chain.

scottgonzalez:
Sounds good to me.

Issue filed: https://github.com/jquery/contribute.jquery.org/issues/107

PR #1509, opened 2015-03-18
Position: Simplify fraction support test

scottgonzalez:
Also makes the test lazy to avoid any potential layouts/recalculations during
initialization.

Fixes #9898
Ref #9899

### DIFF
--- a/tests/unit/position/position_core.js
+++ b/tests/unit/position/position_core.js
@@ -16,8 +16,6 @@ module( "position", {
 	}
 });
 
-TestHelpers.testJshint( "position" );
-
 test( "my, at, of", function() {
 	expect( 4 );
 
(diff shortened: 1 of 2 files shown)


tests/unit/position/position_core.js
```diff
@@ -763,4 +761,6 @@ test( "bug #8710: flip if flipped position fits more", function() {
 	}, "no flip - top fits less" );
 });
 
+TestHelpers.testJshint( "position" );
```
scottgonzalez: This is whacky, but having this run first causes intermittent failures in IE.

jzaefferer: I still wonder if this could be fixed by `positon: absolute` on the support test element.

scottgonzalez: Indeed it does :-)

jzaefferer:
:+1:

scottgonzalez:
Landed in 3970e8c68146a3645f09d54735a40d5cf7eeef9e.

PR #1510, opened 2015-03-18
Effects: Fix style issues in individual effects

jzaefferer:
Skipping ui/effect.js since embedded jquery-color has many issues.


scottgonzalez:
Looks good.

PR #1511, opened 2015-03-18
Tests: Replace resource loader with AMD

scottgonzalez:
Replaces gh-1335.
Replaced by gh-1528.

This reduces the changes to only affect the accordion test suite so it's easier to discuss and make changes.

* [x] Simplify per-test boilerplate
* [x] Convert `domEqual()` to a real assertion
* [ ] Replace custom `QUnit.assert.close()` with qunit-assert-close module
* [ ] Move `domEqual()` to its own module
* [ ] Properly update QUnit (can now be done by rebasing on master)
* [ ] Update all tests (currently just accordion)
* [ ] Remove globals from `tests/.jshintrc`
* [x] Figure out how to use AMD in grunt-contrib-qunit
* [x] Pull in `domEqual()` changes from https://github.com/jquery/jquery-ui/pull/1516
* [x] Update grunt-contrib-qunit and use the bridge from there (https://github.com/gruntjs/grunt-contrib-qunit/pull/94)


jzaefferer:
Reproducing a comment from the original PR, still need to do that:

Can we smash all the files in test/unit/helper/ into one js file? I don't see the point in splitting those up when we use them on every page. And you can get rid of the window.require dependency by exporting the right variable: http://requirejs.org/docs/api.html#config

```js
var require = { ... };
```
Would also remove the need to export window.helper.jqueryUrl at all, since you could just use that directly.

scottgonzalez:
This is ready for a review. I've made a lot of changes from the original implementation. The new implementation significantly reduces the boilerplate for each test.

scottgonzalez:
grunt-contrib-qunit doesn't support AMD :cry:

Gruntfile.js
```diff
@@ -221,6 +221,8 @@ grunt.initConfig({
 		}
 	},
 	qunit: {
+		// TODO: should disable injecting the bridge
```
arschmitz: Blank line before comments :-p

arschmitz:
Just went through this with @scottgonzalez on skype every thing looks good.

:+1:

jzaefferer:
Updated master to QUnit 1.18.0.

scottgonzalez:
Great. I've got a few other tasks to work on right now, but I'll finish this up soon.

scottgonzalez:
I've squashed most of these commits and rebased on master. New PR is at https://github.com/jquery/jquery-ui/pull/1528.

PR #1514, opened 2015-03-20
Slider: Drag selected handle instead of closest

atomiomi:
Drag last changed handle if lastChangedValue is set and target is not a handle

Fixes  #9553

### DIFF
--- a/tests/unit/slider/slider_events.js
+++ b/tests/unit/slider/slider_events.js
@@ -7,28 +7,45 @@ module( "slider: events" );
 // value (even if same as previous value), via mouse(mouseup) or keyboard(keyup)
 // or value method/option"
 test( "mouse based interaction", function() {
-	expect( 4 );
+	expect( 5 );
 
-	var element = $( "#slider1" )
+	var handlePosition,
+		element = $( "#slider1" )
+			.slider({
+				start: function( event ) {
+					equal( event.originalEvent.type, "mousedown", "start triggered by mousedown" );
+				},
+				slide: function( event) {
+					equal( event.originalEvent.type, "mousemove", "slider triggered by mousemove" );
+				},
+				stop: function( event ) {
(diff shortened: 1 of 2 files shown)


ui/slider.js
```diff
@@ -72,6 +72,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		this._mouseSliding = false;
 		this._animateOff = true;
 		this._handleIndex = null;
+		this._lastChangedValue = null;
```
scottgonzalez: This is a bad name for a property that hold an index instead of a value. Let's rename it.

tests/unit/slider/slider_events.js
```diff
@@ -7,28 +7,45 @@ module( "slider: events" );
...
+			values: [ 20, 80 ]
 		});
+	handles = element.find( ".ui-slider-handle" );
+	handles.eq( 1 ).simulate( "drag", { dx: 10 } );
+	handlePosition = handles.eq( 0 ).offset();
 
-	element.find( ".ui-slider-handle" ).eq( 0 )
-		.simulate( "drag", { dx: 10, dy: 10 } );
-
+	element.simulate( "mousedown", { clientX: handlePosition.left + 110, clientY: handlePosition.top } );
```
scottgonzalez: This seems like an odd test. Unless I'm missing something, this would be impossible to occur in real usage, right?

atomiomi: This test checks clicking the track. I suggested that when the user clicks the track last changed handle also should be dragged as it is implemented in jQuery Mobile slider, is it right behavior? I didn't write the test that checks when handles overlap (see next commit).

atomiomi:
Current implementation of `_closestHandleIndex` doesn't find correct closest handle when user clicks the track. I will fix it when you answer what behavior should be like

dmethvin:
Closing as a stale branch.

PR #1515, opened 2015-03-23
Resizable: Add BOTTOM and RIGHT support to ui.resizable

feildmaster:
The issue: ui.resizable is bound to LEFT and TOP, disregarding BOTTOM and RIGHT CSS properties.

By using jQuery.positon, we remove that bind and add support for BOTTOM and RIGHT being set on resizable elements. 

Working demo here: https://jsfiddle.net/feildmaster/5cpy5hag/

Fixes #3011

### DIFF
--- a/ui/resizable.js
+++ b/ui/resizable.js
@@ -329,16 +329,11 @@ $.widget("ui.resizable", $.ui.mouse, {
 
 		this._renderProxy();
 
-		curleft = this._num(this.helper.css("left"));
-		curtop = this._num(this.helper.css("top"));
-
-		if (o.containment) {
-			curleft += $(o.containment).scrollLeft() || 0;
-			curtop += $(o.containment).scrollTop() || 0;
-		}
-
 		this.offset = this.helper.offset();
-		this.position = { left: curleft, top: curtop };
+		this.position = el.position();
+
+		curleft = this.position.left;
+		curtop = this.position.top;
 
 		this.size = this._helper ? {
 				width: this.helper.width(),


feildmaster:
I've signed the CLA. Didn't realize there was one.

feildmaster:
The URL is irrelevant to the commit, I will remove that from the commit directly.

scottgonzalez:
Thanks. This doesn't actually add support for bottom and right positioning though. Actual support would mean using those properties to set the position instead of top and left.

There's also some discussion in #1446 about this approach.

feildmaster:
Hmm... I agree that my "fix" is a little lacking.

However it does let people set bottom/right positions without breaking. I feel like adding true support for BOTTOM/RIGHT would have to change the entire widget..

This does however fix a bug that exists in chrome, and possibly something in firefox, where if you resize the element it shrinks or jumps depending on what it's container is. It's also worth noting that if you add "draggable" to the element, you can then move the element and fix the issue. I assume it's because draggable is doing the same thing I do here, and getting the top/left position. In which case, that interaction would have to be rewritten as well. I believe that a lot of code is dependent on "left" and "top" in the core engine, which means it would be a considerable job to fix it.

I am more than happy to simply patch it on my own script, until this issue gets sorted out, since it does sort of fix the problem I have. I don't think that's a job for someone like me with little experience with the core code to fix through a PR. However, if I get the time and inspiration to do so, I will look over what all would need to be changed for proper support. Since I would prefer it, over using top/left.

Now that I've written all this I realize my title is a bit misleading and incorrect. I will have to take a break and collect my thoughts on how I want to approach this issue.

scottgonzalez:
Closing due to inactivity.

PR #1516, opened 2015-03-23
Null attributes

scottgonzalez:
https://github.com/jquery/jquery/commit/aaeed53e9f1e299975cb6f697c8038ec3a83fa4d

### DIFF
--- a/tests/unit/testsuite.js
+++ b/tests/unit/testsuite.js
@@ -305,11 +305,11 @@ window.domEqual = function( selector, modifier, message ) {
 			result = {};
 		$.each( properties, function( index, attr ) {
 			var value = elem.prop( attr );
-			result[ attr ] = value !== undefined ? value : "";
+			result[ attr ] = value != null ? value : "";
 		});
 		$.each( attributes, function( index, attr ) {
 			var value = elem.attr( attr );
-			result[ attr ] = value !== undefined ? value : "";
+			result[ attr ] = value != null ? value : "";
 		});
 		result.style = getElementStyles( elem[ 0 ] );
 		result.events = $._data( elem[ 0 ], "events" );
--- a/ui/core.js
+++ b/ui/core.js
@@ -116,7 +116,7 @@ $.fn.extend( {
 } );
 
 // selectors
-function focusable( element, isTabIndexNotNaN ) {
+function focusable( element, hasTabindex ) {
 	var map, mapName, img,
 		nodeName = element.nodeName.toLowerCase();
 	if ( "area" === nodeName ) {
@@ -131,8 +131,8 @@ function focusable( element, isTabIndexNotNaN ) {
 	return ( /^(input|select|textarea|button|object)$/.test( nodeName ) ?
 		!element.disabled :
 		"a" === nodeName ?
-			element.href || isTabIndexNotNaN :
-			isTabIndexNotNaN ) &&
+			element.href || hasTabindex :
+			hasTabindex ) &&
 		// the element and all of its ancestors must be visible
 		visible( element );
 }
@@ -157,13 +157,13 @@ $.extend( $.expr[ ":" ], {
 		},
 
 	focusable: function( element ) {
-		return focusable( element, !isNaN( $.attr( element, "tabindex" ) ) );
+		return focusable( element, $.attr( element, "tabindex" ) != null );
 	},
 
 	tabbable: function( element ) {
 		var tabIndex = $.attr( element, "tabindex" ),
-			isTabIndexNaN = isNaN( tabIndex );
-		return ( isTabIndexNaN || tabIndex >= 0 ) && focusable( element, !isTabIndexNaN );
+			hasTabindex = tabIndex != null;
+		return ( !hasTabindex || tabIndex >= 0 ) && focusable( element, hasTabindex );
 	}
 } );
 
--- a/ui/dialog.js
+++ b/ui/dialog.js
@@ -118,7 +118,9 @@ $.widget( "ui.dialog", {
 			index: this.element.parent().children().index( this.element )
 		};
 		this.originalTitle = this.element.attr( "title" );
-		this.options.title = this.options.title || this.originalTitle;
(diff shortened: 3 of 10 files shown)


tjvantoll:
Looks good. Out of curiosity how did you find the files to change? I'm assuming failing tests?

This seems like something that has the potential to break things unexpectedly.

scottgonzalez:
Yup, failing tests. No idea what else will break in the wild. The core team decided to just do it and see what happens. See https://github.com/jquery/jquery/issues/2118.

fnagel:
Code looks good, no actual testing. 

I agree with TJ that this is a potentially very dangerous change. Most non-frontend dev are used to always use strict comparison.

scottgonzalez:
What's potentially dangerous? The core change or the UI change?

scottgonzalez:
@timmywil said to bring this up in the next core meeting. The fact that this change broke UI might cause them to back it out.

tjvantoll:
Just to clarify I meant that the Core change is potentially dangerous.

scottgonzalez:
Ok, that's what I thought. I just wanted to make sure.

fnagel:
What TJ said.

tests/unit/testsuite.js
```diff
@@ -305,11 +305,11 @@ window.domEqual = function( selector, modifier, message ) {
 			result = {};
 		$.each( properties, function( index, attr ) {
 			var value = elem.prop( attr );
-			result[ attr ] = value !== undefined ? value : "";
+			result[ attr ] = value != null ? value : "";
```
gibson042: This strict comparison was fine as-is; `.prop` lacks the DOM semantics that urged us to try changing `.attr`.

gibson042: And looking at the broader context, `domEqual` might still work if you stopped doing this empty-string normalization altogether and just assigned results from `.prop` and `.attr` directly.

ui/core.js
```diff
@@ -157,13 +157,13 @@ $.extend( $.expr[ ":" ], {
 		},
 
 	focusable: function( element ) {
-		return focusable( element, !isNaN( $.attr( element, "tabindex" ) ) );
+		return focusable( element, $.attr( element, "tabindex" ) != null );
```
gibson042: This code isn't equivalent, and it looks like even the old version has bugs for some unusual tabindex values (e.g., "0xFF", "Infinity"). Reviewing https://html.spec.whatwg.org/multipage/interaction.html#the-tabindex-attribute and reviewing browser behavior, I think you want `!isNaN( parseInt( $.attr( element, "tabindex" ), 10 ) )`. See also: https://github.com/jquery/jquery/blob/7b111310977a30d02a8260a7ed6c9c437bb1b0a5/src/attributes/prop.js#L64-L79

ui/core.js
```diff
@@ -157,13 +157,13 @@ $.extend( $.expr[ ":" ], {
...
 	},
 
 	tabbable: function( element ) {
 		var tabIndex = $.attr( element, "tabindex" ),
-			isTabIndexNaN = isNaN( tabIndex );
-		return ( isTabIndexNaN || tabIndex >= 0 ) && focusable( element, !isTabIndexNaN );
+			hasTabindex = tabIndex != null;
+		return ( !hasTabindex || tabIndex >= 0 ) && focusable( element, hasTabindex );
```
gibson042: Same concerns here.

ui/core.js
scottgonzalez: "0xFF" and "Infinity" aren't valid values. We've decided not to account for such markup.

ui/dialog.js
```diff
@@ -118,7 +118,9 @@ $.widget( "ui.dialog", {
 			index: this.element.parent().children().index( this.element )
 		};
 		this.originalTitle = this.element.attr( "title" );
-		this.options.title = this.options.title || this.originalTitle;
+		if ( this.options.title == null && this.originalTitle != null ) {
```
gibson042: That looks like a ride-along bug fix for empty-string `title`. :sunglasses:

tests/unit/testsuite.js
scottgonzalez: We do the normalization because of space delimited attributes, the most common one being `class`. The attribute will start out `undefined`, but after `elem.addClass( "foo" ).removeClass( "foo" )`, the attribute will be an empty string. We consider this equivalent.

For the property check, the aim was just consistency in the logic. I don't think we've ever consider `null` and `undefined` to be non-equivalent in our check.

ui/dialog.js
```diff
@@ -433,10 +435,11 @@ $.widget( "ui.dialog", {
 	},
 
 	_title: function( title ) {
-		if ( !this.options.title ) {
+		if ( this.options.title ) {
+			title.text( this.options.title );
+		} else {
 			title.html( "&#160;" );
 		}
-		title.text( this.options.title );
```
scottgonzalez: Even if core back out the change, this change should stay.

ui/core.js
gibson042: So the presence of a "tabindex" attribute is all that matters, independent of its value? In that case, :+1:.

scottgonzalez: For our intents, yes. We assume that if the attribute exists, it contains a valid value. If someone were to file a bug saying that they did something like `tabindex="bananas"` and it resulted in our selectors finding incorrect elements, we'd close it the same as if they filed a bug about our widgets not working properly when they used the same `id` on multiple elements :-)

scottgonzalez: We used to actually test for invalid values: https://github.com/jquery/jquery-ui/commit/f80d9eb465e428b3900bc0b741392f14ecd859f0#diff-47738b779c41035dcc321cc78634c6a5R87, but we dropped those tests four years ago.

ui/dialog.js
gibson042: I would say the same of every change here. If nothing else, specifically counting on `undefined` seems risky.

scottgonzalez: I think that's legitimate. I'll go ahead and land this.

PR #1517, opened 2015-03-23
Include native dialog in appendTo/ui-front logic

scottgonzalez:
Apply the same logic to tooltips.

Fixes http://bugs.jqueryui.com/ticket/10739

### DIFF
--- a/ui/autocomplete.js
+++ b/ui/autocomplete.js
@@ -365,7 +365,7 @@ $.widget( "ui.autocomplete", {
 		}
 
 		if ( !element || !element[ 0 ] ) {
-			element = this.element.closest( ".ui-front" );
+			element = this.element.closest( ".ui-front,dialog" );
 		}
 
 		if ( !element.length ) {
--- a/ui/selectmenu.js
+++ b/ui/selectmenu.js
@@ -586,7 +586,7 @@ return $.widget( "ui.selectmenu", {
 		}
 
 		if ( !element || !element[ 0 ] ) {
-			element = this.element.closest( ".ui-front" );
+			element = this.element.closest( ".ui-front,dialog" );
 		}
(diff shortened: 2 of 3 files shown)


tjvantoll:
There should probably be tests for this, or at least a reference to the tickets in the source so we remember why we're treating native dialog elements differently.

The code itself looks good to me.

scottgonzalez:
Well, the thing is those comments would be in multiple places and would end up living on for years. Once dialogs become a more common thing, it will just be known that dialogs have crazy stacking logic.

scottgonzalez:
As for documenting why we include `dialog`, I plan on updated http://api.jqueryui.com/theming/stacking-elements/ to have this information.

ui/autocomplete.js
```diff
@@ -365,7 +365,7 @@ $.widget( "ui.autocomplete", {
 		}
 
 		if ( !element || !element[ 0 ] ) {
-			element = this.element.closest( ".ui-front" );
+			element = this.element.closest( ".ui-front,dialog" );
```
jzaefferer: Why's there no space after the comma?

scottgonzalez: It's a habit that creeps in every once in a while. I blame John for always using no spaces in selectors since they can't be munged by minification tools. Off the top of my head, I was able to think of two places we have selectors with commas and one has spaces and one doesn't :-P

I can add the spaces in; I actually prefer it that way.

jzaefferer: I prefer the space, too.

scottgonzalez: I did a search and found all the selectors that weren't using spaces. Here's a PR: https://github.com/jquery/jquery-ui/pull/1521

PR #1518, opened 2015-03-24
Core: Handle IE11 bug with document.activeElement in iframes

scottgonzalez:
Fixes #11539

How has this not come up before?

### DIFF
--- a/ui/core.js
+++ b/ui/core.js
@@ -63,6 +63,13 @@ $.extend( $.ui, {
 			activeElement = document.body;
 		}
 
+		// Support: IE 11 only
+		// IE11 returns a seemingly empty object in some cases when accessing
+		// document.activeElement from an <iframe>
+		if ( !activeElement.nodeName ) {
+			activeElement = document.body;
+		}
+
 		return activeElement;
 	},
 


jzaefferer:
Looks fine, as in, I realize that being angry at IE doesn't help, so why not add more workarounds on top of existing workarounds.

jzaefferer:
PS: https://twitter.com/jacobrossi/status/580493801086705664

jacobrossi:
@jzaefferer @scottgonzalez do either of you know if this still repros in the EdgeHTML engine? if not, can you check at remote.modern.ie? I'll get a bug filed and prioritized if it does.

scottgonzalez:
@jacobrossi I'll be working on a reduced test case tomorrow and I'll test against remote.modern.ie at that time.

scottgonzalez:
It appears to work properly on remote.modern.ie. This is as far as I got for reducing it: http://jsbin.com/fejituname/1/

jacobrossi:
Great, then that means it works OK in Project Spartan too. I realize that doesn't help you out with IE. But it's still good to see it's fixed nonetheless.

PR #1341, opened 2014-09-04
Datepicker using Globalize 1.x

rxaviers:
:boom: WIP...

The goal is to have Datepicker working with Globalize 1.x.

Depends on
- [x] https://github.com/jquery/jquery-ui/pull/1371 `fnagel:calendar-value-pr3`;

Todo
- [ ] Rebase on datepicker branch
- [x] Destroy translation data (weekHeader, prevText, nextText , datePickerRole, currentText). Create user options instead;
- [x] Move `external/date.js` into `ui/` (see #1260 (comment));
- [ ] Fix RTL LTR direction (available at `cldr.main("layout/orientation/characterOrder")`, values are `left-to-right` or `right-to-left`);
- [ ] Get currentText: "Today" from locale.
- [x] Fix Use narrow day names (e.g., "ﺥ" EEEEE) if short (e.g., "ﺎﻠﺨﻤﻴﺳ" EEE) is too lengthy;
- [ ] Update all demos accordingly. (currently, demos/datepicker/localization.html has been updated)

PR #1519, opened 2015-03-24
Calendar: fix month jumping WIP

fnagel:
This PR addresses @jzaefferer bug reports regarding the use of multiple-month and other-months demo / option. Some other related issues (with keyboard control) should be fixed now, too. See Datepicker Wiki (ToDo section).

> * demos/calendar/multiple-months.html: Clicking a date on the second or third month changes the display to show that month as the first month. Only clicking the next/prev buttons should change that.
> * demos/calendar/other-months.html: Very similar problem as the previous one, clicking lead days changes the month

https://github.com/jquery/jquery-ui/pull/1316#issuecomment-56192690

This PR introduces a couple of changes. Mainly removing the selected property from $.date and splitting the $.date object for ongoing modifications of the current value and the $.date object for the rendering of the grid. The view date object is only updated if we really want to show another month. This split helps fixing issues when dealing with multiple month grids, too.

This is work in progress but I wanted to get some feedback. Not sure if this is a good solution or if I'm missing a simpler way to fix these issues.

### DIFF
--- a/ui/calendar.js
+++ b/ui/calendar.js
@@ -142,6 +142,9 @@ return $.widget( "ui.calendar", {
 
 	_needsRefresh: function() {
 		if ( this.date.month() !== this.viewDate.month() || this.date.year() !== this.viewDate.year() ) {
+
+			// Check if the needed day is already present in our grid due
+			// to eachDay option changes (eg. other-months demo)
 			return !this.grid.find(
 					this._sanitizeSelector( "#" + this._getDayId( this.date ) )
 				).length;
--- a/ui/calendar.js
+++ b/ui/calendar.js
@@ -61,8 +61,9 @@ return $.widget( "ui.calendar", {
 		this.labels = Globalize.translate( "datepicker" );
 		this.buttonClickContext = this.element[ 0 ];
 
-		this.date = $.date( this.options.value, this.options.dateFormat ).select();
+		this.date = $.date( this.options.value, this.options.dateFormat );
 		this.date.eachDay = this.options.eachDay;
+		this.options.value = this.date.date();
 
 		this._on( this.element, {
 			"click .ui-calendar-prev": function( event ) {
@@ -332,7 +333,7 @@ return $.widget( "ui.calendar", {
 		var content = "",
 			attributes = [
 				"role='gridcell'",
-				"aria-selected='" + ( day.current ? true : false ) + "'"
+				"aria-selected='" + ( this._isCurrent( day ) ? true : false ) + "'"
 			],
 			selectable = ( day.selectable && this._isValid( new Date( day.timestamp ) ) );
 
@@ -357,7 +358,7 @@ return $.widget( "ui.calendar", {
 		if ( day === this.date && selectable ) {
 			classes.push( "ui-state-focus" );
 		}
-		if ( day.current ) {
+		if ( this._isCurrent( day ) ) {
 			classes.push( "ui-state-active" );
 		}
 		if ( day.today ) {
@@ -381,6 +382,10 @@ return $.widget( "ui.calendar", {
 		}
 
 		return content;
+	},
+
+	_isCurrent: function( day ) {
+		return day.timestamp === this.options.value.getTime();
 	},
 
 	_createButtonPane: function() {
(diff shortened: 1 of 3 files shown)


ui/calendar.js
```diff
@@ -135,16 +132,26 @@ return $.widget( "ui.calendar", {
...
+			this._refresh();
 			this.grid.focus();
 		}
 
 		this._setActiveDescendant();
 	},
 
+	_needsRefresh: function() {
+		if ( this.date.month() !== this.viewDate.month() || this.date.year() !== this.viewDate.year() ) {
+			return !this.grid.find(
```
jzaefferer: What is this checking?

ui/calendar.js
```diff
@@ -183,14 +190,14 @@ return $.widget( "ui.calendar", {
...
 			html = "",
-			currentDate = this.date,
-			months = this.date.months( this.options.numberOfMonths - 1 ),
+			currentDate = this.viewDate,
+			months = this.viewDate.months( this.options.numberOfMonths - 1 ),
 			i = 0;
 
 		for ( ; i < months.length; i++ ) {
 
 			// TODO: Shouldn't we pass date as a parameter to build* fns instead of setting this.date?
```
jzaefferer: Is this comment still relevant? Should at least update it, since you now set `viewDate`

ui/calendar.js
```diff
@@ -61,26 +61,27 @@ return $.widget( "ui.calendar", {
...
+		this.date = $.date( this.options.value, this.options.dateFormat );
+		this.viewDate = this.date.clone();
+		this.viewDate.eachDay = this.options.eachDay;
 
 		this._on( this.element, {
 			"click .ui-calendar-prev": function( event ) {
 				event.preventDefault();
 				this.date.adjust( "M", -this.options.numberOfMonths );
-				this.refresh();
+				this._refresh();
```
jzaefferer: So changing the month still changes both `this.date` and `this.viewDate`. That seems wrong, it should only change the `viewDate`, right?

ui/calendar.js
fnagel: Not sure if this comment was intended as a "fix me" or a "improve me" type of a comment. Could be removed if we are fine with the current implementation from a code quality perspective.

ui/calendar.js
fnagel: This does a basic checking if anything has changed at all (as before, see https://github.com/fnagel/jquery-ui/commit/4c2229fc51dbb70e1dcb96afdaef92d7c41a44eb#diff-563a5392a566d1a29a68b1b5a37c4cb6L139) and if so, checks if the wanted date is already available within our grid.

jzaefferer: Okay, but why is this extra check needed? That is, under what circumstance do `this.date and `this.viewDate` match, but the day isn't rendered?

ui/calendar.js
jzaefferer: Can you try the suggested approach, in a separate PR?

ui/calendar.js
fnagel: Not sure what is meant here. The find() part is needed to make sure we do not have the day available due to eachDay changes (like in other-months demo).

jzaefferer: I see. Can you add that explanation as a comment?

ui/calendar.js
fnagel: Sure, I've added this to my todo right after this PR has been merged.

ui/calendar.js
fnagel: Done.

ui/calendar.js
fnagel: Or should I presume with PR #1341 so we can finally merge this?

ui/calendar.js
fnagel: No, we want to change the current view as well as the working date. Don't confuse this.date with the currently selected date (which is set here: https://github.com/fnagel/jquery-ui/blob/datepicker-fix-jumping/ui/calendar.js#L374-L376).

jzaefferer: So `this.date` is just a temporary variable? And go away, if we go with "Shouldn't we pass date as a parameter to build* fns instead of setting this.date?"? Anyway, should't block this PR, can deal with that later.

ui/calendar.js
jzaefferer: Let's land this, then update #1341.

ui/calendar.js
fnagel: No, it would not go away if this comment has been implemented. It's a helper variable like ````this.viewDate````.

PR #1520, opened 2015-03-24
Selectmenu: Add disabled default options to product selection demo

fnagel:
As @jzaefferer proposed, I've added a "please choose" option with selected and disabled attribute.

### DIFF
--- a/demos/selectmenu/product-selection.html
+++ b/demos/selectmenu/product-selection.html
@@ -44,10 +44,7 @@
 
 		#circle {
 			float: left;
-			background: yellow;
 			border-radius: 50%;
-			width: 150px;
-			height: 150px;
 		}
 	</style>
 </head>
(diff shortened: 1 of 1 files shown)


jzaefferer:
Looks good to me. I'll let someone else take a look as well, hopefully we're not on the wrong track.

fnagel:
Yeah, not sure if the empty initial state is what we want.

scottgonzalez:
What is the intent of this change?

scottgonzalez:
@jzaefferer We're wondering why we explicitly want to demo this behavior since it's just standard `<select>` behavior.

jzaefferer:
We didn't have a single demo or even visual test that was using this standard `<select>` behaviour. I think its worth showing that it still works for `selectmenu`.

scottgonzalez:
Can we find a better demo for it? A product selector that starts with no preview at all isn't great UX.

tjvantoll:
Agreed that this is bad UX. What if we just added a fourth select to the [main demo](http://view.jqueryui.com/master/demos/selectmenu/default.html) with a more common use case for this functionality—maybe salutation?

scottgonzalez:
That sounds good to me.

I'll close this and someone can send a new PR with a different change.

PR #1522, opened 2015-03-26
Slider: handles can't be dragged over each other

atomiomi:
added check: handle's value shouldn't exceed values of next and prev handles

Fixes #4387


scottgonzalez:
As noted in the ticket, this is not desired behavior. This would be a breaking change.

PR #1523, opened 2015-03-27
Core: Work around more IE activeElement bugs

scottgonzalez:
Amazingly 76c27556f48bea48d3787c241d35e190d46c3245 caused regressions because IE 9 - 11 can return `null` for `document.activeElement` when *not* in an iframe. The fun never stops when dealing with focus in IE...

### DIFF
--- a/ui/core.js
+++ b/ui/core.js
@@ -63,6 +63,13 @@ $.extend( $.ui, {
 			activeElement = document.body;
 		}
 
+		// Support: IE 9 - 11 only
+		// IE may return null instead of an element
+		// Interestingly, this only seems to occur when NOT in an iframe
+		if ( !activeElement ) {
+			activeElement = document.body;
+		}
+
 		// Support: IE 11 only
 		// IE11 returns a seemingly empty object in some cases when accessing
 		// document.activeElement from an <iframe>


ui/core.js
```diff
@@ -63,6 +63,13 @@ $.extend( $.ui, {
 			activeElement = document.body;
 		}
 
+		// Support: IE 9 - 11 only
+		// IE may return null instead of an element
+		// Interestingly, this only seems to occur when NOT in an iframe
+		if ( !activeElement ) {
```
dmethvin: Can this just be combined in the conditional below?

scottgonzalez: I prefer the clarity of what work arounds exist over the minuscule byte savings of combining.

dmethvin:
I was a bit concerned about uses of `safeActiveElement` returning `body` but I see that in at least a few places where it's blurred you're taking care to avoid the IE body blur bug, e.g.

https://github.com/jquery/jquery-ui/blob/f33027840cdac5152599da66635981bbe68c6bda/ui/draggable.js#L151

https://github.com/scottgonzalez/jquery-ui/blob/ie-activeelement/ui/core.js#L84-90

And I still hate focus/blur.

jzaefferer:
Looks "good" to me (as good as these can get). I prefer the separate if statement as well, since that keeps the support comments separate.

PR #1525, opened 2015-03-31
Tooltip: Remove name and id from tooltip content

jiabaowu:
Fixes #11272

### DIFF
--- a/tests/unit/tooltip/tooltip_core.js
+++ b/tests/unit/tooltip/tooltip_core.js
@@ -220,3 +220,35 @@ asyncTest( "multiple active delegated tooltips", function() {
 });
 
 }( jQuery ) );
+
+// http://bugs.jqueryui.com/ticket/11272
+test ( "remove radio button conflict with live Region and tooltip content", function () {
+	expect( 3 );
+	var content = "<div id='content'>" +
+			"<input type='radio' name='hobby' checked='checked'><label>option 1</label>" +
+			"<input type='radio' name='hobby'><label>option 2</label>" +
+		"</div>",
+		stripped_content = "<div>" +
+				"<input checked=\"checked\" type=\"radio\"><label>option 1</label>" +
+				"<input type=\"radio\"><label>option 2</label>" +
+			"</div>",
+		element = $( content );
+	$( "#tooltipped1" ).tooltip({
+		content: element,
+		open: function( event, ui ) {
+			equal( ui.tooltip.children().html().toLowerCase(), stripped_content );
+		}
+	}).tooltip( "open" ).tooltip( "destroy" );
+	$( "#tooltipped1" ).tooltip({
+		content: element[0],
+		open: function( event, ui ) {
+			equal( ui.tooltip.children().html().toLowerCase(), stripped_content );
+		}
+	}).tooltip( "open" ).tooltip( "destroy" );
+	$( "#tooltipped1" ).tooltip({
+		content: content,
+		open: function( event, ui ) {
+			equal( ui.tooltip.children().html().toLowerCase(), stripped_content );
+		}
+	}).tooltip( "open" );
+});
--- a/ui/tooltip.js
+++ b/ui/tooltip.js
@@ -280,14 +280,28 @@ $.widget( "ui.tooltip", {
 		tooltipData = this._tooltip( target );
 		tooltip = tooltipData.tooltip;
 		this._addDescribedBy( target, tooltip.attr( "id" ) );
+		// Parse content into DOM remove attributes id and name
+		if ( typeof content === "string" || content.nodeType ) {
+			if ( typeof content === "string" ) {
+				tooltip_content_div = document.createElement("div");
+				tooltip_content_div.innerHTML = content;
+				dom_content = $(tooltip_content_div);
+			} else {
+				dom_content = $("<div>").html( content );
+			}
+			dom_content.find( "[id]" ).removeAttr( "id" );
+			dom_content.find( "[name]" ).removeAttr( "name" );
+			content = dom_content.html();
+		} else if ( content.jquery ) {
+			content.removeAttr( "id" ).find( "[id]" ).removeAttr( "id" );
+			content.removeAttr( "name" ).find( "[name]" ).removeAttr( "name" );
+		}
 		tooltip.find( ".ui-tooltip-content" ).html( content );
-
 		// Support: Voiceover on OS X, JAWS on IE <= 9
 		// JAWS announces deletions even when aria-relevant="additions"
 		// Voiceover will sometimes re-read the entire log region's contents from the beginning
 		this.liveRegion.children().hide();
 		a11yContent = $( "<div>" ).html( tooltip.find( ".ui-tooltip-content" ).html() );
-		a11yContent.removeAttr( "id" ).find( "[id]" ).removeAttr( "id" );
 		a11yContent.appendTo( this.liveRegion );
 
 		function position( event ) {


tests/unit/tooltip/tooltip_core.js
```diff
@@ -220,3 +220,35 @@ asyncTest( "multiple active delegated tooltips", function() {
...
 }( jQuery ) );
+
+// http://bugs.jqueryui.com/ticket/11272
+test ( "remove radio button conflict with live Region and tooltip content", function () {
+	expect( 3 );
+	var content = "<div id='content'>" +
+			"<input type='radio' name='hobby' checked='checked'><label>option 1</label>" +
+			"<input type='radio' name='hobby'><label>option 2</label>" +
+		"</div>",
+		stripped_content = "<div>" +
```
scottgonzalez: camelCase variable names.

tests/unit/tooltip/tooltip_core.js
```diff
@@ -220,3 +220,35 @@ asyncTest( "multiple active delegated tooltips", function() {
...
+				"<input checked=\"checked\" type=\"radio\"><label>option 1</label>" +
```
scottgonzalez: Why are you using single quotes above and escaped double quotes here? Be consistent.

tests/unit/tooltip/tooltip_core.js
```diff
@@ -220,3 +220,35 @@ asyncTest( "multiple active delegated tooltips", function() {
...
+				"<input type=\"radio\"><label>option 2</label>" +
+			"</div>",
+		element = $( content );
+	$( "#tooltipped1" ).tooltip({
+		content: element,
+		open: function( event, ui ) {
+			equal( ui.tooltip.children().html().toLowerCase(), stripped_content );
```
scottgonzalez: This seems a bit brittle. Is it even consistent across all browsers?

tests/unit/tooltip/tooltip_core.js
```diff
@@ -220,3 +220,35 @@ asyncTest( "multiple active delegated tooltips", function() {
...
+	$( "#tooltipped1" ).tooltip({
+		content: element[0],
+		open: function( event, ui ) {
+			equal( ui.tooltip.children().html().toLowerCase(), stripped_content );
+		}
+	}).tooltip( "open" ).tooltip( "destroy" );
+	$( "#tooltipped1" ).tooltip({
+		content: content,
+		open: function( event, ui ) {
+			equal( ui.tooltip.children().html().toLowerCase(), stripped_content );
```
scottgonzalez: Add messages to make it more obvious what's failing if we regress.

tests/unit/tooltip/tooltip_core.js
```diff
@@ -220,3 +220,35 @@ asyncTest( "multiple active delegated tooltips", function() {
...
+				"<input checked=\"checked\" type=\"radio\"><label>option 1</label>" +
+				"<input type=\"radio\"><label>option 2</label>" +
+			"</div>",
+		element = $( content );
+	$( "#tooltipped1" ).tooltip({
+		content: element,
+		open: function( event, ui ) {
+			equal( ui.tooltip.children().html().toLowerCase(), stripped_content );
+		}
+	}).tooltip( "open" ).tooltip( "destroy" );
```
scottgonzalez: Missing line break and indentation. Please see the [style guide](http://contribute.jquery.org/style-guide/js/#chained-method-calls).

ui/tooltip.js
```diff
@@ -280,14 +280,28 @@ $.widget( "ui.tooltip", {
 		tooltipData = this._tooltip( target );
 		tooltip = tooltipData.tooltip;
 		this._addDescribedBy( target, tooltip.attr( "id" ) );
+		// Parse content into DOM remove attributes id and name
+		if ( typeof content === "string" || content.nodeType ) {
+			if ( typeof content === "string" ) {
+				tooltip_content_div = document.createElement("div");
+				tooltip_content_div.innerHTML = content;
+				dom_content = $(tooltip_content_div);
```
scottgonzalez: This is silly, use jQuery.

ui/tooltip.js
```diff
@@ -280,14 +280,28 @@ $.widget( "ui.tooltip", {
...
+			} else {
+				dom_content = $("<div>").html( content );
+			}
```
scottgonzalez: Splitting these is wasteful, they're the exact same logic in jQuery.

ui/tooltip.js
```diff
@@ -280,14 +280,28 @@ $.widget( "ui.tooltip", {
...
+			dom_content.find( "[id]" ).removeAttr( "id" );
+			dom_content.find( "[name]" ).removeAttr( "name" );
+			content = dom_content.html();
+		} else if ( content.jquery ) {
+			content.removeAttr( "id" ).find( "[id]" ).removeAttr( "id" );
+			content.removeAttr( "name" ).find( "[name]" ).removeAttr( "name" );
+		}
```
scottgonzalez: Splitting these is also wasteful.

ui/tooltip.js
```diff
@@ -280,14 +280,28 @@ $.widget( "ui.tooltip", {
...
 		tooltip.find( ".ui-tooltip-content" ).html( content );
-
```
scottgonzalez: Please don't remove existing blank lines; this violates our style guide.

ui/tooltip.js
```diff
@@ -280,14 +280,28 @@ $.widget( "ui.tooltip", {
...
+				dom_content = $("<div>").html( content );
+			}
+			dom_content.find( "[id]" ).removeAttr( "id" );
+			dom_content.find( "[name]" ).removeAttr( "name" );
+			content = dom_content.html();
+		} else if ( content.jquery ) {
+			content.removeAttr( "id" ).find( "[id]" ).removeAttr( "id" );
+			content.removeAttr( "name" ).find( "[name]" ).removeAttr( "name" );
+		}
 		tooltip.find( ".ui-tooltip-content" ).html( content );
```
scottgonzalez: You can't do all of this stripping *before* putting the content into the tooltip. Then the content won't work. You have to follow the previous logic of only stripping for the live region.

scottgonzalez:
Please updated your git config to use your full name so that it matches your CLA signature.

ui/tooltip.js
jiabaowu: Hey scott I am working on all the other fixes mentioned here an I read the jshint output but could you please further explain this comment. Based on my tests if you don't strip before putting the content into the tooltip then it will have already interfered with the other radio buttons on the page. what problems does the stripping cause with the content?

scottgonzalez: What other radios on the page? Can you please provide the use case you're considering?

jiabaowu: I based this jsfiddle off the issue https://jsfiddle.net/o2wk14tq/ .
I think this use-case as shown with his code is selecting some div from the DOM which contains radio inputs and displaying them in a tooltip. If you hover over test, notice that the original radio input is unchecked. by the way thanks for taking the time to review and reply :D

tests/unit/tooltip/tooltip_core.js
jiabaowu: You are definitely right this is probably too brittle I was thinking it might be better to assert that the content does not contain "name" and "id" with a string search? rather than trying to match the exact output of .html()

ui/tooltip.js
scottgonzalez: You're effectively cloning existing content. Can you explain why you would do that for a tooltip?

tests/unit/tooltip/tooltip_core.js
scottgonzalez: You should probably just search for the DOM structure you care about and then make assertions based on the elements' attributes.

ui/tooltip.js
jiabaowu: I would say 
Firstly essentially it is bad practice to do this and the user could definitely find a way around this problem them self. 
Secondly the reason I wanted to fix this issue other than for the user who reported it was that there are only a few things you can put in a tooltip that are going to have external effects. name and id do cause those effects, why not ensure they don't? 
Finally I agree if you think this may not be a good fix as its not immediately apparent why a users id would disappear if they were attempting to select inside the tooltip. In which case it would be fixing one niche use-case and breaking another.

jiabaowu: I finished all the revisions, so thanks so much for the help with the style guide issues, it was a little hard to understand everything =D

scottgonzalez: > name and id do cause those effects, why not ensure they don't?

Because you're ensuring that they *break* for other valid uses. Content supplied as the tooltip should be used as is. Only problems created by the live region are considered bugs here.

jiabaowu: Ok then we agree, I will make a note in the issue with an explanation that this stripping has to be done by the user if they wish to clone existing content into a tooltip

scottgonzalez:
@jiabaowu Are you still interested in this? It's been a few weeks.

jiabaowu:
@scottgonzalez sorry for the lack of contact, I meant to leave a comment on the bug report but I lost my account and the password reset wont send me an email. I think this fix is impossible, but since I refactored the code do you think it is worth resubmitting this PR to just remove "name" from the radio button in liveregion?

scottgonzalez:
Replaced by #1544.

PR #1526, opened 2015-04-01
Selectmenu: Set default max-height

jzaefferer:
Fixes [#11606](http://bugs.jqueryui.com/ticket/11606)

### DIFF
--- a/themes/base/selectmenu.css
+++ b/themes/base/selectmenu.css
@@ -20,6 +20,7 @@
 	overflow: auto;
 	overflow-x: hidden;
 	padding-bottom: 1px;
+	max-height: 20em;
 }
 .ui-selectmenu-menu .ui-menu .ui-selectmenu-optgroup {
 	font-size: 1em;


scottgonzalez:
Should go through the demos to see if there are any changes that should be made. I know at least one demo is explicitly setting a max height. Not sure if it's discussed in any of them.

tjvantoll:
This seems like a sane default to me.

arschmitz:
Noticing a few issues with this PR right off when testing. 

On osx you don't see scroll bars by default. so it is not necessarily obvious that you can scroll the popup depending on font-size. Native osx selectmenus handle this by adding an arrow at the bottom http://jsbin.com/vurupo/1/edit 

On iOS for example there are no scroll bars or indication you can scroll and it cuts off perfectly after 8 and does not look like it scrolls.

Also on iOS 8 the whole menu highlights on first touch and whatever was under your finer when you start scroll is selected when you lift your finger and the menu closes. So you cant just scroll the first time. This does work find when you open a second time and you can select whatever you want after that.

I assume this is the same on master but is not noticeable since you don't ever try to scroll by default.

PR #1530, opened 2015-04-05
Slider: animation of range and handle in sync

atomiomi:
use `stop()` instead of `'queue: false'` to begin animation immediatly

Fixes #9459

### DIFF
--- a/tests/unit/slider/slider_events.js
+++ b/tests/unit/slider/slider_events.js
@@ -29,6 +29,36 @@ test( "mouse based interaction", function() {
 		.simulate( "drag", { dx: 10, dy: 10 } );
 
 });
+
+asyncTest( "range slider animation", function() {
+	expect( 2 );
+
+	var animateDuration = 2000,
+		element = $( "#slider1" )
+		.slider({
+			range: "max",
+			animate: animateDuration,
+			min: 0,
+			max: 100,
+			value: 0
+		}),
+		handle = element.find( ".ui-slider-handle" ),
+		highlight = element.find( ".ui-slider-range" ),
+		left = element.offset().left,
+		width = element.width();
+
+	element.simulate( "mousedown", { clientX: left + width });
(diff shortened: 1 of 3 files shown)


atomiomi:
I have a question about writing unit test that check animation, can you advise

tjvantoll:
Hi @atomiomi,

Thanks for taking the time to contribute to jQuery UI. This looks great and I verified that it fixes the issue: http://jsfiddle.net/tj_vantoll/939btayd/1/.

I only have one comment: we try to only use visual tests for things that can't be unit tested, and this seems like something that you could write a test for. Could you try to replace this PR's visual test with a unit test?

Thanks.

atomiomi:
Hi! Yes, I know about unit testing but I have some doubt with writing test for animation, could you send me example?

tjvantoll:
[This test](https://github.com/jquery/jquery-ui/blob/65f4feb70b75fb7e086e8c9ddc238f97d825c2f9/tests/unit/slider/slider_events.js#L101-166) from slider's test suite is probably the closest to what you're going to need. You can probably simulate two clicks on the slider, and attach an event handler on the second click to ensure that the previous animation gets completed.

For more generic animation unit test examples, [accordion](https://github.com/jquery/jquery-ui/blob/master/tests/unit/accordion/accordion_options.js) has some pretty good tests, and the [effects suite](https://github.com/jquery/jquery-ui/blob/master/tests/unit/effects/effects_core.js) is a pretty good reference as well.

scottgonzalez:
I don't think we really need a test for this.

atomiomi:
I've just finished writing unit test that checks synchroneity of max range slider. I tried to run it without changes that were made in the first commit as was expected it failed. Inclusion of patch makes it successful.

atomiomi:
Ooh I could to not write it =) Anyway it's a good experience

PR #1531, opened 2015-04-05
Revert "Draggable: Set explicit width/height instead of right/bottom"

tjvantoll:
This reverts commit 8eca7b8f45885d20c13f1bf64cad8bee5fc1d5c5.

Fixes #10725


scottgonzalez:
Seems like the right thing to do.

scottgonzalez:
Landed in d0ea32e3ad613eaaa523d0c88c776dab2b26b25f.

PR #1533, opened 2015-04-07
Slider: Range fills all space after changing orientation

atomiomi:
Resets width/height of range

### DIFF
--- a/tests/unit/slider/slider_options.js
+++ b/tests/unit/slider/slider_options.js
@@ -106,7 +106,7 @@ test( "min", function() {
 });
 
 test( "orientation", function( assert ) {
-	expect( 11 );
+	expect( 14 );
 	element = $( "#slider1" );
 
 	options = {
@@ -143,7 +143,7 @@ test( "orientation", function( assert ) {
 
 	element.slider( "destroy" );
 
-	element = $( "#slider1" ).slider({
+	element = $( "#slider2" ).slider({
 		range: "max",
 		min: 0,
 		max: 10
@@ -151,21 +151,27 @@ test( "orientation", function( assert ) {
 
 	element.slider( "option", "value", 5 );
 	element.slider( "option", "orientation", "vertical" );
-	equal( element.width(), element.find( ".ui-slider-range " ).width(),
+	equal( element.find( ".ui-slider-range" ).width(), element.width(),
 		"range should fill all horizontal space after changing orientation to vertical" );
+	equal( element.find( ".ui-slider-range" ).height(), 250,
+		"range should fill vertical space proportionally to handle value" );
 
 	element.slider( "option", "orientation", "horizontal" );
-	equal( element.height(), element.find( ".ui-slider-range " ).height(),
+	equal( element.find( ".ui-slider-range " ).height(), element.height(),
 		"range should fill all vertical space after changing orientation to horizontal" );
+	equal( element.find( ".ui-slider-range " ).width(), 250,
+		"range should fill horizontal space proportionally to handle value" );
 
 	element.slider( "destroy" );
 
-	element = $( "#slider1" ).slider({
+	element = $( "#slider2" ).slider({
 		range: true,
 		min: 0,
 		max: 100
 	});
 	element.slider( "option", { values: [ 60, 70 ] } );
+	notEqual( element.find( ".ui-slider-range " ).position().left, 0,
+		"range should not pull over to the track's border" );
 	element.slider( "option", "orientation", "vertical" );
 	equal( element.find( ".ui-slider-range " ).position().left, 0,
 		"range should pull over to the track's border" );
--- a/tests/unit/slider/slider_options.js
+++ b/tests/unit/slider/slider_options.js
@@ -106,7 +106,7 @@ test( "min", function() {
 });
 
 test( "orientation", function( assert ) {
-	expect( 9 );
+	expect( 11 );
 	element = $( "#slider1" );
 
 	options = {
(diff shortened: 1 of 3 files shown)


atomiomi:
For example let's take max range slider. If orientation is changed from horizontal to vertical then the range will fill horizontal space proportionally to value of handle before changing. I reproduced this bug with 1.10 and 1.11 version of jQuery UI in Chrome and Firefox. 

Test link:  http://jsfiddle.net/f52g9u47/4/. To reproduce this bug follow this steps:
1. Change orientation to vertical. Everything is ok. Change it back to horizontal.
1. Drag the handle to the middle of the track.
2. Set orientation to vertical
3. Now range doesn't fill all space. If you change back orientation  to horizontal, range will also fill only half vertical space.

It works in min and simple range too.

atomiomi:
I didn't create a ticket because of following error: "Maximum number of posts per hour for this IP".  I'll create it after a while.

ui/slider.js
```diff
@@ -440,7 +440,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 				this._removeClass( "ui-slider-horizontal ui-slider-vertical" )
 					._addClass( "ui-slider-" + this.orientation );
 				this._refreshValue();
-
+				this._refreshRange( value );
```
scottgonzalez: Please restore this blank line, it's required by our style guide (all comments must be preceded by a blank line).

ui/slider.js
```diff
@@ -564,6 +564,18 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		return this.max;
 	},
 
+	_refreshRange: function ( orientation ) {
+		if ( this.options.range === false ) {
```
scottgonzalez: Let's just move this conditional to the one place this is called:

```js
if ( this.options.range ) {
    this._refreshRange( value );
}
```

tests/unit/slider/slider_options.js
```diff
@@ -143,6 +143,34 @@ test( "orientation", function( assert ) {
...
+	element = $( "#slider1" ).slider({
+		range: "max",
+		min: 0,
+		max: 10
+	});
+
+	element.slider( "option", "value", 5 );
+	element.slider( "option", "orientation", "vertical" );
+	equal( element.width(), element.find( ".ui-slider-range " ).width(),
+		"range should fill all horizontal space after changing orientation to vertical" );
```
scottgonzalez: Let's add a test for height here as well, and for width in the next block.

tests/unit/slider/slider_options.js
```diff
@@ -143,6 +143,34 @@ test( "orientation", function( assert ) {
...
+	element.slider( "destroy" );
+
+	element = $( "#slider1" ).slider({
+		range: true,
+		min: 0,
+		max: 100
+	});
+	element.slider( "option", { values: [ 60, 70 ] } );
+	element.slider( "option", "orientation", "vertical" );
+	equal( element.find( ".ui-slider-range " ).position().left, 0,
```
scottgonzalez: Test the position prior to the change to verify that it's actually changing in response to the orientation.

scottgonzalez:
Can you try filing the ticket again?

ui/slider.js
atomiomi: I thought to do so. I may be wrong but by this approach it will be necessary to write the such check if this function is invoked in another method. Correct me if  I'm missing something.

scottgonzalez: That's fine, we have tons of checks for whether or not there's a range slider. All of those conditionals will go away eventually anyway since range sliders will become their own widget.

tests/unit/slider/slider_options.js
atomiomi: Do you mean that test should check that height of range is between handles?

scottgonzalez: There are explicit values applied in the stylesheet. They're in `em`s though, so you might want to set them to `px` in a `<style>` within the HTML so that you have an exact value to work with.

atomiomi: Is it proper to use `.width()`? This test works fine
```js
element.slider( "option", "value", 5 );
equal( element.find( ".ui-slider-range" ).height(), element.height() / 2, "range should be between handles" );
```

atomiomi:
http://bugs.jqueryui.com/ is unavailable, I'll try later

scottgonzalez:
The server is back up.

tests/unit/slider/slider_options.js
scottgonzalez: I'd use an explicit value instead of comparing to the main element.

atomiomi:
For some reason system created two tickets, one of them should be removed - #12205, #12206

atomiomi:
I corrected tests. What additional test for height of vertical slider did you mean? Or I misunderstood you

scottgonzalez:
The range is controlled in two directions. Right now only one dimension is checked per orientation. The test should confirm that both dimensions are adjusted properly.

atomiomi:
Ooh I got it, but how for example width range of horizontal slider should be calculated? My proposition: also specify the slider width then set the value and calculate the range width according to the total width and slider value

scottgonzalez:
Sounds good.

atomiomi:
Tests are done. It's strange that Travis doesn't check the new commit, isn't it?

scottgonzalez:
They're working through some issues: http://www.traviscistatus.com/

scottgonzalez:
Thanks, this looks good now.

PR #1534, opened 2015-04-07
draggable: fixed helper parent retrieval when dragging between sortables

patrickmaster:
when using draggable connected to multiple sortables, I was having issues when dragging from one sortable to another that was in distance of 0px. because of dragging from one sortable to another (without dragging outside any sortables), the original draggable helper parent was lost.
I just stored the draggable parent (if available which means different from "parent") for all the time of dragging interaction with sortables so it may be restored

### DIFF
--- a/ui/draggable.js
+++ b/ui/draggable.js
@@ -703,6 +703,12 @@ $.ui.plugin.add( "draggable", "connectToSortable", {
 			item: draggable.element
 		});
 
+		// Store draggable helper's container if it's not set to 'parent'
+		// so restoring it works as expected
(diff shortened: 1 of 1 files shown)


patrickmaster:
To see the issue, please go to this [fiddle](http://jsfiddle.net/patrickmaster/gh7Lzqoq/). Try dragging "Drag me down" item, move it from one sortable to the other and eventually drag the "Drag me down" outside all of the sortables. It should be then located at pretty much random position

scottgonzalez:

PR #1535, opened 2015-04-08
Menu: Ignore bubbled mouseenter events on parent items

jzaefferer:
Test uses QUnit 2.x API - inconsistent with the other tests, but also a good reference.

Fixes [#11641](http://bugs.jqueryui.com/ticket/11641)


tests/unit/menu/menu_events.js
```diff
@@ -111,6 +111,47 @@ asyncTest( "handle focus of menu with active item", function() {
 	});
 });
 
+test( "handle mouseenter on nested menu item", function( assert ) {
+	assert.expect( 8 );
+	$.ui.menu.prototype.delay = 1;
```
scottgonzalez: This needs to be reset at the end.

jzaefferer: Fixed

scottgonzalez:
Looks good.

PR #1536, opened 2015-04-13
Accordion: correct height calculated when closed

apushak:
Fixes #11938

### DIFF
--- a/tests/unit/accordion/methods.js
+++ b/tests/unit/accordion/methods.js
@@ -10,6 +10,18 @@ var equalHeight = testHelper.equalHeight,
 
 module( "accordion: methods", setupTeardown() );
 
+test( "collapsedHeight", function() {
+	expect( 1 );
+	var element = $( "#collapsible" ).accordion(),
+		content = element.find( ".ui-accordion-content" ),
+		height = content.height();
+	element.accordion( "destroy" ).accordion({
+		active: false,
+		collapsible: true
+	});
+	equal( content.height(), height );
+});
+
 test( "destroy", function( assert ) {
 	expect( 1 );
 	assert.domEqual( "#list1", function() {
--- a/tests/unit/accordion/methods.js
+++ b/tests/unit/accordion/methods.js
@@ -10,18 +10,6 @@ var equalHeight = testHelper.equalHeight,
 
 module( "accordion: methods", setupTeardown() );
 
-test( "collapsedHeight", function() {
-	expect( 1 );
-	var element = $( "#collapsible" ).accordion(),
-		content = element.find( ".ui-accordion-content" ),
-		height = content.height();
-	element.accordion( "destroy" ).accordion({
-		active: false,
-		collapsible: true
-	});
-	equal( content.height(), height );
-});
-
 test( "destroy", function( assert ) {
 	expect( 1 );
 	assert.domEqual( "#list1", function() {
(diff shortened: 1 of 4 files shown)


apushak:
I have no idea why that unit test is failing. All the tests pass when I run them on my machine

scottgonzalez:
Can we get a test for this? `heightStyle: "fill"` needs to be updated as well.

ui/accordion.js
```diff
@@ -361,7 +361,10 @@ return $.widget( "ui.accordion", {
 			maxHeight = 0;
 			this.headers.next()
 				.each( function() {
+					var display = $( this ).css( "display" );
+					$( this ).css( "display", "block" );
```
scottgonzalez: Just use `$( this ).show()`.

scottgonzalez: Well, more explicitly, follow this logic: https://github.com/jquery/jquery-ui/blob/adcc8eff431847b4897666369310cb83275a8282/ui/dialog.js#L634-L642

apushak:
I've changed it to use is( ":visible" ) and show() / hide(), but that unit test is still failing while it passes if I run it on my machine. It seems like calling show() and hide() causes some change in the element that assert.domEqual detects, but not when I run the test locally.

scottgonzalez:
I'm not sure what's going on with that failure either. Can you go ahead with the rest of the changes? I'll try to dig into the Travis failure.

apushak:
I've added a unit test.

heightStyle "fill" uses height() - innerHeight() which just measures padding, and is not effected by the visibility of the element, so no changes should be necessary. I tested this in Chrome, Firefox, and IE.

tests/unit/accordion/methods.js
```diff
@@ -10,6 +10,18 @@ var equalHeight = testHelper.equalHeight,
 
 module( "accordion: methods", setupTeardown() );
 
+test( "collapsedHeight", function() {
```
scottgonzalez: You're testing a set of options, not a method, so this is in the wrong file and the test title doesn't really make sense.

scottgonzalez: Add `// http://bugs.jqueryui.com/ticket/11938` above the test.

scottgonzalez:
The changes and test look good. Just a minor note about the placement of the test. If you'd prefer, I can make that change myself when I land this. Otherwise, I'll just wait to hear back from you. Thanks.

scottgonzalez:
Also, we still have no idea why that test is failing in Travis, but we'll keep digging.

scottgonzalez:
Can you rebase on master and see if that helps at all with Travis? It's a clean rebase, I tested locally.

apushak:
I did a rebase on master which did not fix the Travis issue. I also moved and renamed the unit test.

scottgonzalez:
I'm still trying to figure out what's causing the test to fail. So far I've tracked it down to this difference:

```diff
< "WebkitTransformOrigin": "500px 22px",
> "WebkitTransformOrigin": "492px 22px",
< "WebkitPerspectiveOrigin": "500px 22px",
> "WebkitPerspectiveOrigin": "492px 22px",
< "width": "1000px",
> "width": "984px",
```

scottgonzalez:
It's been almost a year, so you've probably forgotten about this by now. But I have good news! We just upgraded to Phantom 2 and the tests now pass on Travis.

PR #1537, opened 2015-04-15
Add jQuery as a dependency

cusspvz:
I'm currently developing a package compiler / loader and manager for browsers and WebView apps frameworks such as phonegap/cordova. This PR will defines `jquery@2.x` as a production dependency for those who `require('jquery-ui')`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,9 @@
 	"scripts": {
 		"test": "grunt"
 	},
-	"dependencies": {},
+	"dependencies": {
+		"jquery": "2.*"
+	},
 	"devDependencies": {
 		"commitplease": "2.0.0",
 		"grunt": "0.4.2",


scottgonzalez:
But that's wrong for everyone who uses npm, which is effectively everyone contributing to the project.

cusspvz:
I would like to offer official **requirehit** support for **jquery** and **jquery-ui**.

Would you consider to have it supported via our custom configuration file as soon as we release a production-ready version?

scottgonzalez:
Nope. You can do that in your own package.

cusspvz:
We cannot, at all, set dependencies for each package. Seems that `jquery-ui` will be unsupported by our *sugar* methods. Thanks for your time.

PR #1538, opened 2015-04-15
Tabs: Rename ui-tab class to ui-tabs-tab for consistent prefixes

jzaefferer:
Ref jquery/api.jqueryui.com#256
Ref #7053

I figured that we can keep it simple for this trivial change and not create separate unit test files.

### DIFF
--- a/tests/unit/tabs/core.js
+++ b/tests/unit/tabs/core.js
@@ -9,7 +9,7 @@ var state = testHelper.state;
 module( "tabs: core" );
 
 test( "markup structure", function( assert ) {
-	expect( 17 );
+	expect( 20 );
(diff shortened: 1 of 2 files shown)


scottgonzalez:
This isn't documented, but if we're going to deprecate, we should create a ticket for the deprecation and and a ticket for the removal.

jzaefferer:
Created http://bugs.jqueryui.com/ticket/12061 and http://bugs.jqueryui.com/ticket/12063 and updated this PR to reference the first ticket.

scottgonzalez:
Looks good.

arschmitz:
:+1:

PR #1539, opened 2015-04-17
Side-Scroll: Bar takes up whole width after resizing

atomiomi:
`mouseup` event didn't trigger when mouse wasn't on handle

Fixes #9728


scottgonzalez:
This is actually a pretty bad demo for us to have. I'd prefer to just delete it; I'll discuss this with the team.

scottgonzalez:
After discussing this with the team, we've decided to delete this demo.

PR #1540, opened 2015-04-17
Slider: Pass 0 value if slider reaches it

atomiomi:
If changed value is equal to zero then the last changed value is passed to `ui` object. This is due to checking in `_uiHash` method which defines whether the new value was passed or not.

Test link: https://jsfiddle.net/Lhqryk47/
1. Drag the handle on the right side
2. Value must be between 1 and 100
3. Now drag the handle on the minimum value (zero)
4. Passed value will be 1 or greater it depends on drag speed

### DIFF
--- a/ui/slider.js
+++ b/ui/slider.js
@@ -299,13 +299,13 @@ return $.widget( "ui.slider", $.ui.mouse, {
 
 	_uiHash: function( index, value, values ) {
 		var uiHash = {
-			handle: this.handles[ index ],
-			handleIndex: index,
-			value: value || this.value()
-		};
+				handle: this.handles[ index ],
+				handleIndex: index,
+				value: typeof value !== "undefined" ? value : this.value()
+			};
 
 		if ( this._hasMultipleValues() ) {
-			uiHash.value = value || this.values( index );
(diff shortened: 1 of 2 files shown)


ui/slider.js
```diff
@@ -299,13 +299,13 @@ return $.widget( "ui.slider", $.ui.mouse, {
...
 	_uiHash: function( index, value, values ) {
 		var uiHash = {
-			handle: this.handles[ index ],
-			handleIndex: index,
-			value: value || this.value()
-		};
+				handle: this.handles[ index ],
+				handleIndex: index,
+				value: typeof value !== "undefined" ? value : this.value()
+			};
```
jzaefferer: Please undo this indent change, on this line and the three above.

atomiomi:
Indents are corrected.

jzaefferer:
Looks good. I'll let @scottgonzalez  review and merge.

scottgonzalez:
Thanks. Filed the bug as http://bugs.jqueryui.com/ticket/12196, replaced `typeof value !== "undefined"` with `value !== undefined` for consistency, and merged into master.

PR #1541, opened 2015-04-18
Accordion: Use requirejs

jzaefferer:
Now that we have requirejs in externals, updating demos to use AMD is a lot easier. This PR currently changes just one demo, but that's enough to see the changes in the code (looks good to me) and the [changes in behaviour](http://view.jqueryui.com/amd-demos/demos/accordion/default.html). The latter looks rather bad, since you can now see the unstyled content well before the enhancements are applied.

Would like to hear especially from @arschmitz on this, since you've been using AMD for Mobile a lot longer.

### DIFF
--- a/demos/accordion/default.html
+++ b/demos/accordion/default.html
@@ -4,15 +4,13 @@
 	<meta charset="utf-8">
 	<title>jQuery UI Accordion - Default functionality</title>
 	<link rel="stylesheet" href="../../themes/base/all.css">
-	<script src="../../external/jquery/jquery.js"></script>
-	<script src="../../ui/core.js"></script>
-	<script src="../../ui/widget.js"></script>
-	<script src="../../ui/accordion.js"></script>
 	<link rel="stylesheet" href="../demos.css">
+	<script src="../require-config.js"></script>
+	<script src="../../external/requirejs/require.js"></script>
 	<script>
-	$(function() {
+	require( [ "ui/accordion" ], function() {
 		$( "#accordion" ).accordion();
-	});
+	} );
 	</script>
 </head>
 <body>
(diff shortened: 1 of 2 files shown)


arschmitz:
So on mobile we have a few ways that more or less completely cover this behavior ( which i have to admit is pretty ugly )

First in demos we do one of 2 things. In the development demos there is a php concat that includes the entire built file essentially, and in production we process the demos and replace with an actual built release file.

The second way we hide this is the in mobile the entire page is hidden during rendering ( regardless of how it is loaded )

The third way we allow this to be covered ( if your doing require.js and not using the full page model ). Is through use of our "enhanced" option which allows you to supply ready to go markup so there is no flash since the DOM is not manipulated we just basically set variables and event handlers up on create in this case. 

None of these address the issue here in a way that makes sense though i don't think for UI. I guess we could add something that gets included in all demos that adds display none until require finishing loading everything but this seems a bit much.

jzaefferer:
> You can see a demo of this on @jzaefferer branch [here](http://view.jqueryui.com/amd-demos-flash/demos/accordion/default-hide.html). Basically it adds display:none to the page and in the require callback removes it so you never see any un-styled content.

That makes it behave like our current demos, which block rendering as well. I wonder if we can add something to `require-config.js` to avoid the `.show()` call inside the demo code itself. Set some requirejs callback that runs when it finished loading...?

scottgonzalez:
I suppose we can also just decide that we don't care about this. Once the demos are on jqueryui.com, they'll be using a script tag pointing to the CDN.

jzaefferer:
That sounds like we should implement the changes for `jqueryui.com` before updating more demos. Then we can test both together.

jzaefferer:
Replaced by #1557

PR #1543, opened 2015-04-20
Position: Remove IE6 workarounds in tests

scottgonzalez:


jzaefferer:
:+1: assuming tests still pass everywhere...

scottgonzalez:
I'm having some issues with BrowserStack right now, but I did get successful runs in every browser that I could actually use (which included IE8).

scottgonzalez:
As expected, tests passed in all browsers: http://swarm.jquery.org/job/385

PR #1545, opened 2015-04-22
Effect: Make .transfer() callback optional

jzaefferer:
Adds two clunky tests to at least check that no exception is thrown.

Fixes #12223

### DIFF
--- a/tests/unit/effects/core.js
+++ b/tests/unit/effects/core.js
@@ -258,6 +258,32 @@ test( "createPlaceholder: preserves layout affecting properties", function() {
 	deepEqual( before.outerHeight, placeholder.outerHeight( true ), "height preserved" );
 });
 
+module( "transfer" );
+
+asyncTest( "transfer() without callback", function() {
+	expect( 0 );
+
+	// Verify that the effect works without a callback
+	$( "#elem" ).transfer( {
+		to: ".animateClass",
+		duration: 1
+	} );
+	setTimeout( function() {
+		start();
+	}, 25 );
+} );
+
+asyncTest( "transfer() with callback", function() {
+	expect( 1 );
+	$( "#elem" ).transfer( {
+		to: ".animateClass",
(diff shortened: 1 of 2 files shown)


tests/unit/effects/core.js
```diff
@@ -258,6 +258,29 @@ test( "createPlaceholder: preserves layout affecting properties", function() {
...
+	}, 25 );
+} );
+
+asyncTest( "transfer() with callback", function() {
+	expect( 0 );
+	$( "#elem" ).transfer( {
+		to: ".animateClass",
+		duration: 1
+	}, function() {
+		start();
```
scottgonzalez: Seems like you could add an `ok( true, "callback invoked" )` to at least have an assertion here.

tests/unit/effects/core.js
```diff
@@ -258,6 +258,29 @@ test( "createPlaceholder: preserves layout affecting properties", function() {
...
+asyncTest( "transfer() without callback", function() {
+	expect( 0 );
+	$( "#elem" ).transfer( {
+		to: ".animateClass",
+		duration: 1
+	} );
+	setTimeout( function() {
+		start();
+	}, 25 );
+} );
```
scottgonzalez: Perhaps add a comment explaining what we're testing here.

jzaefferer:
Updated

tests/unit/effects/core.js
```diff
@@ -258,6 +258,32 @@ test( "createPlaceholder: preserves layout affecting properties", function() {
...
+
+asyncTest( "transfer() with callback", function() {
+	expect( 0 );
```
scottgonzalez: Is this actually passing? Should be `expect( 1 )`.

jzaefferer: Yeah, that was stupid. Fixed.

scottgonzalez:
Looks good.

PR #1546, opened 2015-04-27
Add methods for finding labels and forms for controls

arschmitz:
This adds 2 new `$.fn` methods to core, and a new prop to `$.ui.escapeId`.

`$.fn.labels` which mimics the native `control.labels` ( and uses it where possible ). This method unlike the native property also works on detached DOM nodes, and in browsers not supporting `control.labels`. 

`$.fn.form` which fixes the native `control.form` in IE8 where setting the `form` attribute breaks the `form` property by changing it to a string. 

`$.ui.escapeId` is a regex for escaping `id` and similar to be able to use them as jQuery selectors. Its needed anytime you are building a selector from the id or some other attribute of an element because there are characters valid in the HTML that need to be escaped when used as a selector. 

This also updates selectmenu to use both `$.fn.labels` and `$.ui.escapeId` and it updates tabs to use `$.ui.escapeId`. `$.fn.form` is needed both by checkboxradio in the button rewrite but also by the from reset extension @scottgonzalez is working on. Checkboxradio will also use `$.ui.escapeId`. jquery-mobile will also use all three of these.

Note: This does add 3 new exceptions to `htmllint:bad` for the labels tests. These are needed to make sure we match that native implementation. While the markup may not be valid it is supported by every browser, so it is required otherwise we could get label / input mismatches which would break visual state and underlying state consistency for inputs and labels in things like checkboxradio. Specifically this can happen with inputs wrapped in multiple labels, and inputs which are wrapped in a label which is associated with an input besides the one it contains based on its `for` attribute.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -214,10 +214,7 @@ grunt.initConfig({
 					/Start tag seen without seeing a doctype first/,
 					/Element “head” is missing a required instance of child element “title”/,
 					/Element “object” is missing one or more of the following/,
-					/The “codebase” attribute on the “object” element is obsolete/,
-					/The element “label” must not appear as a descendant of the “label” element./,
-					/Any “input” descendant of a “label” element with a “for” attribute must have an ID value that matches that “for” attribute./,
-					/Any “input” descendant of a “label” element with a “for” attribute must have an ID value that matches that “for” attribute./
+					/The “codebase” attribute on the “object” element is obsolete/
 				]
 			},
 			src: htmllintBad
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -214,7 +214,10 @@ grunt.initConfig({
 					/Start tag seen without seeing a doctype first/,
 					/Element “head” is missing a required instance of child element “title”/,
 					/Element “object” is missing one or more of the following/,
-					/The “codebase” attribute on the “object” element is obsolete/
+					/The “codebase” attribute on the “object” element is obsolete/,
+					/The element “label” must not appear as a descendant of the “label” element./,
+					/Any “input” descendant of a “label” element with a “for” attribute must have an ID value that matches that “for” attribute./,
+					/Any “input” descendant of a “label” element with a “for” attribute must have an ID value that matches that “for” attribute./
 				]
 			},
 			src: htmllintBad
--- a/tests/unit/core/core.html
+++ b/tests/unit/core/core.html
@@ -173,6 +173,7 @@
 		<input id="dom-explicit-form-attr-other" form="form-detached">
 	</form>
 </div>
+<div id="weird-['x']-id"></div>
 </div>
 </body>
 </html>
--- a/tests/unit/core/core.html
+++ b/tests/unit/core/core.html
@@ -149,31 +149,38 @@
 		</div>
 	</div>
 </div>
-<div id="form-fragment">
-	<input id="body-form">
-	<input id="body-form-attr" form="form-test">
-	<form id="form-test">
-		<input id="implicit-form">
-		<input id="explicit-form-attr-parent" form="form-test">
-	</form>
-	<form id="form-test-2">
-		<input id="implicit-form-2">
-		<input id="explicit-form-attr-other" form="form-test">
-	</form>
-</div>
-<div id="form-dom">
-	<input id="dom-body-form">
-	<input id="dom-body-form-attr" form="form-detached">
-	<form id="form-detached">
-		<input id="dom-implicit-form">
-		<input id="dom-explicit-form-attr-parent" form="form-detached">
-	</form>
-	<form id="form-detached-2">
-		<input id="dom-implicit-form-2">
-		<input id="dom-explicit-form-attr-other" form="form-detached">
-	</form>
-</div>
+
 <div id="weird-['x']-id"></div>
+</div>
+<div id="form-test">
+	<input id="body:_implicit_form">
+	<input id="body:_explicit_form" form="form-1">
+	<form id="form-1">
+		<input id="form-1:_implicit_form">
+		<input id="form-1:_explicit_form" form="form-1">
+	</form>
+	<form id="form-2">
+		<input id="form-2:_implicit_form">
+		<input id="form-2:_explicit_form_other_form" form="form-1">
+	</form>
+</div>
+<div id="form-test-detached">
+	<input id="fragment:_implicit_form">
+
+	<!-- Support: Chrome > 33
+	Chrome when in a fragment with an input which has a form attribute and is not in any form
+	the form property returns the proper form however resetting the form does not reset the 
+	input the following input is commented out to stop the test from failing
+	<input id="fragment:_explicit_form" form="form-3">
+	-->
+	<form id="form-3">
+		<input id="form-3:_implicit_form">
+		<input id="form-3:_explicit_form" form="form-3">
+	</form>
+	<form id="form-4">
+		<input id="form-4:_implicit_form">
+		<input id="form-4:_explicit_form_other_form" form="form-3">
+	</form>
 </div>
 </body>
 </html>
--- a/tests/unit/core/core.html
+++ b/tests/unit/core/core.html
@@ -108,11 +108,6 @@
 
 <div id="dimensions" style="float: left; height: 50px; width: 100px; margin: 1px 12px 11px 2px; border-style: solid; border-width: 3px 14px 13px 4px; padding: 5px 16px 15px 6px;"></div>
 
-<!-- Support: Core 1.9.x Only
-There should really be a input wrapped in multiple labels below. However a bug in Core 1.9.x
-Makes it sort the dom out of order in detached fragments in some browsers. Its a super duper edge case to fail,
-( You need an input wrapped with 2 labels, in a detached DOM fragment, with no single root element, and a top level sibling label, while using core 1.9 ), but would still like to test this once we can. For now we
-Just wont worry about a test since its such an edge case. -->
 <div id="labels-fragment">
 	<label for="test">1</label>
 	<div>
--- a/tests/unit/core/core.html
+++ b/tests/unit/core/core.html
@@ -108,14 +108,21 @@
 
 <div id="dimensions" style="float: left; height: 50px; width: 100px; margin: 1px 12px 11px 2px; border-style: solid; border-width: 3px 14px 13px 4px; padding: 5px 16px 15px 6px;"></div>
 
+<!-- Support: Core 1.9.x Only
+There should really be a input wrapped in multiple labels below. However a bug in Core 1.9.x
+Makes it sort the dom out of order in detached fragments in some browsers. Its a super duper edge case to fail,
+( You need an input wrapped with 2 labels, in a detached DOM fragment, with no single root element, and a top level sibling label, while using core 1.9 ), but would still like to test this once we can. For now we
+Just wont worry about a test since its such an edge case. -->
 <div id="labels-fragment">
 	<label for="test">1</label>
 	<div>
 		<div>
 			<form>
 				<label for="test">2</label>
-				<label>3
-					<input id="test">
+				<label for="test-2">
+					<label>3
+						<input id="test">
+					</label>
 				</label>
 				<label for="test">4</label>
 			</form>
@@ -142,7 +149,11 @@
 		<div>
 			<div>
 				<form>
-					<label for="test">10</label>
+					<label for="test">
+						10
+						<input id="test-2">
+					</label>
+					<label for="test">11</label>
 				</form>
 			</div>
 		</div>
--- a/tests/unit/core/core.html
+++ b/tests/unit/core/core.html
@@ -119,10 +119,8 @@ Just wont worry about a test since its such an edge case. -->
 		<div>
 			<form>
 				<label for="test">2</label>
-				<label for="test-2">
-					<label>3
-						<input id="test">
-					</label>
+				<label>3
+					<input id="test">
 				</label>
 				<label for="test">4</label>
 			</form>
@@ -149,11 +147,8 @@ Just wont worry about a test since its such an edge case. -->
 		<div>
 			<div>
 				<form>
-					<label for="test">
-						10
-						<input id="test-2">
-					</label>
-					<label for="test">11</label>
+					<input id="test-2">
+					<label for="test">10</label>
 				</form>
 			</div>
 		</div>
--- a/tests/unit/core/core.html
+++ b/tests/unit/core/core.html
@@ -108,6 +108,70 @@
 
 <div id="dimensions" style="float: left; height: 50px; width: 100px; margin: 1px 12px 11px 2px; border-style: solid; border-width: 3px 14px 13px 4px; padding: 5px 16px 15px 6px;"></div>
 
+<div id="labels-fragment">
+	<label for="test">1</label>
+	<div>
+		<div>
+			<form>
+				<label for="test">2</label>
+				<label>3
+					<input id="test">
+				</label>
+				<label for="test">4</label>
+			</form>
+			<label for="test">5</label>
+		</div>
+		<div>
+			<div>
+				<form>
+					<label for="test">6</label>
+				</form>
+			</div>
+		</div>
+	</div>
+	<div>
+		<div>
+			<form>
+				<label for="test">7</label>
+				<label>
+				</label>
+				<label for="test">8</label>
+			</form>
+			<label for="test">9</label>
+		</div>
+		<div>
+			<div>
+				<form>
+					<label for="test">10</label>
+				</form>
+			</div>
+		</div>
+	</div>
+</div>
+<div id="form-fragment">
+	<input id="body-form">
+	<input id="body-form-attr" form="form-test">
+	<form id="form-test">
+		<input id="implicit-form">
+		<input id="explicit-form-attr-parent" form="form-test">
+	</form>
+	<form id="form-test-2">
+		<input id="implicit-form-2">
+		<input id="explicit-form-attr-other" form="form-test">
+	</form>
+</div>
+<div id="form-dom">
+	<input id="dom-body-form">
+	<input id="dom-body-form-attr" form="form-detached">
+	<form id="form-detached">
+		<input id="dom-implicit-form">
+		<input id="dom-explicit-form-attr-parent" form="form-detached">
+	</form>
+	<form id="form-detached-2">
+		<input id="dom-implicit-form-2">
+		<input id="dom-explicit-form-attr-other" form="form-detached">
+	</form>
+</div>
 </div>
 </body>
 </html>
--- a/tests/unit/core/core.html
+++ b/tests/unit/core/core.html
@@ -152,6 +152,10 @@
 
 <div id="weird-['x']-id"></div>
 </div>
+
+<!-- This is intentionally outside the test fixture. We don't want this
+markup to be removed and reinserted between tests, as it will remove saved
+refrences in the tests. -->
 <div id="form-test">
 	<input id="body:_implicit_form">
 	<input id="body:_explicit_form" form="form-1">
@@ -168,9 +172,9 @@
 	<input id="fragment:_implicit_form">
 
 	<!-- Support: Chrome > 33
-	Chrome when in a fragment with an input which has a form attribute and is not in any form
-	the form property returns the proper form however resetting the form does not reset the 
-	input the following input is commented out to stop the test from failing
+	when an input with a form attribute is inside a fragment, and not contained by any form,
+	the form property returns the proper form. However resetting the form does not reset the
+	input. The following input is commented out to stop the test from failing in this case.
 	<input id="fragment:_explicit_form" form="form-3">
 	-->
 	<form id="form-3">
--- a/tests/unit/core/core.js
+++ b/tests/unit/core/core.js
@@ -138,9 +138,6 @@ test( "uniqueId / removeUniqueId", function() {
 	equal( el.attr( "id" ), null, "unique id has been removed from element" );
 });
 
-// Support: Core 1.9 Only, IE8 Only
-// The use of $.trim() in the two tests below this is to account for IE8 missing
-// string.trim() We need to trim the values at all because of a bug in core 1.9.x
 test( "labels", function() {
 	expect( 2 );
 
@@ -150,6 +147,9 @@ test( "labels", function() {
 	function testLabels( testType ) {
 		var labels = dom.find( "#test" ).labels(),
 			found = labels.map( function() {
+
+				// Support: Core 1.9 Only
+				// We use $.trim() because core 1.9.x silently fails when white space is present
 				return $.trim( $( this ).text() );
 			} ).get();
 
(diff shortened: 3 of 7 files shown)


scottgonzalez:
> While the markup may not be valid it is supported by every browser

It doesn't matter what happens natively if it's invalid markup. We never support invalid markup.

Gruntfile.js
```diff
@@ -214,7 +214,10 @@ grunt.initConfig({
 					/Start tag seen without seeing a doctype first/,
 					/Element “head” is missing a required instance of child element “title”/,
 					/Element “object” is missing one or more of the following/,
-					/The “codebase” attribute on the “object” element is obsolete/
+					/The “codebase” attribute on the “object” element is obsolete/,
+					/The element “label” must not appear as a descendant of the “label” element./,
+					/Any “input” descendant of a “label” element with a “for” attribute must have an ID value that matches that “for” attribute./,
+					/Any “input” descendant of a “label” element with a “for” attribute must have an ID value that matches that “for” attribute./
```
scottgonzalez: Why is this duplicated?

arschmitz: mistake

scottgonzalez:
Since removing support for nested labels is likely to change a lot, I'll wait before reviewing this.

arschmitz:
It only actually removes the filter here https://github.com/jquery/jquery-ui/pull/1546/files#diff-9642379d08d860be4c2831b0ff5b9a85R168 which is why i think making sure we have the proper label to ensure the DOM matches the state of the widget is more important then saying we don't support invalid markup.

scottgonzalez:
It doesn't matter how small the code is for the support. Invalid markup is never supported and we absolutely should not test against it.

tests/unit/core/core.html
```diff
@@ -108,6 +108,76 @@
 
 <div id="dimensions" style="float: left; height: 50px; width: 100px; margin: 1px 12px 11px 2px; border-style: solid; border-width: 3px 14px 13px 4px; padding: 5px 16px 15px 6px;"></div>
 
+<!-- Support: Core 1.9.x Only
+There should really be a input wrapped in multiple labels below. However a bug in Core 1.9.x
```
scottgonzalez: This is no longer relevant.

ui/core.js
```diff
@@ -88,7 +88,9 @@ $.extend( $.ui, {
 		if ( element && element.nodeName.toLowerCase() !== "body" ) {
 			$( element ).blur();
 		}
-	}
+	},
+
+	escapeId: new RegExp( /([!"#$%&'()*+,./:;<=>?@[\]^`{|}~])/g )
```
scottgonzalez: What's the deal with `new RegExp()` and passing a literal?

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
 				$( this ).removeAttr( "id" );
 			}
 		} );
+	},
+
+	form: function() {
+		var form;
+
+		if ( this[ 0 ].form && typeof this[ 0 ].form !== "string" ) {
```
scottgonzalez: The string check needs a support comment.

ui/core.js
arschmitz: Me being dumb. no idea what i was thinking here

ui/core.js
scottgonzalez: Actually, the function should get a support comment.

arschmitz: Yeah i should change the one below and move it above the method probably

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
...
+			return this.pushStack( $( this[ 0 ].form ) );
```
scottgonzalez: I know I said we should support `.pushStack()`, but I was only thinking of the public API. This method doesn't need it.

arschmitz: WFM!

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
...
+			return this.pushStack( [] );
+		}
+
+		// Support: IE8 only ( the rest of the method )
+		// IE8 supports the form property, but not the form attribute, worse yet if you supply the
+		// form attribute it overwrites the form property with the string. Other supported browsers
+		// like all other IE's and Android 2.3 support the form attribute not at all or partially.
+		// We don't care in those cases because they still always return an element. We are not
+		// trying to fix the form attribute here, only deal with the prop supplying a string.
+		form = this.closest( "form" );
```
scottgonzalez: Why not just return `this.closest( "form" )`?

arschmitz: There used to be more logic here ill update.

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
 				$( this ).removeAttr( "id" );
 			}
 		} );
+	},
+
+	form: function() {
```
scottgonzalez: I think this can be simplified to:

```js
var form = this.prop( "form" );
return typeof form === "string" ? this.closest( "form" ) : $( form );
```

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
...
+
+				// We don't search against the document in case the element
+				// is disconnected from the DOM
+				ancestor = this.parents().last();
+
+				// get a full set of top level ancestors
+				ancestors = ancestor.add( ancestor.length ? ancestor.siblings() : this.siblings() );
+
+				// Create a selector for the label based on the id
+				selector = "label[for='" + this.attr( "id" ).replace( $.ui.escapeId, "\\$1" ) + "']";
```
arschmitz: This should be using `id` above thats the reason its stored lol

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
...
+			return this.pushStack( form );
+		}
+		return this.pushStack( [] );
+	},
+
+	labels: function() {
+		var ancestor, selector, id, labels, ancestors;
+
+			// Check control.labels first
+			if ( this[ 0 ].labels !== undefined && this[ 0 ].labels.length > 0 ) {
```
scottgonzalez: whacky indentation

scottgonzalez: `this[ 0 ].labels && this[ 0 ].labels.length`

ui/core.js
arschmitz: Is there a reason to not just do 
```js
return typeof this[ 0 ].form === "string" ? this.closest( "form" ) : $( this[ 0 ].form );
```

It avoids the small function call over head removes a line and is still under 100 chars?

ui/core.js
scottgonzalez: Why is this just a regex instead of a function that actually escapes an id?

arschmitz: probably a better solution since you still need to know the other half of the escape

ui/core.js
scottgonzalez: You're replacing a function call with array indexing and property accesses. Either way, the performance isn't something to care about.

The reason I prefer the two line version is because it's clearer what's happening, and I prefer avoiding repeated `this[ 0 ]` where possible. Open to what others want to do.

scottgonzalez: What I really care about here is the `form` variable. I don't care if we use `.prop()` or direct property access.

arschmitz: I don't really care either way ill just change it.

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
...
+
+				// get a full set of top level ancestors
+				ancestors = ancestor.add( ancestor.length ? ancestor.siblings() : this.siblings() );
+
+				// Create a selector for the label based on the id
+				selector = "label[for='" + this.attr( "id" ).replace( $.ui.escapeId, "\\$1" ) + "']";
+
+				// Support: Core 1.7
+				// Once we drop support for 1.7 we can just use .addBack( selector ) here
+				// Check both the ancestors and the contents of the ancestors for matching labels
```
scottgonzalez: This is why you should do your testing inside jQuery UI ;-)

`.addBack()` will always exist in jQuery UI. Just scroll down in this file a bit.

arschmitz: DOH!

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
...
+				labels = labels.add( ancestors.filter( selector ) );
+				labels = labels.add( ancestors.find( selector ) );
+
+			}
+
+			// Return whatever we have found for labels
+			return this.pushStack( $.unique( labels ) );
```
scottgonzalez: Remind me why `$.unique()` is used here.

ui/core.js
```diff
@@ -126,6 +128,67 @@ $.fn.extend( {
...
+
+			// Check control.labels first
+			if ( this[ 0 ].labels !== undefined && this[ 0 ].labels.length > 0 ) {
+				return this.pushStack( this[ 0 ].labels );
+			}
+
+			// Support: IE <= 11, FF <= 37, Android <=2.3 only
+			// Above browsers do not support control.labels everything below is to support them
+			// as well as document fragments control.labels does not work on document fragments anywhere
+			labels = this.parents( "label" );
```
scottgonzalez: Potentially invert the below conditional.

```js
var id = this.attr( "id" );
if ( !id ) {
    return this.eq( 0 ).parents( "label" );
}

// Handle controls with id here...
```

tests/unit/core/core.js
```diff
@@ -138,4 +138,89 @@ test( "uniqueId / removeUniqueId", function() {
 	equal( el.attr( "id" ), null, "unique id has been removed from element" );
 });
 
+// Support: Core 1.9 Only, IE8 Only
+// The use of $.trim() in the two tests below this is to account for IE8 missing string.trim()
```
scottgonzalez: Move the support comment to where the support actually exists, not all the way at the top of the function.

tests/unit/core/core.js
```diff
@@ -138,4 +138,89 @@ test( "uniqueId / removeUniqueId", function() {
...
+	expect( 2 );
+
+	var expected = [ "1", "2", "3", "4", "5", "6", "7", "8", "9", "10" ],
+		foundFragment = [],
+		foundDom = [],
+		dom = $( "#labels-fragment" ),
+		domLabels = dom.find( "#test" ).labels(),
+		fragmentLabels = $( $.trim( dom.html() ) ).find( "#test" );
+
+	domLabels.each( function(){
```
scottgonzalez: Missing space.

tests/unit/core/core.js
```diff
@@ -138,4 +138,89 @@ test( "uniqueId / removeUniqueId", function() {
...
+		foundDom.push( $.trim( $( this ).text() ) );
+	} );
+	deepEqual( foundDom, expected,
+		"Labels finds labels all labels in the DOM, and sorts them in DOM order" );
+
+	// Remove fragment DOM so id's are not duplicated and we know we are finding the detached labels
+	dom.remove();
+	fragmentLabels = fragmentLabels.labels();
```
scottgonzalez: Why not just use the removed elements?

tests/unit/core/core.js
scottgonzalez: Could use `.map( fn ).get()` instead of manually building an array.

scottgonzalez:
I still have more to review, but I'll wait for some updates.

ui/core.js
arschmitz: To sort properly in jQuery 1.7 / 1.8

ui/core.js
arschmitz: I debated this. The reason i didn't was the early return here does not gain us anything, there is still only one conditional no logic is skipped,  it only really serves to add an extra return. I also thought the intended logic was clearer this way since the idea is to return all the labels never just a subset. 

Im fine with changing if you prefer the early return since this is what we generally prefer but that was my reasoning being the current  code organization.

tests/unit/core/core.js
arschmitz: It's used 5 times in 2 tests all on different lines so i was not sure the best place to put it. I did want to duplicate 5 times and putting it in the first test at first use seemed wrong since it would have no apparent connection to second test. Thoughts?

arschmitz:
@scottgonzalez updated

tests/unit/core/core.js
scottgonzalez: The support comment should be directly above every place that support is needed.

Uses of `$.trim()` don't need support comments.

ui/core.js
scottgonzalez: In what situations are the elements not sorted properly? This needs a support comment.

ui/core.js
```diff
@@ -88,6 +90,10 @@ $.extend( $.ui, {
 		if ( element && element.nodeName.toLowerCase() !== "body" ) {
 			$( element ).blur();
 		}
+	},
+
+	escapeId: function( id ) {
```
scottgonzalez: Use a closure so you can get `idEscape` to be local and then return the actual function. See `.uniqueId()` for an example of this.

ui/core.js
```diff
@@ -126,6 +132,49 @@ $.fn.extend( {
 				$( this ).removeAttr( "id" );
 			}
 		} );
+	},
+
+	// Support: IE8 Only
+	// IE8 does not support the form attribute and when it is supplied it over wirtes the form prop
+	// with a string so we need to try and work around this by looking for the closest form if the
+	// prop is a string.
```
scottgonzalez: Missing some punctuation. Can also be reduced a bit:

```js
// IE8 does not support the form attribute and when it is supplied. It overwrites the form prop
// with a string, so we need to find the proper form.
```

ui/core.js
```diff
@@ -126,6 +132,49 @@ $.fn.extend( {
...
+		var ancestor, selector, id, labels, ancestors;
+
+		// Check control.labels first
+		if ( this[ 0 ].labels && this[ 0 ].labels.length ) {
+			return this.pushStack( this[ 0 ].labels );
+		}
+
+		// Support: IE <= 11, FF <= 37, Android <=2.3 only
+		// Above browsers do not support control.labels everything below is to support them
+		// as well as document fragments control.labels does not work on document fragments anywhere
```
scottgonzalez: Missing punctuation:

```js
// Above browsers do not support control.labels. Everything below is to support them
// as well as document fragments. control.labels does not work on document fragments anywhere.
```

ui/core.js
```diff
@@ -126,6 +132,49 @@ $.fn.extend( {
...
+
+	labels: function() {
+		var ancestor, selector, id, labels, ancestors;
+
+		// Check control.labels first
+		if ( this[ 0 ].labels && this[ 0 ].labels.length ) {
+			return this.pushStack( this[ 0 ].labels );
+		}
+
+		// Support: IE <= 11, FF <= 37, Android <=2.3 only
```
scottgonzalez: Missing space for Android version.

ui/core.js
```diff
@@ -126,6 +132,49 @@ $.fn.extend( {
...
+
+		// Look for the label based on the id
+		id = this.attr( "id" );
+		if ( id ) {
+
+			// We don't search against the document in case the element
+			// is disconnected from the DOM
+			ancestor = this.parents().last();
+
+			// get a full set of top level ancestors
```
scottgonzalez: capitalize

ui/core.js
```diff
@@ -126,6 +132,49 @@ $.fn.extend( {
...
+
+			// Create a selector for the label based on the id
+			selector = "label[for='" + $.ui.escapeId( id ) + "']";
+
+			labels = labels.add( ancestors.find( selector ).addBack( selector ) );
+
+		}
+
+		// Return whatever we have found for labels
+		return this.pushStack( $.unique( labels ) );
```
scottgonzalez: Needs a support comment with explanation for the `$.unique()` usage.

ui/tabs.js
```diff
@@ -713,7 +713,7 @@ $.widget( "ui.tabs", {
 	_getIndex: function( index ) {
 		// meta-function to give users option to provide a href string instead of a numerical index.
 		if ( typeof index === "string" ) {
-			index = this.anchors.index( this.anchors.filter( "[href$='" + index + "']" ) );
+			index = this.anchors.index( this.anchors.filter( "[href$='" + $.ui.escapeId( index ) + "']" ) );
```
scottgonzalez: Is the escaping the same? If so, we should definitely rename the method.

arschmitz: The escaping is the same and i meant to leave a comment about needing a better name.

arschmitz: Though if i'm not mistaken index here is actually an id is it not?

scottgonzalez: It's a URL.

arschmitz: Ah right sorry we use urls as a form of id in a similar case in mobile. Was thinking of that. Either way though yes this does work here to sanitize this for a selector. We use this in mobile as part of our navigation as we need to frequently sanitize things for use as a selector.

arschmitz:
All updated

tests/unit/core/core.html
```diff
@@ -108,6 +108,79 @@
...
+					<input id="test-2">
+					<label for="test">10</label>
+				</form>
+			</div>
+		</div>
+	</div>
+</div>
+
+<div id="weird-['x']-id"></div>
+</div>
```
scottgonzalez: Is this a typo?

tests/unit/core/core.html
```diff
@@ -108,6 +108,79 @@
...
+	<form id="form-1">
+		<input id="form-1:_implicit_form">
+		<input id="form-1:_explicit_form" form="form-1">
+	</form>
+	<form id="form-2">
+		<input id="form-2:_implicit_form">
+		<input id="form-2:_explicit_form_other_form" form="form-1">
+	</form>
+</div>
+<div id="form-test-detached">
```
scottgonzalez: This is unclosed.

tests/unit/core/core.html
```diff
@@ -108,6 +108,79 @@
...
+	<input id="fragment:_implicit_form">
+
+	<!-- Support: Chrome > 33
+	Chrome when in a fragment with an input which has a form attribute and is not in any form
+	the form property returns the proper form however resetting the form does not reset the 
+	input the following input is commented out to stop the test from failing
+	<input id="fragment:_explicit_form" form="form-3">
```
scottgonzalez: Crazy grammar.

tests/unit/core/core.html
arschmitz: if you mean that these are not in the test fixture, No. Though i did mean to add a comment to explain why.
If you place these in the test fixture they get reset / replaced between each async test which is not what i want. because then any saved references are to a detached fragment, not attached DOM which i specifically want to test.

tests/unit/core/core.html
arschmitz: It's closed on line 184

scottgonzalez: This is related to the above typo comment. Definitely need a comment if there's code that intentionally outside of the fixture.

arschmitz: Gotcha and yes i agree need a comment for sure

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+
+	testLabels( "the DOM" );
+
+	// Detach the dom to test on a fragment
+	dom.detach();
+	testLabels( "document fragments" );
+
+} );
+
+var domAttached = $( "#form-test" );
```
scottgonzalez: Wrap this whole block in an IIFE to clarify the scope.

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+		var labels = dom.find( "#test" ).labels(),
+			found = labels.map( function() {
+
+				// Support: Core 1.9 Only
+				// We use $.trim() because core 1.9.x silently fails when white space is present
+				return $.trim( $( this ).text() );
+			} ).get();
+
+		deepEqual( found, expected,
+			"Labels finds all labels in " + testType + ", and sorts them in DOM order" );
```
scottgonzalez: Drop "Labels" or change to ".labels()".

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+	}
+
+	testLabels( "the DOM" );
+
+	// Detach the dom to test on a fragment
+	dom.detach();
+	testLabels( "document fragments" );
+
```
scottgonzalez: Remove this blank line.

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+			input.val( "changed" );
+
+			reset();
+
+			// If there is a form we reset just that. If there is not a form, reset every form.
+			// The idea is if a form is found resetting that form should reset the input.
+			// If no form is found no amount of resetting should change the value.
+			( form.length ? form : dom.find( "form" ).addBack( "form" ) ).trigger( "reset" );
+
```
scottgonzalez: Remove this blank line.

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+		} );
+	} );
+}
+
+testForm( "form: attached: ", domAttached );
+testForm( "form: detached: ", domDetached );
+
+
```
scottgonzalez: Remove this blank line.

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
 
+test( "labels", function() {
+	expect( 2 );
+
+	var expected = [ "1", "2", "3", "4", "5", "6", "7", "8", "9", "10" ],
+		dom = $( "#labels-fragment" );
+
+	function testLabels( testType ) {
+		var labels = dom.find( "#test" ).labels(),
+			found = labels.map( function() {
```
scottgonzalez: Let's not use onevar here (or above).

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+
+	inputs.each( function() {
+		var input = $( this );
+
+		asyncTest( name + this.id.replace( /_/g, " " ), function() {
+			expect( 1 );
+			var form = input.form(),
+
+				// If input has a form the value should reset to "" if not it should be "changed"
+				value = form.length ? "" : "changed";
```
scottgonzalez: Drop onevar.

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+
+			function reset() {
+				setTimeout( function() {
+					equal( input.val(), value, "Proper form found for #" + input.attr( "id" ) );
+					start();
+				});
```
scottgonzalez: missing space

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+			}
+
+			input.val( "changed" );
+
+			reset();
```
scottgonzalez: Why is this a function instead of just being inline? It's only used once.

scottgonzalez: Move this after the reset call so it's more obvious what's happening.

tests/unit/core/core.js
```diff
@@ -138,4 +138,71 @@ test( "uniqueId / removeUniqueId", function() {
...
+
+			// If there is a form we reset just that. If there is not a form, reset every form.
+			// The idea is if a form is found resetting that form should reset the input.
+			// If no form is found no amount of resetting should change the value.
+			( form.length ? form : dom.find( "form" ).addBack( "form" ) ).trigger( "reset" );
```
scottgonzalez: Actually reset, don't trigger the event.

ui/core.js
```diff
@@ -88,7 +88,15 @@ $.extend( $.ui, {
 		if ( element && element.nodeName.toLowerCase() !== "body" ) {
 			$( element ).blur();
 		}
-	}
+	},
+
+	// Internal use only
+	escapeSelector: ( function() {
+		var selectorEscape = /([!"#$%&'()*+,./:;<=>?@[\]^`{|}~])/g;
+		return function( id ) {
```
scottgonzalez: Change the parameter name.

ui/core.js
```diff
@@ -126,6 +134,48 @@ $.fn.extend( {
...
+
+		// Check control.labels first
+		if ( this[ 0 ].labels && this[ 0 ].labels.length ) {
+			return this.pushStack( this[ 0 ].labels );
+		}
+
+		// Support: IE <= 11, FF <= 37, Android <= 2.3 only
+		// Above browsers do not support control.labels. Everything below is to support them
+		// as well as document fragments. control.labels does not work on document fragments
+		labels = this.parents( "label" );
```
scottgonzalez: We should probably be explicit and limit to `this.eq( 0 )` (below as well).

arschmitz: i kept it this way simply to match the native implementation ( it will find these and return them even though they are not valid )

arschmitz: I agree i will update this

scottgonzalez: I'm not sure what you mean. The native implementation can't possibly be invoked on a collection.

arschmitz: Yup misunderstood your comment i updated to simply say i agree

arschmitz:
@scottgonzalez updated

tests/unit/core/core.html
```diff
@@ -108,6 +108,83 @@
...
+	<form id="form-2">
+		<input id="form-2:_implicit_form">
+		<input id="form-2:_explicit_form_other_form" form="form-1">
+	</form>
+</div>
+<div id="form-test-detached">
+	<input id="fragment:_implicit_form">
+
+	<!-- Support: Chrome > 33
+	when an input with a form attribute is inside a fragment, and not contained by any form,
```
scottgonzalez: Much better :-) Just need to start with a capital letter here.

tests/unit/core/core.js
```diff
@@ -138,4 +138,67 @@ test( "uniqueId / removeUniqueId", function() {
...
+	var dom = $( "#labels-fragment" );
+
+	function testLabels( testType ) {
+		var labels = dom.find( "#test" ).labels();
+		var found = labels.map( function() {
+
+				// Support: Core 1.9 Only
+				// We use $.trim() because core 1.9.x silently fails when white space is present
+				return $.trim( $( this ).text() );
+			} ).get();
```
scottgonzalez: Outdent; that's one of the beauties of not using onevar :-)

tests/unit/core/selector.js
```diff
@@ -254,4 +254,11 @@ test( "tabbable - dimensionless parent with overflow", function() {
 	isTabbable( "#dimensionlessParent", "input" );
 });
 
+test( "escapeSelector", function() {
+	expect( 1 );
+
+	equal( $( "#" + $.ui.escapeSelector( "weird-['x']-id" ) ).length, 1,
+		"Escape id properly escapes selectors to use as an id" );
```
scottgonzalez: Drop "Escape id" from the beginning.

scottgonzalez:
Just a few minor things. You can land this after making those updates.

PR #1549, opened 2015-05-05
Update JSCS

scottgonzalez:
Upgrade to latest grunt-jscs and remove all overrides in the Gruntfile by fixing style violations (powered by JSCS).

This will likely cause conflicts for existing PRs, but most open PRs are stale and should just be closed. Thoughts on landing this?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,6 +3,7 @@ module.exports = function( grunt ) {
 "use strict";
 
 var
+
 	// files
 	coreFiles = [
 		"core.js",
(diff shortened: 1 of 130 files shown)


jzaefferer:
I don't like the idea of fixing only spacing issues across many files, while not fixing any indent issues (at least that's what I got at a glance). If we break a bunch of PRs, we should do it at least just once.

jzaefferer:
Also making any changes to datepicker.js in master seems like an even worse idea.

arschmitz:
Could you also back out changes to anything for button that will be a MASSIVE pain

scottgonzalez:
Backing out any changes means I don't want to do anything right now. The whole point of this is to clean up the build tasks.

arschmitz:
As far as button in the button PR its already removed from the bad and is passing. i was hoping to just not have to deal with a bunch of white space conflicts since every single line changes except for a few brackets.

That being said if it means having the whole lib passing ill deal with the conflicts. I cant speak about datepicker though.

scottgonzalez:
It can wait. We're close to getting things merged to master.

PR #1550, opened 2015-05-09
Selectmenu: Handle form resets

tjvantoll:
Fixes #12236


scottgonzalez:
We're actually building a mixin for this since multiple widgets need to handle this and we don't want to bind tons of event handlers to the same form for effectively the same operation (one for every instance). See https://github.com/scottgonzalez/jquery-ui/compare/3cc0f75%5E...8d51824 for the solutions we were looking at.

PR #1552, opened 2015-05-12
Adds ability to use function in the `pointAt` option

extempl:
Draggable: Adds ability to use function in the `pointAt` option

### DIFF
--- a/ui/draggable.js
+++ b/ui/draggable.js
@@ -385,6 +385,9 @@ $.widget("ui.draggable", $.ui.mouse, {
 		if ($.isArray(obj)) {
 			obj = { left: +obj[0], top: +obj[1] || 0 };
 		}
+		if ($.isFunction(obj)) {
+			obj = obj.call( this.element[ 0 ] )
+		}
 		if ("left" in obj) {
 			this.offset.click.left = obj.left + this.margins.left;
 		}


scottgonzalez:
Please just follow the process I linked to before.

extempl:
I trying. 
I already signed that CLA and followed http://contribute.jquery.org/commits-and-pull-requests/ page.

extempl:
Maybe the problem is that I signed CLA twice with different names (but with the same email), but I do not know now what CLA data is recorded and which I should use.

scottgonzalez:
You're missing some very crucial steps. For example:

> Unless you are making a rather minor change, it is generally a good idea to file an issue on the appropriate bug tracker, explaining your idea before writing code or submitting a PR, especially when introducing new features.

You're introducing a new feature here, so we absolutely must have a ticket and that ticket must explain why this feature should be added.

Also:

> Be sure to include new tests when fixing bugs or adding features.

As for the CLA check, you need to configure git to use your real name; it must match the CLA signature exactly.

extempl:
Ah. I thought, this IS minor change and "why" quite obvious.

extempl:
Also, I can't find any test related to `pointAt` option or `_adjustOffsetFromHelper` method.

extempl:
@scottgonzalez You know, I created issue, corrected my system preferences with name (anyway, wtf with that? Why just do not check email and that's it?), created another PR (https://github.com/jquery/jquery-ui/pull/1553), but... Uh. I won't do anything more here, I'm tired. You can close it, maybe someone else will do that as it should be.
Ah, yes, and there are problems with login (hangs on GET after login POST), posting issue (hangs captcha... wait, captcha? Seriously? Too much issues from 127.0.0.1? Oh, ok.). Use Youtrack. Or github.

Too much problems to commit 3-line code.

scottgonzalez:
> Also, I can't find any test related to pointAt option or _adjustOffsetFromHelper method.

That's because there's no such option as `pointAt` and internal methods aren't tested.

PR #1553, opened 2015-05-13
Draggable: Adds ability to use function in the `pointAt` option

extempl:
Draggable: Adds ability to use function in the `pointAt` option
Fixes #12588


scottgonzalez:
Commit messages are wrong, since there's no such option as `pointAt`. Also still waiting on an explanation of why this is needed (please respond in the ticket) and this needs tests if the feature is approved.

PR #1555, opened 2015-05-13
[WIP] Form reset mixin

scottgonzalez:
Everything is thrown together right now, but I wanted to get some discussion about the tests. Is this good enough or are there more/different things that should be tested?

Note: This is built on top of gh-1554.

### DIFF
--- a/tests/unit/form-reset-mixin/core.js
+++ b/tests/unit/form-reset-mixin/core.js
@@ -1,0 +1,97 @@
+define( [
+	"jquery",
+	"lib/common",
+	"ui/widget",
+	"ui/form-reset-mixin"
+], function( $, common ) {
+
+module( "widget factory", {
+	setup: function() {
+		$.widget( "ui.testWidget", [ $.ui.formResetMixin, {
+			_create: function() {
+				this._bindFormResetHandler();
+			},
+			_destroy: function() {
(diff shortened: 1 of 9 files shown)


tests/unit/form-reset-mixin/core.js
```diff
@@ -0,0 +1,97 @@
...
+			refresh: function() {
+				$.ui.testWidget.refreshed.push( this.element.attr( "id" ) );
+			}
+		} ] );
+
+		$.ui.testWidget.refreshed = [];
+	},
+
+	teardown: function() {
+		if ( $.ui ) {
```
jzaefferer: Is this `if` actually needed? Looks like `$.ui` these will always be defined.

jzaefferer:
This looks good to me. I considered checking that the data actually references the right instances, but the `$.ui.testWidget.refreshed` assertion implies that already. Can't think of anything else to test.

tests/unit/form-reset-mixin/core.js
scottgonzalez: Yeah, it's always defined. I had copied this from the widget tests, where it may not exist yet.

scottgonzalez:
Ok, I'll remove that one conditional in the tests and land this and gh-1554.

PR #1556, opened 2015-05-14
Remove event/alias and deprecated core module dependencies 

arschmitz:
This also adds an additional version of jQuery to the tests which excluded the event/alias and deprecated modules. 

Note: this does not fully work or pass tests until we also update mousewheel and simulate. The changes in both of these cases are trivial and just a couple lines.

### DIFF
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -309,14 +309,14 @@ $.Widget.prototype = {
 		// we can probably remove the unbind calls in 2.0
 		// all event bindings should go through this._on()
 		this.element
-			.unbind( this.eventNamespace )
+			.off( this.eventNamespace )
 			.removeData( this.widgetFullName );
 		this.widget()
-			.unbind( this.eventNamespace )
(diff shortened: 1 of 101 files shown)


ui/widget.js
```diff
@@ -531,9 +531,9 @@ $.Widget.prototype = {
 				eventName = match[ 1 ] + instance.eventNamespace,
 				selector = match[ 2 ];
 			if ( selector ) {
-				delegateElement.delegate( selector, eventName, handlerProxy );
+				delegateElement.on( eventName, selector, handlerProxy );
```
jzaefferer: Do we still need the `if` and two separate calls?

arschmitz: yup

scottgonzalez:
This makes me happy and sad. But the sad part is because of bad design in jQuery. This is good to land. Once the dependencies are updated, I'd like the custom build to become the default version that's used in tests.

PR #1557, opened 2015-05-14
Use AMD bootstrap in demos

arschmitz:
This adds an AMD bootstrap based off the one we use in tests for loading demos. For now this just switches the accordion and datepicker demos to cover the different use cases we have. PR for jqueryui.com to properly handle new demo layout coming shortly.

It also removes the accordion hover intent demo and a ie 6 /7 specific work around in the accordion demos


demos/accordion/fillspace.html
```diff
@@ -4,13 +4,9 @@
...
 	<link rel="stylesheet" href="../../themes/base/all.css">
-	<script src="../../external/jquery/jquery.js"></script>
-	<script src="../../ui/core.js"></script>
-	<script src="../../ui/widget.js"></script>
-	<script src="../../ui/mouse.js"></script>
-	<script src="../../ui/resizable.js"></script>
-	<script src="../../ui/accordion.js"></script>
 	<link rel="stylesheet" href="../demos.css">
+	<script src="../require-config.js"></script>
+	<script src="../../external/requirejs/require.js"></script>
```
jzaefferer: Let's clean up the mixing of `<script>` and `<style>` as well?

demos/accordion/default.html
```diff
@@ -4,15 +4,11 @@
...
-	<script src="../../external/jquery/jquery.js"></script>
-	<script src="../../ui/core.js"></script>
-	<script src="../../ui/widget.js"></script>
-	<script src="../../ui/accordion.js"></script>
 	<link rel="stylesheet" href="../demos.css">
-	<script>
-	$(function() {
+	<script src="../require-config.js"></script>
+	<script src="../../external/requirejs/require.js"></script>
+	<script src="../bootstrap.js">
```
jzaefferer: Let's merge `require-config.js` and `bootstrap.js`, use `require.config()` in it, after loading `require.js`.

demos/accordion/index.html
```diff
@@ -12,7 +12,6 @@
 	<li><a href="fillspace.html">Fill space</a></li>
 	<li><a href="no-auto-height.html">No auto height</a></li>
 	<li><a href="collapsible.html">Collapse content</a></li>
-	<li><a href="hoverintent.html">Open on hoverintent</a></li>
```
jzaefferer: It won't be missed.

demos/accordion/fillspace.html
arschmitz: +1 i thought about doing this but wanted to leave as is until i asked

demos/bootstrap.js
```diff
@@ -0,0 +1,25 @@
+(function() {
+
+	// Find the script element
+	var scripts = document.getElementsByTagName( "script" );
+	var script = scripts[ scripts.length - 1 ];
+
+	// Read the modules
+	var modules = script.getAttribute( "data-modules" );
+	var demoScripts = script.getAttribute( "data-modules" );
```
jzaefferer: Looks like an unused leftover, delete this line.

demos/bootstrap.js
```diff
@@ -0,0 +1,25 @@
...
+	var path = window.location.pathname;
```
jzaefferer: Use just once, inline.

demos/bootstrap.js
```diff
@@ -0,0 +1,25 @@
...
+	var pathParts = path.split( "/" );
+	var pathLength = pathParts.length;
```
jzaefferer: Used just once, inline.

demos/bootstrap.js
```diff
@@ -0,0 +1,25 @@
...
+	if ( modules ) {
```
jzaefferer: Could rewrite with tertiary:
```js
modules = modules ? modules.replace( /^\s+|\s+$/g, "" ).split( /\s+/ ) : [];
```

demos/datepicker/datepicker-fr.js
```diff
@@ -6,7 +6,7 @@
 	if ( typeof define === "function" && define.amd ) {
 
 		// AMD. Register as an anonymous module.
-		define([ "../jquery.ui.datepicker" ], factory );
+		define([ "datepicker" ], factory );
```
jzaefferer: I'd expect `"ui/datepicker"` here.

jzaefferer: Nevermind, `baseUrl: "../../ui"` already includes `ui`.

arschmitz: the require config sets the base to `../../ui`

demos/datepicker/localization.html
```diff
@@ -4,23 +4,15 @@
...
-	<script src="datepicker-ar.js"></script>
-	<script src="datepicker-fr.js"></script>
-	<script src="datepicker-he.js"></script>
-	<script src="datepicker-zh-TW.js"></script>
 	<link rel="stylesheet" href="../demos.css">
-	<script>
-	$(function() {
+	<script src="../require-config.js"></script>
+	<script src="../../external/requirejs/require.js"></script>
+	<script src="../bootstrap.js" data-modules="datepicker-ar.js datepicker-fr.js datepicker-he.js datepicker-zh-TW.js">
```
jzaefferer: Should these have the `.js` extension?

jzaefferer:
Overall looking good. Merging the two bootstrap files will make it even better.

demos/datepicker/localization.html
arschmitz: Yes its to load them via url and avoid the base since loading files from within the demos is rare ( maybe only this one ) where as requiring modules from ui is in many many demos

jzaefferer: Okay. Maybe add a note about that to bootstrap.js?

demos/bootstrap.js
jzaefferer: I'm wondering if we should stick with the pattern we use for tests, where we don't infer the "base" script, but already require at least one module to be specified.

arschmitz: my personal thought was in the case of demos this was fine we don't need to handle nearly the variety of situations in demos just compare the bootstrap file complexity

jzaefferer: Alright, let's stick with inferring then.

arschmitz:
@jzaefferer addressed your comments

jzaefferer:
Nice, looks good to me.

jzaefferer:
I've pushed this branch to `origin` (this repo) for testing with `jqueryui.com`.

demos/datepicker/datepicker-fr.js
scottgonzalez: I don't like that.

arschmitz: The datepicker localization demo, is the only demo in the whole repo, that has JavaScript files that are not within the `ui/` so i think it makes sense to use this as the `baseUrl`.

scottgonzalez: Well, regardless, this file can't be modified directly. It's currently out-of-date, but this needs to be an exact copy of the source file.

arschmitz: > this file can't be modified directly. It's currently out-of-date, but this needs to be an exact copy of the source file.

Where is the source file?

scottgonzalez: `ui/i18n`

arschmitz: I see why are we keeping the same file in 2 places?

scottgonzalez: So that we get the files during deployment to jqueryui.com.

arschmitz: That could easily be handled by the `build-demos` task in jqueryui.com

scottgonzalez: I'd be happy to have some other solution that avoids this problem. I'm just not sure what we'd do other than parsing the actual dependencies and determining when we grab files that aren't `/ui/*` so that we know they won't be in the CDN build.

arschmitz: This is already in my pull on jqueryui.com i already need to special case these files. So it wont be any harder to special case them in this way and it would simplify the bootstrap and avoid this issue completely because they would be within the baseUrl already.

arschmitz: I updated my PR on jqueryui.com to handle this as well as this PR to include removal of these from the demo and using the ones in the `ui/i18n` folder

scottgonzalez:
What's the status of this?

arschmitz:
@scottgonzalez Waiting for your review

demos/bootstrap.js
```diff
@@ -0,0 +1,25 @@
...
+			jquery: "../../external/jquery/jquery"
+		}
+	} );
+
+	modules = modules ? modules.replace( /^\s+|\s+$/g, "" ).split( /\s+/ ) : [];
+	modules.push( pathParts[ pathParts.length - 2 ] );
+
+	require( modules, function() {
+		$( "body" ).css( "visibility", "visible" );
+		eval( $( script ).html() );
```
scottgonzalez: It might be better to use `$.globalEval()` here.

demos/bootstrap.js
```diff
@@ -0,0 +1,25 @@
...
+	} );
+})();
```
scottgonzalez: Outdent the full function body.

arschmitz:
This is now update to cover all the demos

demos/bootstrap.js
```diff
@@ -0,0 +1,54 @@
...
+	"effect-fade",
+	"effect-fold",
+	"effect-highlight",
+	"effect-puff",
+	"effect-pulsate",
+	"effect-scale",
+	"effect-shake",
+	"effect-size",
+	"effect-slide"
+];
```
scottgonzalez: I'm worried about this getting out of date (though honestly, I don't anticipate more effects). I suppose it will be obvious if this gets out of date though. Either the new effect just won't work in the demos or the demos will throw an error if we delete an effect.

demos/bootstrap.js
```diff
@@ -0,0 +1,54 @@
...
+modules.push( pathParts[ pathParts.length - 2 ] );
+
+require( modules, function() {
+	$( "body" ).css( "visibility", "visible" );
+
+	// We wrap the code in an IIFE so that return statements work and we
+	// dont polute the global scope with variables
+	$.globalEval( "(function(){ " + $( script ).html() + "})();" );
+} );
+})();
```
scottgonzalez: Add a blank line above this.

jzaefferer:
None of the demos work for me. I load them, for example, as http://localhost/jquery-ui/demos/accordion/default.html

It fails to load http://localhost/external/jquery/jquery.js - note the missing "jquery-ui" part in the path.

I wanted to test `demos/effect/easing.html` since I don't understand how that loads the right modules.

jzaefferer:
Opening the globalize-dependent spinner demos causes errors when loading the locale files, e.g. demos/spinner/decimal.html throws "Uncaught TypeError: Cannot read property 'addCultureInfo' of undefined" in both globalize.culture.de-DE.js:28 and globalize.culture.ja-JP.js:28

The error doesn't always happen, so I guess sometimes they load in the "right" order, but mostly, I guess, this is failing due to missing dependencies from the culture files to the main globalize file.

arschmitz:
@jzaefferer thats odd works fine for me i originally had this problem but it should be fixed by 
https://github.com/jquery/jquery-ui/pull/1557/files#diff-7f4d985c65eaba301a9f9aadcba63f0cR34

jzaefferer:
That link points at a datepicker demo diff, probably not what you intended.

I guess you're pointing at the shim in bootstrap.js - that makes sense, though I've never seen it uses with a wildcard.

If I simulate a "slower" network ("Good 3G" is bad enough), I always get the error.

jzaefferer:
Oh, just saw your new commit, that should fix it.

demos/bootstrap.js
arschmitz: If we added a new effect it would also throw an error when you tried to run it if it was missing from this list.

demos/bootstrap.js
jzaefferer: Looks like this is the only remaining "issue".

jzaefferer:
Still need to add `demos/bootstrap.js` to lint tasks, and fix a bunch of issues in there. Maybe add `demos/**/*.js`?

arschmitz:
@jzaefferer will do

arschmitz:
@jzaefferer updated

demos/bootstrap.js
```diff
@@ -0,0 +1,60 @@
+/* globals window:true, document:true, $:true */
```
jzaefferer: `$` isn't used anymore.

jzaefferer: That's also why the Travis build failed.

arschmitz:
@jzaefferer updated to remove `$`

jzaefferer:
Tested this properly, with the 1.12 DB changes and https://github.com/jquery/jqueryui.com/pull/113. Still need to adjust the demo iframe and font-size, but that's a separate issue I'll deal with. For now, let's land this (and update core-brekaup on top) and land https://github.com/jquery/jqueryui.com/pull/113 in a 1-12 branch. I'll then update the 1-12 branch with the font-size fix, along with jquery-wp-content (probably also a ui-1-12 branch).

jzaefferer:
ping @arschmitz - let's finish this, as outlined above, will also massively simplify the demos-half of the core-breakup.

PR #1558, opened 2015-05-14
Draggable: Adds ability to use function in the `cursorAt` option

extempl:
Draggable: Adds ability to use function in the `cursorAt` option

Fixes #12588


extempl:
@scottgonzalez Tried to commit with forced name, but still with no luck.

PR #1559, opened 2015-05-18
Build: Stop testing against core 1.10.1

arschmitz:
This version is completely broken on testswarm

In IE accessing a cross domain iframe will throw an error in jQuery 1.10.1 and makes testswarm fail across the board in IE

Ref http://bugs.jquery.com/ticket/13980

Fixes #12745


mzgol:
:+1:

scottgonzalez:
But isn't this a *new* problem?

arschmitz:
It's as long as test swarm has logs for tried to reply on ticket but my password seems to have corrupted again :-/

arschmitz:
Testswarm is cross-domain, the main page is swarm.jquery.org, the frame is on builds.jenkins.jquery.com which triggers the bug i linked ( thank you m_gol for tracking this down ) not sure how this would have ever passed. I logged into testswarm with IE11 to confirm this was the bug causing this also. I also verified these test all pass on all IE's without times outs. This is specific to testswarm and that it runs in an iframe and cross domain.

scottgonzalez:
This is changing the wrong thing. We still need to be able to test against 1.10.1, but not submit it to TestSwarm. Discussed in IRC, @arschmitz will just make the correct change and push to master.

jzaefferer:
This: https://github.com/jquery/jquery-ui/blob/master/build/tasks/testswarm.js#L8

PR #1561, opened 2015-05-24
Frisian (fy) translation

robin0van0der0v:
I translated the Dutch localization to Frisian, a minority language in the Netherlands.


robin0van0der0v:
I did sign the CLA, see #1562.

scottgonzalez:
Thanks, but we're not accepting any new locales since we're in the process of switching to CLDR.

PR #1562, opened 2015-05-24
Translated to Talossan (tzl)

robin0van0der0v:
I translated the localization file to the Talossan language.
Source: my own knowledge in combination with the official word list


scottgonzalez:
Thanks, but we're not accepting any new locales since we're in the process of switching to CLDR.

PR #1565, opened 2015-06-17
Position: Guard against passing window to offset

arschmitz:
Starting in jQuery 3.0 this will throw an error

### DIFF
--- a/ui/position.js
+++ b/ui/position.js
@@ -142,12 +142,13 @@ $.position = {
 	getWithinInfo: function( element ) {
 		var withinElement = $( element || window ),
 			isWindow = $.isWindow( withinElement[ 0 ] ),
-			isDocument = !!withinElement[ 0 ] && withinElement[ 0 ].nodeType === 9;
+			isDocument = !!withinElement[ 0 ] && withinElement[ 0 ].nodeType === 9,
+			hasOffset = !isWindow && !isDocument;
 		return {
 			element: withinElement,
 			isWindow: isWindow,
 			isDocument: isDocument,
-			offset: withinElement.offset() || { left: 0, top: 0 },
+			offset: hasOffset ? $( element ).offset() : { left: 0, top: 0 },
 			scrollLeft: withinElement.scrollLeft(),
 			scrollTop: withinElement.scrollTop(),
 			width: withinElement.outerWidth(),


ui/position.js
```diff
@@ -142,12 +142,13 @@ $.position = {
...
 			isWindow = $.isWindow( withinElement[ 0 ] ),
-			isDocument = !!withinElement[ 0 ] && withinElement[ 0 ].nodeType === 9;
+			isDocument = !!withinElement[ 0 ] && withinElement[ 0 ].nodeType === 9,
+			hasOffset = element && ( !isWindow && !isDocument );
 		return {
 			element: withinElement,
 			isWindow: isWindow,
 			isDocument: isDocument,
-			offset: withinElement.offset() || { left: 0, top: 0 },
+			offset: hasOffset ? $( element ).offset() || false : false || { left: 0, top: 0 },
```
jzaefferer: What's with the `|| false`? That looks like it guards against the reverted change.

scottgonzalez: Also, what's with the `false ||`?

And why isn't this using `withinElement` anymore?

ui/position.js
```diff
@@ -142,12 +142,13 @@ $.position = {
 	getWithinInfo: function( element ) {
 		var withinElement = $( element || window ),
 			isWindow = $.isWindow( withinElement[ 0 ] ),
-			isDocument = !!withinElement[ 0 ] && withinElement[ 0 ].nodeType === 9;
+			isDocument = !!withinElement[ 0 ] && withinElement[ 0 ].nodeType === 9,
+			hasOffset = element && ( !isWindow && !isDocument );
```
scottgonzalez: Why are these grouped? All the operations are commutative. I'd only expect a group if you were negating the group.

arschmitz: These are variable definitions?

arschmitz:
This needs another review

ui/position.js
jzaefferer: I think Scott meant that the parens are unnecessary. Ergo
```js
hasOffset = element && !isWindow && !isDocument;

File: ui/position.js [hunk 3]
```diff
@@ -142,12 +142,13 @@ $.position = {
...
 		return {
 			element: withinElement,
 			isWindow: isWindow,
 			isDocument: isDocument,
-			offset: withinElement.offset() || { left: 0, top: 0 },
+			offset: hasOffset ? $( element ).offset() : { left: 0, top: 0 },
```
jzaefferer: Not addressed, from Scott's comment: Why isn't this using `withinElement` anymore?

ui/position.js
jzaefferer: As below, why isn't this using `withinElement` instead of `element`?

jzaefferer:
Reviewed!

ui/position.js
arschmitz: Because the fact that we were using `withinElement` IS THE BUG it the entire reason for the change. `withinElement` can be window or document and both of these must be guarded against as they will throw errors. Hence the check your commenting on.

jzaefferer: That makes sense. Might be worth a comment in the code, maybe not...

scottgonzalez: Can't this just be reduced to `!isWindow && !isDocument`? It doesn't matter if the user provided the original element.

arschmitz: Yup that should be fine

arschmitz:
Fixed last comment

jzaefferer:
Looks good.

scottgonzalez:
Looks good.

jzaefferer:
Merge it!

PR #1566, opened 2015-06-18
maybe a typo

spat-ne-hochu:
Uncaught TypeError: this.menu.trigger is not a function (autocomplete.js:265)


jzaefferer:
Not quite a typo, but definitely wrong. Was introduced here, needs to be reverted to just call menu's `.blur()`: https://github.com/jquery/jquery-ui/commit/8b4ce807cd97e3cb953995934d6c4f614de9fa03#diff-312047f9a8fc4a32a6429caa4a07d9f3L265

jzaefferer:
Replaced by gh-1568

PR #1567, opened 2015-06-23
Droppable - Wrong Scaled Element Dimensions

anthonyjimenez:
http://bugs.jqueryui.com/ticket/13278

getBoundingClientRect() returns the correct height and width for scaled values and also for non-scaled elements. It also takes into account all the widths added by margins, offset, padding... Source: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect


scottgonzalez:
jQuery UI should never go down to this level. You've already discovered that this a feature request for jQuery core, which has been closed as won't fix.

PR #1568, opened 2015-07-01
Autocomplete: Revert mistaken event alias change

jzaefferer:
This is a call on the menu widget, not using an event alias.

Ref 8b4ce807cd97e3cb953995934d6c4f614de9fa03
Ref #12770

After staring at the surrounding code for a while, I figured that its not worth the effort to try and add a unit test.

Replaces gh-1566

### DIFF
--- a/ui/autocomplete.js
+++ b/ui/autocomplete.js
@@ -262,7 +262,7 @@ $.widget( "ui.autocomplete", {
 				if ( this.isNewMenu ) {
 					this.isNewMenu = false;
 					if ( event.originalEvent && /^mouse/.test( event.originalEvent.type ) ) {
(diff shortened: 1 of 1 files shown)


arschmitz:
:+1:  This was my mistake in the original PR

jzaefferer:
@scottgonzalez okay with leaving this test-less?

scottgonzalez:
Sure. One day I'll rewrite all of this logic and add a test ;-) We've actually diverged from the Awesome Bar, so we have some UX improvements to make that should change this logic.

arschmitz:
on the bright side not using alias's any more makes it so we cant confuse methods like this in the future!

PR #1569, opened 2015-07-01
Core: Split up into individual modules

arschmitz:
This breaks the core file up into small individual modules and updates all their dependencies


jzaefferer:
How about putting these into their own folder? Or moving widgets into a folder? Something something with less clutter?

arschmitz:
In mobile we have a widgets folder

arschmitz:
Though moving the core stuff into its own folder is a lot easier since it wont require changes to every test file :-)

ui/data.js
```diff
@@ -0,0 +1,39 @@
+/*!
```
jzaefferer: This should mention the deprecation somewhere.

jzaefferer:
I'd be fine with a `core` folder. Or shared, utils, platform, helpers, plugins, tools...

arschmitz:
This now moves the widgets into their own folder and the effects into their own.

ui/data.js
jzaefferer: Bump

jzaefferer:
Folder changes look good.

This still deletes `ui/core.js`, ticket supposedly said to keep that with imports to all the extracted modules (Trac is down, can't check).

There's missing deprecation comments for `disable/enableSelection` and `$.ui.ie` (not sure why I had that comment on `data.js`, didn't belong there).

arschmitz:
@jzaefferer ill add the comments and file but this actually is missing updating all the demos adding that right now.

jzaefferer:
Why do some of the new files have copyright and AMD-property headers (like "//>>label"), others don't, e.g. `ui/form.js`?

arschmitz:
I followed what the convention seemed to be already if its a file that is a "public" method that will show in DB it gets a header if its a private micro module it does not.

jzaefferer:
There's no file that specifies `ui/version.js` as a dependency. I guess the recreated `ui/core.js` will, but that is also just temporary. How will the file get included in a UI build afterwards?

arschmitz:
@jzaefferer thats correct nothing should depend on. IT will be in the core. After it will have to be hard coded into the build. How is this handled today if you only select components that do not depend on core?

jzaefferer:
I guess its just missing. I'd be fine with moving it back to `core.js` and just deleting it entirely when the time comes.

jzaefferer:
Reviewed the rest of the commits, didn't notice any other issues. Will do some actual testing when demos are updated.

arschmitz:
@scottgonzalez what are your thoughts since you said to make this its own module?

arschmitz:
@jzaefferer demos are updated

scottgonzalez:
I don't actually have a good answer for this. I'd say that we should just make version a dependency on everything, but I'm not sure if that'd be annoying for users doing custom builds.

jzaefferer:
Can we make this file (more) useful by having everything depend on it, and drop the `$.ui` definition everywhere else?

scottgonzalez:
Sounds ok to me.

arschmitz:
Widgets already have @version defined and define their namespace if it does not exist. This is also not a "dependency" things function just fine with out it. I think it would make since for all the "core" files to depend on it maybe but not widgets.

scottgonzalez:
It's not a dependency in terms of needed functionality, but we always expect the version to be there. I'd prefer to have the widgets declare it as a dependency as well.

arschmitz:
@scottgonzalez it seems like we should remove it from the individual widgets then as it just seems redundant.

scottgonzalez:
It's not redundant. The version is in the prototype so that you can check the version on an instance. This is useful since you can load multiple versions simultaneously.

arschmitz:
@jzaefferer this is good to go again full passing tests on all commits

ui/core.js
```diff
@@ -14,334 +14,23 @@
...
-		for ( i = 0; i < set.length; i++ ) {
-			if ( instance.options[ set[ i ][ 0 ] ] ) {
-				set[ i ][ 1 ].apply( instance.element, args );
-			}
-		}
-	}
-};
-
-} ) );
+( function() {
```
jzaefferer: Should mark this file as deprecated.

tests/unit/core/selector.js
```diff
@@ -1,6 +1,10 @@
 define( [
 	"jquery",
-	"ui/core"
+	"ui/core",
```
jzaefferer: This looks wrong. We shouldn't depend on `ui/core` anywhere anymore. This is the only one.

arschmitz: yes this should have been removed when tabbable was added

demos/bootstrap.js
```diff
@@ -40,13 +81,16 @@ require.config( {
...
+
 // Replace effects all shortcut modules with all the effects modules
 if ( modules && modules.indexOf( "effects-all" ) !== -1 ) {
 	modules = modules.replace( /effects-all/, effectsAll.join( " " ) );
 }
 
 modules = modules ? modules.replace( /^\s+|\s+$/g, "" ).split( /\s+/ ) : [];
-modules.push( pathParts[ pathParts.length - 2 ] );
+component = pathParts[ pathParts.length - 2 ];
+modules.push( component );
```
jzaefferer: The `component` var seems pointless, just inline it here.

demos/bootstrap.js
```diff
@@ -9,21 +10,61 @@ var script = scripts[ scripts.length - 1 ];
...
+
+function getPath( module ) {
+	for ( var i = 0; i < widgets.length; i++ ) {
+		if ( widgets[ i ] === module ) {
+			return "widgets/" + module;
+		}
+	}
+	for ( var j = 0; j < effectsAll.length; j++ ) {
+		if ( effectsAll[ j ] === module ) {
+			return "effects/" + module;
```
jzaefferer: This looks strange - the logic is the same as for the `widgets` list, but the "effects/" prefix is part of the list, don't you end up with "effects/effects/..."?

ui/disable-selection.js
```diff
@@ -0,0 +1,46 @@
+/*!
+ * jQuery UI Disable Selection @VERSION
```
jzaefferer: This is deprecated

demos/bootstrap.js
arschmitz: There is an error here it needs to check if it contains the string not `===` because this is processing when a demo requires a specific effect not all the effects.

ui/escape-selector.js
```diff
@@ -0,0 +1,21 @@
...
+		define( [ "jquery", "./version" ], factory );
+	} else {
+
+		// Browser globals
+		factory( jQuery );
+	}
+} ( function( $ ) {
+
+// Internal use only
+return $.ui.escapeSelector = ( function() {
```
jzaefferer: Why do we still export it, if its for internal use only?

ui/ie.js
```diff
@@ -0,0 +1,13 @@
...
+
+		// AMD. Register as an anonymous module.
+		define( [ "jquery", "./version" ], factory );
+	} else {
+
+		// Browser globals
+		factory( jQuery );
+	}
+} ( function( $ ) {
+return $.ui.ie = !!/msie [\w.]+/.exec( navigator.userAgent.toLowerCase() );
```
jzaefferer: Deprecated

ui/keycode.js
```diff
@@ -0,0 +1,46 @@
+/*!
+ * jQuery UI Keycodes @VERSION
```
jzaefferer: Keep `keycode` singular? Also applies to label below.

ui/keycode.js
```diff
@@ -0,0 +1,46 @@
...
+ * http://jqueryui.com
+ *
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: Keycodes
+//>>group: Core
+//>>description: Maps keycodes for events
```
jzaefferer: "Provide keycodes as keynames"?

ui/keycode.js
```diff
@@ -0,0 +1,46 @@
...
+//>>docs: http://api.jqueryui.com/category/effects-core/
```
jzaefferer: Wrong URL, also the next one.

ui/scroll-parent.js
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: scrollParent
+//>>group: Core
+//>>description: Get the closest ancestor element that is scrollable.
+//>>docs: http://api.jqueryui.com/category/scrollParent/
+//>>demos: http://jqueryui.com/scrollParent/
```
jzaefferer: These don't exist. And I doubt that we'll write a demo for this. API docs seem fine, but it won't be a category.

ui/tabbable.js
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: focusable
+//>>group: Core
+//>>description: Selects elements which can be tabbed to.
+//>>docs: http://api.jqueryui.com/tabbable-selector/
+//>>demos: http://jqueryui.com/tabbable/
```
jzaefferer: Docs URL is correct, demo doesn't exist, and probably shouldn't.

ui/unique-id.js
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: uniqueId
+//>>group: Core
+//>>description: Functions to generate and remove uniqueId's
+//>>docs: http://api.jqueryui.com/category/uniqueId/
+//>>demos: http://jqueryui.com/uniqueId/
```
jzaefferer: Both wrong

jzaefferer:
Made it to the end. No structural issues, but some details that need to be addressed.

ui/escape-selector.js
arschmitz: Im not sure what you mean what other namespace would we put it on?

jzaefferer: Nevermind

ui/keycode.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: Keycode
+//>>group: Core
+//>>description: Provide keycodes as keynames
+//>>docs: http://api.jqueryui.com/category/keycodes/
```
jzaefferer: Still wrong URL, the current documentation is at http://api.jqueryui.com/jQuery.ui.keyCode/

ui/keycode.js
```diff
@@ -0,0 +1,46 @@
...
+//>>demos: http://jqueryui.com/keycodes/
```
jzaefferer: This doesn't exist and probably never will.

ui/data.js
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: :data
+//>>group: Core
+//>>description: Selects elements which have data stored under the specified key.
+//>>docs: http://api.jqueryui.com/data-selector/
+//>>demos: http://jqueryui.com/data/
```
jzaefferer: Remove

ui/disable-selection.js
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: disableSelection
+//>>group: Core
+//>>description: Disable selection of text content within the set of matched elements.
+//>>docs: http://api.jqueryui.com/disableSelection/
+//>>demos: http://jqueryui.com/disableSelection/
```
jzaefferer: Remove

ui/focusable.js
```diff
@@ -0,0 +1,65 @@
...
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: focusable
+//>>group: Core
+//>>description: Selects elements which can be focused.
+//>>docs: http://api.jqueryui.com/focusable-selector/
+//>>demos: http://jqueryui.com/focusable/
```
jzaefferer: Remove

ui/labels.js
```diff
@@ -0,0 +1,63 @@
...
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: labels
+//>>group: Core
+//>>description: Find all the labels associated with a given input
+//>>docs: http://api.jqueryui.com/category/labels/
+//>>demos: http://jqueryui.com/labels/
```
jzaefferer: Remove

ui/labels.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: labels
+//>>group: Core
+//>>description: Find all the labels associated with a given input
+//>>docs: http://api.jqueryui.com/category/labels/
```
jzaefferer: Should be "http://api.jqueryui.com/labels/", also need to file a ticket to create that page.

jzaefferer:
A bunch more wrong URLs, but I think I found them all.

arschmitz:
@jzaefferer updated

jzaefferer:
What do we need the remaining concat grunt task for? The output isn't used anywhere. Whoever built the 1.10 npm package still hasn't updated it to 1.11, so there's no point in worrying about that - though we should be publishing jQuery UI on npm...

ui/core.js
```diff
@@ -14,334 +14,24 @@
 //>>docs: http://api.jqueryui.com/category/ui-core/
 //>>demos: http://jqueryui.com/
```
jzaefferer: Let's remove this as well.

ui/labels.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Copyright jQuery Foundation and other contributors
+ * Released under the MIT license.
+ * http://jquery.org/license
+ */
+
+//>>label: labels
+//>>group: Core
+//>>description: Find all the labels associated with a given input
+//>>docs: http://api.jqueryui.com/labels/
```
jzaefferer: I've filed https://github.com/jquery/api.jqueryui.com/issues/278 for this. Is the "ponyfill for `input.labels`" description accurate enough?

arschmitz: Yup. Though I question if most people will know what a ponyfill is.

jzaefferer: That's okay, we don't have to use that term in the docs.

arschmitz:
@jzaefferer anything else on this?

jzaefferer:
Demos link in ui/core.js and grunt concat task, see comments above.

jzaefferer:
Remove the now unused dependency, then this is good to land.

arschmitz:
Doing it now

arschmitz:
Closed in eeb962027e251dff9ae37e78934f56107f47951f

PR #1571, opened 2015-07-02
Fix a bug when setting the width of the object dynamically

danielravina:
Related to this issue here:
http://stackoverflow.com/questions/17820381/weird-bug-with-jquery-ui-resizable/31174197#31174197


scottgonzalez:
Thanks for the patch, but please start by reading the [contribution guidelines](https://github.com/jquery/jquery-ui/blob/master/CONTRIBUTING.md). We need a ticket describing the problem, ideally with a reduced test case. We'll also need unit tests showing that the patch fixes the problem. Please also make sure to sign the CLA.

ui/resizable.js
```diff
@@ -344,16 +344,16 @@ $.widget("ui.resizable", $.ui.mouse, {
...
-				height: el.height()
+				width: el.outerWidth(),
+				height:el.outerHeight()
 			};
 
 		this.originalSize = this._helper ? {
 				width: el.outerWidth(),
 				height: el.outerHeight()
 			} : {
-				width: el.width(),
```
dmethvin: I think the clue that there is something more going on here not addressed by the PR is that this change essentially negates the need for the `this._helper` check--the two objects are now the same! Also, you would want a test case that shows this change actually fixes a situation that was broken.

danielravina:
Found the real source of the problem.
I set `padding: 12px` on my `.ui-resizable` which screwed up the calculation. outerWidth, and outerHeight ignores the padding; that's why my change is working for me and it's not passing the test.
I will close it. Thanks!

PR #1573, opened 2015-07-06
slider: adding missing unit test cases for values method

dekajp:


jzaefferer:
Updated this to split the test into two separate tests, one for `step: 1`, one for `step:3 `. Also updated the commit message. Thanks!

PR #1576, opened 2015-07-09
Added belgian french locale for datepicker

YevheniiPuzyrenko:
Added  belgian french locale for datepicker.


scottgonzalez:
Thanks, but we're moving to CLDR for all locale data. As a result, we're not accepting any new locales.

PR #1513, opened 2015-03-19
Button Re-write

arschmitz:
Replaces #1415

PR #1577, opened 2015-07-11
Tests: Shift all tests to use no globals

apsdehal:
This PR shifts all the tests to use no global QUnit variables and methods.

### DIFF
--- a/tests/lib/common.js
+++ b/tests/lib/common.js
@@ -1,6 +1,7 @@
 define( [
-	"jquery"
-], function( $ ) {
+	"qunit",
+	"jquery" ],
+function( QUnit, $ ) {
 
 var exports = {};
 
@@ -8,39 +9,39 @@ function testWidgetDefaults( widget, defaults ) {
 	var pluginDefaults = $.ui[ widget ].prototype.options;
 
 	// Ensure that all defaults have the correct value
-	test( "defined defaults", function() {
+	QUnit.test( "defined defaults", function( assert ) {
 		var count = 0;
 		$.each( defaults, function( key, val ) {
-			expect( ++count );
+			assert.expect( ++count );
 			if ( $.isFunction( val ) ) {
-				ok( $.isFunction( pluginDefaults[ key ] ), key );
+				assert.ok( $.isFunction( pluginDefaults[ key ] ), key );
 				return;
 			}
-			deepEqual( pluginDefaults[ key ], val, key );
+			assert.deepEqual( pluginDefaults[ key ], val, key );
 		} );
 	} );
 
 	// Ensure that all defaults were tested
-	test( "tested defaults", function() {
+	QUnit.test( "tested defaults", function( assert ) {
 		var count = 0;
 		$.each( pluginDefaults, function( key ) {
-			expect( ++count );
-			ok( key in defaults, key );
+			assert.expect( ++count );
+			assert.ok( key in defaults, key );
 		} );
 	} );
 }
 
 function testWidgetOverrides( widget ) {
 	if ( $.uiBackCompat === false ) {
-		test( "$.widget overrides", function() {
-			expect( 4 );
+		QUnit.test( "$.widget overrides", function( assert ) {
+			assert.expect( 4 );
 			$.each( [
 				"_createWidget",
 				"destroy",
 				"option",
 				"_trigger"
 			], function( i, method ) {
-				strictEqual( $.ui[ widget ].prototype[ method ],
+				assert.strictEqual( $.ui[ widget ].prototype[ method ],
 					$.Widget.prototype[ method ], "should not override " + method );
 			} );
 		} );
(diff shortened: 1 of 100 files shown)


jzaefferer:
For the Intern migration this looks good. For the full migration to the QUnit 2.0 API its still missing asyncTest(), which I've filed an issue for in the underlying tool: https://github.com/apsdehal/qunit-migrate/issues/4

We still have a work-in-progress PR that changes all test defines (#1569) and one that changes a ton of tests for one widget (#1513). @arschmitz tells me that the changes here are all automated, so to avoid merge conflicts we should land the other two PRs first (should happen in the next two weeks), then apply the automatic changes once more to do the migration.

jzaefferer:
@apsdehal we've landed the button rewrite in master. And you've addressed the `asyncTest()` migration. Could you update this PR based on the current master branch? Thanks.

apsdehal:
Will do it ASAP.

scottgonzalez:
@apsdehal I see you pushed some updated a few days ago. Is this now ready for another review?

apsdehal:
@scottgonzalez No, not yet. I will ping you whenever this is ready.

apsdehal:
@scottgonzalez It is ready for a round of review. Only test failing is one in effects, which if I change to asyncTest back works smoothly.

jzaefferer:
The issue is in the ".hide() with step" test in effects/core.js, which passes `start` as a complete callback. Needs to be replaced.

apsdehal:
@jzaefferer Issue has been fixed now and the build is passing.

tests/unit/effects/core.js
```diff
@@ -132,18 +137,19 @@ asyncTest( "animateClass works with colors", function() {
...
-			present( test.css( "backgroundColor" ), [ "#ffffff", "#fff", "rgb(255, 255, 255)" ], "Color is final" );
-			start();
+			assert.present( test.css( "backgroundColor" ), [ "#ffffff", "#fff", "rgb(255, 255, 255)" ], "Color is final" );
+			ready();
 		}
 	} );
 } );
 
-asyncTest( "animateClass calls step option", 1, function() {
+QUnit.test( "animateClass calls step option", 1, function( assert ) {
```
jzaefferer: Could you convert the expect argument to `assert.expect` calls as well?

jzaefferer: Could you convert the expect argument to assert.expect calls as well?

jzaefferer: Could you convert the expect argument to assert.expect calls as well? (3rd attempt)

jzaefferer:
I don't now why GitHub instantly marks by diff comment as outdated. So:

> +QUnit.test( "animateClass calls step option", 1, function( assert ) {

Could you convert the expect argument to assert.expect calls as well?

apsdehal:
@jzaefferer Fixed.

jzaefferer:
I guess those 2 instances were the only ones still in master, nice.

apsdehal:
Yes, I ran a regex check on the whole repo. Those were the only two.

tests/lib/common.js
```diff
@@ -87,9 +88,10 @@ exports.testJshint = function( module ) {
 		return;
 	}
 
-	asyncTest( "JSHint", function() {
-		require( [ "jshint" ], function() {
-			expect( 1 );
+	QUnit.test( "JSHint", function( assert ) {
+        var ready = assert.async();
+        require( [ "jshint" ], function() {
```
scottgonzalez: Convert indentation to tabs.

scottgonzalez:
Please add a commit at the end that removes all the QUnit globals from https://github.com/jquery/jquery-ui/blob/435ed7e8d2e0221635494eedfbe22616f2a4814f/tests/.jshintrc#L20

apsdehal:
@scottgonzalez I have addressed your comments.

jzaefferer:
Testing against custom QUnit build (pre-2.0) based on https://github.com/jquery/qunit/pull/976 

`tests/unit/all.html` isn't working due to https://github.com/JamesMGreene/qunit-composite/issues/9

At least `tests/unit/dialog/dialog.html` and `tests/unit/effects/effects.html` fail due to https://github.com/JamesMGreene/qunit-assert-close/issues/4

There might be other failures, but its hard to tell with the many errors coming from `qunit-assert-close`.

@apsdehal would you be interested in patching qunit-composite, qunit-assert-close or both? That would help a lot.

apsdehal:
@jzaefferer Sure, will do it!

jzaefferer:
@scottgonzalez updating those externals shouldn't block this, after all the PR works just fine, and QUnit 2.0 isn't even released. Based on my testing with QUnit pre-2.0, the changes here work fine. Unless I'm missing something, we can merge this and release rc2. (I'm not sure why I thought that rc2 should wait).

scottgonzalez:
@jzaefferer I'm not sure what externals you're referring to. Did you mean the globals? @apsdehal already removed them from the JSHint config. I asked for that to be part of this PR to ensure that none of the globals were missed during the update.

jzaefferer:
Thanks a lot @apsdehal!

PR #1578, opened 2015-07-16
Sortable: Setting table row placeholder height to be same as sorted row

kevincupp:
jQuery ticket: http://bugs.jqueryui.com/ticket/13662

Past history that led to the issue: https://github.com/jquery/jquery-ui/commit/bd47bd4ace3789d9eb302b0dce6f6e042d08a7f1 https://github.com/jquery/jquery-ui/commit/9711c54c6d3d7ecffa9bfccc205522be1f4aa148

Here is my proposed fix. It was difficult coming up with a solution that would mostly preserve previous behavior while also fixing this, but I ultimately ended up using the `forcePlaceholderSize` config. My thinking was if you've set that, any CSS you've hoped would apply to table row height wouldn't matter and we could safely set an inline height.

### DIFF
--- a/tests/unit/sortable/options.js
+++ b/tests/unit/sortable/options.js
@@ -253,15 +253,54 @@ test("{ dropOnEmpty: true }, default", function() {
 test("{ dropOnEmpty: false }", function() {
 	ok(false, "missing test - untested code is broken code.");
 });
+*/
 
-test("{ forcePlaceholderSize: false }, default", function() {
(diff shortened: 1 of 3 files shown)


kevincupp:
I should note also another option is to go back to the pre-1.10 behavior of setting an inline height on placeholder table rows anyway. If folks don't want it, they can set a placeholder class. I'd prefer this, but recognized it would change current behavior slightly and may not always be preferable.

PR #1579, opened 2015-07-22
Mouse: Fix delay timeout clearing upon mouseup

lsharir:
A timeout is created when a mousedown is called in ui-mouse, but it is not cleared when a mouseup is fired.

When executing multiple clicks shorter than the duration of the delay, all hell breaks loose, and the functionality of the delay option is unclear.

This fix allows the user to start fresh with a new delay timeout whenever a mousedown is fired.

### DIFF
--- a/ui/mouse.js
+++ b/ui/mouse.js
@@ -177,6 +177,10 @@ return $.widget("ui.mouse", {
 			this._mouseStop(event);
 		}
 
+		if (this.options.delay && this._mouseDelayTimer) {
+			clearTimeout(this._mouseDelayTimer);
+		}
+
 		mouseHandled = false;
 		return false;
 	},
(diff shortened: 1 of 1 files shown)


arthurvr:
You'll need to sign [the CLA](http://contribute.jquery.org/CLA/status/?repo=jquery-ui&sha=b61397da963c7252bba28eec10b887261b49230f).

NirHemed:
:+1: I really need this fix

yotmic:
Required fix +1

shaibo:
This really helps in making our UI responsive

royresh:
We really need it

eg4000:
Good fix!

ui/mouse.js
```diff
@@ -177,6 +177,10 @@ return $.widget("ui.mouse", {
 			this._mouseStop(event);
 		}
 
+		if (this.options.delay && this._mouseDelayTimer) {
```
scottgonzalez: Since this property is never cleared out, the check isn't necessary.

lsharir:
That's right @scottgonzalez ,
Thank you

PR #1583, opened 2015-08-05
Core: Remove ancestor visibility requirement from :focusable selector

westonruter:
This is news to me, but it turns out that an element that has an ancestor with `visibility:hidden` can itself have `visibility:visible` and it will show that nested element even thought everything in the wrapped elements up to the ancestor will be hidden. Example: http://jsfiddle.net/westonruter/xuuzv5cu/

This being the case, the `:focusable` selector needs to remove the check for ancestors being visible.

This issue was discovered in a ticket on WordPress Core: https://core.trac.wordpress.org/ticket/33258#comment:10

### DIFF
--- a/tests/unit/core/selector.js
+++ b/tests/unit/core/selector.js
@@ -125,13 +125,16 @@ test( "focusable - disabled elements", function() {
 } );
 
 test( "focusable - hidden styles", function() {
-	expect( 8 );
+	expect( 10 );
 
 	isNotFocusable( "#displayNoneAncestor-input", "input, display: none parent" );
 	isNotFocusable( "#displayNoneAncestor-span", "span with tabindex, display: none parent" );
 
 	isNotFocusable( "#visibilityHiddenAncestor-input", "input, visibility: hidden parent" );
 	isNotFocusable( "#visibilityHiddenAncestor-span", "span with tabindex, visibility: hidden parent" );
+
+	isFocusable( "#nestedVisibilityOverrideAncestor-input", "input, visibility: visible parent but visibility: hidden grandparent" );
+	isFocusable( "#nestedVisibilityOverrideAncestor-span", "span with tabindex, visibility: visible parent but visibility: hidden grandparent " );
 
 	isNotFocusable( "#displayNone-input", "input, display: none" );
 	isNotFocusable( "#visibilityHidden-input", "input, visibility: hidden" );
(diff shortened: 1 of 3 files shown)


arschmitz:
@westonruter we updated the core file last week ( there is not really a core file any more ) breaking it up into individual modules. Could you rebase / update this to account for that please?

westonruter:
@arschmitz done!

jzaefferer:
Diff looks good to me.

@arschmitz want to take a look as well? I guess we can take the PR description, put it in a Trac ticket a reference it when merging this PR.

arschmitz:
This looks good to me and will be faster which is good for an already slow selector

tests/unit/core/selector.js
```diff
@@ -130,14 +130,14 @@ test( "focusable - hidden styles", function() {
 	isNotFocusable( "#displayNoneAncestor-input", "input, display: none parent" );
 	isNotFocusable( "#displayNoneAncestor-span", "span with tabindex, display: none parent" );
 
-	isNotFocusable( "#visibilityHiddenAncestor-input", "input, visibility: hidden parent" );
-	isNotFocusable( "#visibilityHiddenAncestor-span", "span with tabindex, visibility: hidden parent" );
+	isFocusable( "#visibilityHiddenAncestor-input", "input, visibility: hidden parent" );
+	isFocusable( "#visibilityHiddenAncestor-span", "span with tabindex, visibility: hidden parent" );
```
scottgonzalez: This is definitely not correct. You cannot focus something that's not visible.

scottgonzalez:
Every assertion that was changed is incorrect. In addition, the scenario described above isn't actually tested.

tests/unit/core/selector.js
westonruter: @scottgonzalez Something can be visible even if (unexpectedly) its ancestor is not visible. See http://jsfiddle.net/westonruter/xuuzv5cu/

scottgonzalez: I understand that, but this is not what's happening here.

jzaefferer: @westonruter could you look into this again? Thanks.

westonruter: I guess I don't understand what is happening here then.

scottgonzalez: None of these elements have `visibility: visible` on them, so they're all still hidden.

westonruter:
Thanks all. I did a deeper dive into this and fixed the issues you identified. Part of my problem was I wasn't that `:visible` isn't a straight equivalent to checking if `visibility` is `visible`. I also added tests to demonstrate the nested visibility override behavior, and rebased against the latest from master.

scottgonzalez:
Thanks. I've filed a jQuery UI ticket for this and merged it.

scottgonzalez:
This change caused lots of failures in IE 8.

scottgonzalez:
The IE 8 failure was caused by an incorrect default of `visibility: inherit` instead of `visibility: visible`. Honestly, I've always thought the spec was wrong and the IE 8 behavior was correct, but there's nothing we can do about that. This did reveal an actual bug in our code though, since we weren't handling the case of `visibility: inherit`. I've put together a patch in #1605.

PR #1584, opened 2015-08-06
Make selectmenu play nice in iPad

Serabe:
We were having problems with `selectmenu` in iPad due to it not being closed when tapping outside the area. This fixes it.

### DIFF
--- a/ui/selectmenu.js
+++ b/ui/selectmenu.js
@@ -418,7 +418,7 @@ return $.widget( "ui.selectmenu", {
 	},
 
 	_documentClick: {
-		mousedown: function( event ) {
+		"mousedown touchend": function( event ) {
 			if ( !this.isOpen ) {
 				return;
 			}


Serabe:
Just signed the jQuery Foundation CLA a few moments ago, how long does it take to update the check?

arschmitz:
@Serabe your name does not match you can see the error by clicking the details link http://contribute.jquery.org/CLA/status/?repo=jquery-ui&sha=d7104c4425c4e594ce60afca6d4e35a934f09a06

Regarding selectmenu jQuery ui does not currently support touch events and sadly it will take much bigger changes then this for us to do so. We plan to do so by using PEP https://github.com/jquery/pep and just switching to pointer events. We will not be adding any touch specific event listeners, so I'm going to close this.

Temporarily you can use jQuery UI touch punch http://touchpunch.furf.com/ which should fix any touch related issues.

Serabe:
Thank you, I'll go with that

arschmitz:
@Serabe oh sorry and to actually answer your question about how long it takes to update the CLA status takes ~1 minute to update after you sign as long as everything matches.

PR #1585, opened 2015-08-08
Sortable: Fix position detection if browser's window is scrolled

Tomas-M:
If the browsers window is scrolled, and some element is dragged between several connected containers, it had a problem in detection where to insert the placeholder. This patch fixes it.

Detailed description of the issue including a picture is here:
http://stackoverflow.com/questions/31887539/jquery-ui-sortable-on-scrolled-webpage

### DIFF
--- a/ui/widgets/sortable.js
+++ b/ui/widgets/sortable.js
@@ -907,7 +907,7 @@ return $.widget("ui.sortable", $.ui.mouse, {
 			floating = innermostContainer.floating || this._isFloating(this.currentItem);
 			posProperty = floating ? "left" : "top";
 			sizeProperty = floating ? "width" : "height";
-			axis = floating ? "clientX" : "clientY";
+			axis = floating ? "pageX" : "pageY";
 
 			for (j = this.items.length - 1; j >= 0; j--) {
 				if(!$.contains(this.containers[innermostIndex].element[0], this.items[j].item[0])) {


arschmitz:
@Tomas-M Can you please sign the CLA and make sure the name matches your git config

Tomas-M:
ok done

Tomas-M:
I am unable to 'refresh' the CLA check here. And I don't care, really. I just tried to help ....

jzaefferer:
@Tomas-M you've signed with "Tomas-M", but we need a full name on the CLA and it needs to match your git author ("Tomas Matejicek"). Could you please try once more? Thanks.

Tomas-M:
I actually did both!
Tomas-M
and
Tomas Matejicek

Tried again, please retry. Anyway, I really do not understand why this madness...

jzaefferer:
@Tomas-M sorry for the trouble. We're still improving the CLA checker to avoid this kind of situation.

scottgonzalez:
I can't reproduce the behavior you described on Stack Overflow using http://jsfiddle.net/kqobjatv/1/. I'm going to close this for now since it's so old and we can't reproduce. Can you please file a ticket at http://bugs.jqueryui.com/newticket with another reduced test case and instructions? Thanks.

Tomas-M:
Still reproducible.
See the image, which describes how it SHOULD behave, and how it behaves incorrectly instead
rnxp8

Seems like my patch is not included in latest stable jquery ui 1.11.4

To fix the error, you just need to remove this line in jquery ui:
    axis = floating ? "clientX" : "clientY";
and replace it by this line:
    axis = floating ? "pageX" : "pageY";

It's really such simple. If you do not like to fix it, screw you. Sorry, I feel like idiot, I actually spend time and also real money to actually FIX this for you, for free, and you don't accept the fix

scottgonzalez:
I'm not sure why you're so aggressive. I reviewed the patch and it looked reasonable. So I pulled it down to test. I ran it against some existing test cases for scrolling bugs and those weren't fixed with the patch. I then followed along through Stack Overflow to your fiddle and couldn't reproduce the buggy behavior inside jsFiddle.

Are you experiencing this bug in only certain browsers? Without a reproducible test case, we can't ensure that anything is fixed. A screenshot isn't sufficient.

Tomas-M:
I'm not aggressive I am sad.
Sorry about that.

I am testing your URL http://jsfiddle.net/kqobjatv/1/ in Chrome, FireFox, Internet Explorer.
The bug is there for all browsers. Still reproducible.

The problem is that only chrome will scrollTop() the page automatically (I do not know why other browsers won't).

The page with table rows must be scrolled for the bug to appear. Then it is reproducible in all browsers. Hope this helps.

Tomas-M:
Only after the page is scrolled, the bug takes effect. It is because the code in jquery ui wants to use clientX/Y while it should use pageX/Y ... 
image

arschmitz:
@Tomas-M i reproduced this there is a key step in reproducing you missed which is this only happens the first time the connected sortable updates so you need to drag slowly as you do it or you can miss it completely and it wont happen on subsequent moves.

I looked at this as @scottgonzalez was trying to reproduce and did not see it initially either now that we can reproduce we can confirm this fix

Thank you for your quick reply

arschmitz:
Whats actually happening here is that the sortable sizes are changing do to the initial update which causes the first one to grow and the second to shrink. The offset in the intended position is this difference it does not happen on subsequent updates because the size is not changed then.

arschmitz:
This was actually an exact duplicate of a PR submitted almost 5 months previous so I landed that PR and closed this one

PR #1586, opened 2015-08-12
Form consistency

arschmitz:
This adds padding to spinner inputs to make them match the size of buttons and selectmenus.
It also fixes some widgets not using the default border color

http://view.jqueryui.com/master/demos/spinner/default.html vs 
http://view.jqueryui.com/form-consistency/demos/spinner/default.html 

Fixes http://bugs.jqueryui.com/ticket/13997

### DIFF
--- a/themes/base/spinner.css
+++ b/themes/base/spinner.css
@@ -19,7 +19,7 @@
 	border: none;
 	background: none;
 	color: inherit;
-	padding: 0;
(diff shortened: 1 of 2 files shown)


jzaefferer:
Looks good to me, both visually and the diff.

fnagel:
+1

Tested on Win8.1 Chrome.

PR #1587, opened 2015-08-18
Object support without label and value property

vedatk:
And now, autocomplete custom object must have label and value property. This obligation is removed. Two options are added: itemLabel, itemValue. This options are used key for object property. If it is not configured, default values are "label" and "value".


scottgonzalez:
Sorry, but we're not going to increase the API surface for something that's so easily accomplished in user code. In the future, please read the [contributing guidelines](https://github.com/jquery/jquery-ui/blob/master/CONTRIBUTING.md) before sending a PR that changes the API.

PR #1588, opened 2015-08-21
Style updates based on latest version of style guide

arschmitz:
Updates to follow latest version of style guide in all files.

### DIFF
--- a/ui/effect.js
+++ b/ui/effect.js
@@ -13,7 +13,7 @@
 //>>docs: http://api.jqueryui.com/category/effects-core/
 //>>demos: http://jqueryui.com/effect/
 
-(function( factory ) {
+( function( factory ) {
 	if ( typeof define === "function" && define.amd ) {
 
 		// AMD. Register as an anonymous module.
@@ -23,7 +23,7 @@
 		// Browser globals
 		factory( jQuery );
 	}
-}(function( $ ) {
+}( function( $ ) {
 
 var dataSpace = "ui-effects-",
 	dataSpaceStyle = "ui-effects-style",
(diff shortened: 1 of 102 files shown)


scottgonzalez:
I looked through ~50% of this and it seems good to land.

ui/widget.js
```diff
@@ -246,7 +257,7 @@ $.widget.bridge = function( name, object ) {
 	};
 };
 
-$.Widget = function( /* options, element */ ) {};
+$.Widget = function( /* Options, element */ ) {};
```
jzaefferer: This one is kinda wrong. Won't hurt anyone, but still...

ui/effect.js
```diff
@@ -1316,8 +1332,8 @@ function standardAnimationOption( option ) {
 	return false;
 }
 
-$.fn.extend({
-	effect: function( /* effect, options, speed, callback */ ) {
+$.fn.extend( {
+	effect: function( /* Effect, options, speed, callback */ ) {
```
jzaefferer: "same" here, as below for $.Widget.

ui/widget.js
scottgonzalez: Hmmm... I wonder if we can get JSCS to catch that scenario.

@markelog

arschmitz: @scottgonzalez It DID catch this which is why it became uppercase. 

@markelog @mikesherov It seems like the `requireCapitalizedComments` is completely unenforceable in a realistic sense all of these valid comments fail with it 

```
// http://google.com

// #3456 this does xxx
// to yyy so that zzz.

function( /* option */ ) {

/*

test(  "i dont want this to run", function() { 

*/
```

There are probably more cases then just this, but because of these ones I was forced to disable this rule within jQuery UI ( though i believe i got most if not all the valid comments fixed before disabling )

scottgonzalez: I meant I wonder if we can get JSCS to catch the scenario of an inline comment inside an argument list so that it doesn't apply the capitalization. But the other examples you've posted do make this seems very hard to enforce.

arschmitz: @scottgonzalez just to be clear JSCS did not capitalize it I did, trying to get this passing. I meant to ( and did before landing ) revert them. JSCS only warns on first letter capitalization. The changes for this in the PR for comment capitalization were done with a search and replace. I was trying to get 100% passage with out having to have this rule exceptions from the jQuery preset. However after finding the examples above i decided this was not possible.

markelog: @arschmitz
>  ( and did before landing )

I wounder, did you aware that not add spasing in parentheses in English now? :-)

Yeah, `requireCapitalizedComments` requires some work - https://github.com/jscs-dev/node-jscs/issues/1686 could you add cases you find in there?

>  I was trying to get 100% passage with out having to have this rule exceptions from the jQuery preset

I guess if rule mess things up instead of helping, we probably should consider removing it from the preset until its working right

PR #1589, opened 2015-08-25
Jscs v2

arschmitz:
More style updates to please jscs moved overrides to .jscsrc

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -135,13 +135,9 @@ grunt.initConfig({
 		}
 	},
 
-	// Remove the requireSpacesInsideParentheses override once everything is fixed
 	jscs: {
 		all: {
 			options: {
-				requireCapitalizedComments: null,
-				fix: true,
-				maxErrors: 5000000,
 				config: ".jscsrc"
 			},
 			files: {
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -139,7 +139,10 @@ grunt.initConfig({
 	jscs: {
 		all: {
 			options: {
-				requireCapitalizedComments: null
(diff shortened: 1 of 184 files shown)


arschmitz:
@scottgonzalez fixed the commit message

Gruntfile.js
```diff
@@ -135,11 +135,10 @@ grunt.initConfig({
 		}
 	},
 
-	// Remove the requireSpacesInsideParentheses override once everything is fixed
 	jscs: {
 		all: {
 			options: {
-				requireCapitalizedComments: null
+				config: ".jscsrc"
```
jzaefferer: Is this actually needed? If there's a bug with maxErrors we should report it and override maxErrors, but keep letting it use its default config file finding.

arschmitz: I assume you meant to comment on the max errors line above but yeah its not a bug its just the default is 50.

jzaefferer: No, I meant this line. The default `config: true` should work just fine. In my own limited testing it seemed like `jscs` wouldn't report any errors unless I set `config` explicitly. Based on our conversation elsewhere it sounded like there's a bug where jscs wouldn't report any errors at all when there are more errors than maxErrors.

arschmitz: For the config file yes true should be fine. The bug you mentioned with max errors was already reported to them https://github.com/jscs-dev/node-jscs/issues/1722 it is unrelated to this as its only related to running with fix. I added it here just because the normal max is 50 and i believe we want to output them all.

jzaefferer:
Apart from the config property above, this looks good. It even fails when there's an error!

PR #1590, opened 2015-08-25
Calendar / Datepicker: WIP

fnagel:
Multiple fixes and improvements from wiki todo list.

* Fixes weekday name jumping when clicking the last day of the last
grid (numberOfMonths option)
* Fix view when moving between multiple months with keyboard
* Some basic tests for numberOfMonths option and the above mentioned issues
* Fixed some issues in icon trigger and min / max demo
* Use escapeSelector 
* Fix $.ui.calendarDate tests
* Make header prev / next buttons min / max option aware
* Minor bugfixes and CGL fixes


fnagel:
@jzaefferer Managed to fix quite a few bugs you mentioned centuries ago in some PR comments :-D

fnagel:
@rxaviers Could you please take a look at this failing test? Really not sure what to do here. Did the API change that much?

https://travis-ci.org/jquery/jquery-ui/builds/77296024#L245
https://github.com/fnagel/jquery-ui/blob/datepicker/tests/unit/date/core.js#L20

rxaviers:
@fnagel are you referring to the failing test due to `Date.prototype.setMonth()`? It has failed, because months in the Date object ranges from 0-11 (not 1-12) (ref [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)).

rxaviers:
Oh wait, it's not the native Date. But, the custom $.date UI uses... Checking...

rxaviers:
@fnagel the problem lies in https://github.com/fnagel/jquery-ui/blob/datepicker/external/date.js#L78. `getDay()` returns the day of the week. Instead, you want to use `getDate()`, to return the day of the month. They are indeed confusing method names. Blame JavaScript Date. :P

In your case, when you `.setMonth( 9 )`, it was taking `0` for the day of the week and using it as the day of the month, in which case you rewind one day and go to the last day of the previous month.

fnagel:
@rxaviers That seems to work, but I wonder why this was committed in the first place. 
Whatever, it's fixed and all tests are green! Thanks for your quick help!

tests/unit/date/helper.js
```diff
@@ -0,0 +1,30 @@
+define( [
+	"jquery",
+	"globalize",
+	"globalize/date",
+	"lib/helper"
+], function( $, Globalize, helper ) {
+
+return $.extend( helper, {
+	getAttributes: function( locale ) {
```
fnagel: @rxaviers Since we now need to pass attributes parameter I adopted what we do in calendar itself. Is this a good solution or is there a better way to do so?

tests/unit/calendar/calendar.html
```diff
@@ -6,7 +6,7 @@
 
 	<script src="../../../external/requirejs/require.js"></script>
 	<script src="../../lib/css.js" data-modules="core calendar"></script>
-	<script src="../../lib/bootstrap.js" data-widget="datepicker"></script>
```
fnagel: @scottgonzalez Is this fix correct or was "datepicker" intended?

tests/unit/date/helper.js
```diff
@@ -0,0 +1,30 @@
+define( [
+	"jquery",
+	"globalize",
+	"globalize/date",
+	"lib/helper"
```
rxaviers: Note "lib/helper" is inaccessible. Move "globalize/data" to the last item.

tests/unit/date/helper.js
rxaviers: I'm missing context I guess. Do you mean creating the globalize instance below and returning the object of formatters generated by it?

tests/unit/date/helper.js
fnagel: Works as expected, what do you mean by "inaccessible"? I adopted this from here: https://github.com/jquery/jquery-ui/blob/master/tests/unit/dialog/helper.js#L3

rxaviers: It shouldn't work.
 - $ = "jquery"
 - Globalize = "globalize"
 - helper = "globalize/date"
 - undefined = "lib/helper"

fnagel: I have no idea ;-D

@scottgonzalez Can you shed some light on this?

tests/unit/date/helper.js
fnagel: I want to know if this is the right place and way to generate the needed parameters for former $.date object tests. Helpful?

rxaviers: I'll defer this question to @jzaefferer. I don't have an opinion where to set this in the tests. I can say I'm ok though using Globalize this way to generate for needed formatters. :+1:

tests/unit/calendar/calendar.html
scottgonzalez: This is correct. The calendar test page should load the calendar test files :-)

tests/unit/date/helper.js
scottgonzalez: If the intent is to add a `getAttributes` method to the helper for these tests, then this is incorrect. @rxaviers' mapping in his most recent comment shows what's actually happening, so you're modifying `globalize/date` instead of `lib/helper`.

fnagel: Fixed.

demos/datepicker/icon-trigger.html
```diff
@@ -15,13 +15,18 @@
 				open: function() {
 					allowOpen = false;
 				}
-			});
+			} ),
+			widget = datepicker.datepicker( "widget" );
 
-		$( "<img src='images/calendar.gif' alt='Open Datepicker'>")
+		$( "<a href='#'>")
```
scottgonzalez: We should avoid fake links like this. Use a button, or just the image, instead.

demos/datepicker/icon-trigger.html
```diff
@@ -30,7 +35,7 @@
 <p>Date: <input type="text" id="datepicker"></p>
 
 <div class="demo-description">
-<p>Click the icon next to the input field to show the datepicker.  Set the datepicker to open on focus (default behavior), on icon click, or both.</p>
+<p>Click the icon next to the input field to show the datepicker. This demo will be removed soon. Please see our migration guide.</p>
```
scottgonzalez: Why are we removing this demo?

tests/unit/calendar/options.js
```diff
@@ -232,7 +235,65 @@ test( "min / max", function() {
...
+		container.find( "tbody:last td[id]:last" ).attr( "id" ),
+		"calendar-2015-9-31",
+		"Correct id set for last day of third grid"
+	);
+
+	// Test for jumping in weekday rendering after click on last day of last grid
+	container.find( "tbody:last td[id]:last button" ).trigger( "mousedown" );
+	equal( container.find( "thead:last th:last" ).text(), "Sa",
+		"After mousedown last month: Last day is Saturday"
+	);
```
scottgonzalez: I'm having troubling figuring out what this is testing. Why would clicking any date change the headings?

tests/unit/calendar/options.js
```diff
@@ -232,7 +235,65 @@ test( "min / max", function() {
...
+
+	// Test if using cursor down to go to the next month advances three month
+	container.find( "tbody:first td[id]:first button" ).trigger( "mousedown" );
+	$( document.activeElement ).simulate( "keydown", { keyCode: $.ui.keyCode.LEFT } );
+	equal( container.find( ".ui-calendar-month:first" ).text(), "May",
+		"After move to previous month: First month is May"
+	);
+});
```
scottgonzalez: Let's add another assertion for moving in the other direction.

tests/unit/calendar/options.js
```diff
@@ -232,7 +235,65 @@ test( "min / max", function() {
...
+
+	element
+		.calendar( "option", { max: null } )
+		.calendar( "value", "1/4/09" )
+		.calendar( "option", { min: minDate, max: maxDate } );
+	ok( nextButton.hasClass( "ui-state-disabled" ), "Other year: Next button disabled" );
+	prevButton.simulate( "click" );
+	ok( nextButton.hasClass( "ui-state-disabled" ), "Other year: Next button disabled after click" );
+	prevButton.simulate( "click" );
+	ok( !nextButton.hasClass( "ui-state-disabled" ), "Other year: Next button enabled after click" );
```
scottgonzalez: Let's add another test for having a date below the min.

ui/widgets/calendar.js
```diff
@@ -150,7 +151,14 @@ return $.widget( "ui.calendar", {
 		}
 
 		if ( this._needsRefresh() ) {
-			this._refresh();
+			if ( this.options.numberOfMonths > 1 && this.date.year() === this.viewDate.year()) {
```
scottgonzalez: Missing space before the last closing paren.

ui/widgets/calendar.js
```diff
@@ -225,6 +233,10 @@ return $.widget( "ui.calendar", {
 			} )
 			.html( pickerHtml );
 
+		this.prevButton = $( ".ui-calendar-prev", this.element );
```
scottgonzalez: `this.element.find( ".ui-calendar-prev" )`

demos/datepicker/icon-trigger.html
fnagel: Jörn noticed this should be usable with keyboard only, thats why I added a link. Switching to a button seems like a good solution.

demos/datepicker/icon-trigger.html
fnagel: From the wiki

> keep this demo until we work on the migration guide for the new API, where we include this example, then delete the demo as it's promoting bad practice

tests/unit/calendar/options.js
fnagel: We had an issue with changing header text (day names) when using multiple month demo. This was caused by invalid usage of the internal date variables. This test makes sure this won't happen again.

ui/widgets/calendar.js
fnagel: Fixed

ui/widgets/calendar.js
fnagel: Fixed.

tests/unit/calendar/options.js
fnagel: Added

tests/unit/calendar/options.js
fnagel: Added

demos/datepicker/icon-trigger.html
```diff
@@ -15,13 +15,17 @@
 				open: function() {
 					allowOpen = false;
 				}
-			});
+			} ),
+			widget = datepicker.datepicker( "widget" );
 
-		$( "<img src='images/calendar.gif' alt='Open Datepicker'>")
+		$( "<button>")
```
jzaefferer: Missing space before closing parent. I'm going to ignore other style issues, until we've got tooling to find these. 

@arschmitz what happened to your project to lint HTML embedded JS and CSS?

demos/datepicker/icon-trigger.html
jzaefferer: Let's add "since its promoting bad practice" to the description.

tests/unit/calendar/options.js
jzaefferer: Can you make that more explicit by checking the month instead?

tests/unit/calendar/options.js
```diff
@@ -232,7 +235,81 @@ test( "min / max", function() {
...
+
+	// Test for jumping in weekday rendering after click on last day of last grid
+	container.find( "tbody:last td[id]:last button" ).trigger( "mousedown" );
+	equal( container.find( "thead:last th:last" ).text(), "Sa",
+		"After mousedown last month: Last day is Saturday"
+	);
+
+	// Test if using cursor to go to the next / prev month advances three month
+	container.find( "tbody:first td[id]:first button" ).trigger( "mousedown" );
+	$( document.activeElement ).simulate( "keydown", { keyCode: $.ui.keyCode.LEFT } );
```
jzaefferer: Why do you have to trigger two events? Also applies to the next group.

ui/widgets/calendar.js
```diff
@@ -150,7 +151,14 @@ return $.widget( "ui.calendar", {
...
 
 		if ( this._needsRefresh() ) {
-			this._refresh();
+			if ( this.options.numberOfMonths > 1 && this.date.year() === this.viewDate.year() ) {
+				this.viewDate.adjust( "M", this.options.numberOfMonths *
+					( this.date.month() > this.viewDate.month() ? 1 : -1 )
+				);
+				this.refresh();
+			} else {
+				this._refresh();
```
jzaefferer: Based on usage here and method naming, I can't tell the difference between `_refresh()` and `refresh()`. Maybe rename `_refresh()` and make callers call both that and `refresh()`?

Btw. the multi line comment in front of `refresh()` doesn't help either, it mostly sounds like a "avoid calling this" warning.

tests/unit/date/core.js
```diff
@@ -1,19 +1,21 @@
...
 
 module( "date: core" );
 
+var attributes = testHelper.getAttributes( "en" );
+
 test( "Instantiation", function() {
 	expect( 2 );
-	ok( new $.date() instanceof $.date, "constructor function" );
-	ok( $.date() instanceof $.date, "instantiation without new" );
+	ok( new $.ui.calendarDate( null, attributes ) instanceof $.ui.calendarDate, "constructor function" );
```
jzaefferer: Let's move `external/date.js` to `ui/calendar-date.js`? Update the folder for this file to "calendar-date" as well.

tests/unit/date/core.js
```diff
@@ -1,19 +1,21 @@
...
 	"jquery",
-	"globalize",
+	"./helper",
 	"date"
-], function( $, Globalize ) {
+], function( $, testHelper ) {
 
 module( "date: core" );
 
+var attributes = testHelper.getAttributes( "en" );
```
jzaefferer: Rename to enAttributes?

tests/unit/date/helper.js
```diff
@@ -0,0 +1,30 @@
+define( [
+	"jquery",
+	"globalize",
+	"lib/helper",
+	"globalize/date"
+], function( $, Globalize, helper ) {
+
+return $.extend( helper, {
+	getAttributes: function( locale ) {
```
jzaefferer: `external/date.js` calls `attributes.parse( date );`, yet this doesn't export a `parse` function, seems fishy. Missing code coverage?

jzaefferer:
There's this comment on calendar.js:

```
// TODO: Keep button even if its optional?
```
Since AMD/requirejs has no concept of optional modules, yes, keep it. Custom requirejs builds can still exclude modules as needed.

tests/unit/date/helper.js
jzaefferer: `ui/widgets/calendar.js` does provide the `parse` property, so this seems just missing in this helper (along with missing tests).

jzaefferer: Or `$.ui.calendarDate` should reject a string as the first argument, only accepting a `Date`.

external/date.js
```diff
@@ -228,7 +228,7 @@ $.extend( _Date.prototype, {
 	equal: function( other ) {
```
jzaefferer: The TODO comment above can be removed.

jzaefferer: Also, is this method still used anywhere? I don't see any calls in `ui/widgets/calendar.js`.

jzaefferer:
Done! Feel free to add non-regression tasks to the list, as usual...

tests/unit/date/core.js
fnagel: We currently do not use any other attributes object. Where needed, we change Globalize locale directly.

tests/unit/date/core.js
fnagel: We've already changed this once, so I would like to get some more opinion on this issue.

@scottgonzalez @arschmitz @rxaviers

scottgonzalez: I still don't really understand the scope of this object. Is it useful on its own? Is it intended to be used by others?

One thing that's clear to me though is that it cannot simultaneously be in `/external/` and use the `$.ui` namespace.

external/date.js
fnagel: Right, seems unused within calendar and datepicker. Perhaps we should reconsider this when we decide what happens with this file at all.

@rxaviers What's your opinion on that?

tests/unit/date/core.js
fnagel: > Is it useful on its own? Is it intended to be used by others?

Yes. One could use it to ease up working with JS date objects without using calendar or datepicker at all. Pretty sure there are more fancy date helpers out there, but...

demos/datepicker/icon-trigger.html
fnagel: > Let's add "since its promoting bad practice" to the description.

Done.

demos/datepicker/icon-trigger.html
fnagel: > Missing space before closing parent.

Fixed.

external/date.js
fnagel: > The TODO comment above can be removed.

Done.

tests/unit/calendar/options.js
fnagel: No, that would make this test useless. It's about the changing day names.

tests/unit/calendar/options.js
fnagel: We need to focus the first / last day of a month, then using arrow keys to jump to prev / next month. Could probably use plain focus on that first / last day.

tests/unit/date/helper.js
fnagel: Yeah, using attributes within $.date has been changed "lately" when Globalize 1.0 has been added, but without extending the tests. Added to the wiki todo list.

ui/widgets/calendar.js
fnagel: >  I can't tell the difference between _refresh() and refresh()

One sets this.viewDate, the other does not. Could be removed by adding ````this.viewDate.setTime( this.date.date().getTime() );```` multiple times. The _refresh had more code in it before but was stripped down over the time, perhaps its time to remove it.

This multi line comment was added loooong time ago. To me it sounds like a "we need to check if the basic idea works at all" comment as refresh is used a lot (basically on every interaction). Anyway, some (not all) of the concerns have been fixed (like changing text instead of regenerate the buttons) lately. I would say we keep this until we fixed some more known issues and tested this for a11y.

fnagel:
> // TODO: Keep button even if its optional?
> > Since AMD/requirejs has no concept of optional modules, yes, keep it. Custom requirejs builds can > > still exclude modules as needed.

Removed.

external/date.js
rxaviers: My opinion about the `external/date.js` in general?

If it's an external library, it should live in its own repo. If it's an UI internal thing, it should live somewhere in `ui/`.

Just let me know if you were asking about anything else. :)

tests/unit/calendar/options.js
jzaefferer: Let's do that, makes more sense.

ui/widgets/calendar.js
jzaefferer: Can we rename `_refresh` and drop the `refresh()` call from it? Then any caller that currently calls `_refresh()` will have to call both methods. That seems like a good tradeoff.

tests/unit/date/core.js
jzaefferer: I still think we should move it to `ui/`. The other "utilities" in there are also of limited use, like `ui/focusable.js`, which is only used by dialog.

external/date.js
jzaefferer: Again, let's move to `ui/`. And drop anything not in use anywhere.

tests/unit/calendar/options.js
fnagel: Not sure why but that won' work as expected.

ui/widgets/calendar.js
fnagel: Mhh, not sure if that makes sense. Do we want a method (perhaps "_updateView") with only one line of code in it? 
Perhaps we should remove it completely (already thought about this) and just add the additional line four times. Or just rename the method.

jzaefferer: Let's remove it for now, that might help come up with a better abstraction later.

tests/unit/calendar/options.js
jzaefferer: Okay. Can you add a comment to capture that?

external/date.js
fnagel: Moved to ui namespace and removed all unused methods.

fnagel: ps: Removed all unused methods but not "equal()" as its used within the date object itself.

tests/unit/calendar/options.js
fnagel: Comment added.

fnagel:
@jzaefferer @scottgonzalez Fixed all open comments.

jzaefferer:
Travis build failed with a jscs error. That should be trivial to fix.

jzaefferer:
Other than the style issues, this looks good to land. As usual, a rebase on master will be interesting, since we've got more style check changes there.

scottgonzalez:
We've reached the point of no more rebases for datepicker, so it'll just be a merge.

jzaefferer:
Sure, as long as we get the jscs changes into the datepicker branch.

fnagel:
Fixed jscs error.

jzaefferer:
I don't see any new commits. Did you merge master?

fnagel:
No, just rebase changed this commit to fix the jscs error: d650fa9a7e127eb6950f93034c92accb65bcb2c3

fnagel:
@jzaefferer I've just pushed again. Merged with master and jscs errors fixed. Please take a look!

jzaefferer:
Okay, let's land this.

PR #1591, opened 2015-08-27
Added bower.json

andremleblanc:


scottgonzalez:
Why would you want to add this to 1-10-stable?

andremleblanc:
We're using an older library that utilizes 1-10-stable and I'd like to load it through bower into my Rails app using Rails-Assets.org.

scottgonzalez:
Why aren't you using https://github.com/components/jqueryui? `bower install jquery-ui#1.10.4` already works just fine.

andremleblanc:
Oops! My trouble ended up being a gem conflict, thanks for your help.

PR #1592, opened 2015-09-03
Component: ui.slider - setting range from true to false should result in one handle instead of two

blingerson:
Switching range from true to false should remove second handle.

http://jsfiddle.net/1639sgx8/

As demonstrated in the fiddle when one programmatically changes the range option from true to false the second handle remains. This fix simply takes the min value as the new single handle value.

### DIFF
--- a/tests/unit/slider/options.js
+++ b/tests/unit/slider/options.js
@@ -374,12 +374,12 @@ test( "range", function( assert ) {
 		max: 10,
 		step: 1
 	}).slider( "option", "range", false );
-	equal( element.find( ".ui-slider-handle" ).length, 2, "range switch from true to false, both handles remain" );
+	equal( element.find( ".ui-slider-handle" ).length, 1, "range switch from true to false, only one handle should remain" );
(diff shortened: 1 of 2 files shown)


scottgonzalez:
It shouldn't, actually. As soon as the range exists, the slider becomes multi-valued. Removing the range can't automatically change a two value slider to a one value slider.

blingerson:
Is there a workaround for this? I noticed this behavior for min and max (it sets the slider to single value going from true to "min" or "max". In my case I have a situation where I am using the slider with an operator chooser... ie =, <>, <, >, <=, >= (single values) between and outside (multi value range). Everything works besides switching range to = <> for example. In this case I want one handle and no coloring for min max range. Thanks.

scottgonzalez:
Set the `values` option yourself. The number of handles **must** change for `min` and `max` since those can only work with one handle.

blingerson:
Thanks Scott, right before you responded, I tried changing values with $(element).slider( 'option', 'values', null ) and that did the trick.

PR #1593, opened 2015-09-09
Untested concept of adding image width and height to add possibility …

ywarnier:
…to size icon when using SVG.

When using an SVG icon (now supported by most browsers, including mobile browsers), jquery-ui should allow datepicker to receive width and height attributes to define the size of the icon display. Otherwise, the SVG icon might appear huge (like this).
datepicker-svg

This PR hasn't been tested because I'm not familiar with the jquery-ui building process, but I prefer contributing something partial than not contributing at all. Hopefully someone will be able to test it properly and integrate it?

SVG clock for testing can be found here: https://unstable.chamilo.org/main/img/icons/svg/attendance.svg

### DIFF
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -107,6 +107,8 @@ function Datepicker() {
 		appendText: "", // Display text following the input box, e.g. showing the format
 		buttonText: "...", // Text for trigger button
 		buttonImage: "", // URL for trigger button image
+		buttonImageWidth: "", // in case the image is SVG, we need a defined width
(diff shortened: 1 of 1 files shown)


ywarnier:
Now I tested my changes locally (directly modifying a jquery-ui.js file) and it worked fine.

scottgonzalez:
Thanks, but we're not accepting any changes to the API. Datepicker is in the middle of a complete rewrite, so any changes here will be removed in the next version anyway. You can already accomplish this same thing on your own by generating the images yourself.

PR #1594, opened 2015-09-09
Update datepicker height dynamically when users change month or year

LucasZheng:
## Issue
Correct ui:
correct_ui
Incorrect ui:
wrong_ui
## Description
Users change the month in the Datepicker panel, when there are six weeks in a panel, Datepicker will hidden below input element.


scottgonzalez:
I haven't tested the patch, but just based on the description, this sounds like the datepicker is going to jump on month change which would be pretty bad UX. If you're interested in discussing this further, please start with a bug report and described in the [contributing guidelines](https://github.com/jquery/jquery-ui/blob/master/CONTRIBUTING.md). Thanks.

PR #1595, opened 2015-09-11
Fixed date format to correct format. Plays well with jquery globalize.

jspeaker:
The date format has been changed to the correct format per wikipedia. This date format integrates correctly with the jquery globalize plugin, which is the problem I needed to resolve for integration.
https://en.wikipedia.org/wiki/Date_and_time_notation_in_South_Korea

### DIFF
--- a/ui/i18n/datepicker-ko.js
+++ b/ui/i18n/datepicker-ko.js
@@ -25,7 +25,7 @@ datepicker.regional.ko = {
 	dayNamesShort: [ "일","월","화","수","목","금","토" ],
 	dayNamesMin: [ "일","월","화","수","목","금","토" ],
 	weekHeader: "주",
-	dateFormat: "yy-mm-dd",
+	dateFormat: "yy. mm. dd.",
 	firstDay: 0,
 	isRTL: false,
 	showMonthAfterYear: true,


ui/i18n/datepicker-ko.js
```diff
@@ -25,7 +25,7 @@ datepicker.regional.ko = {
 	dayNamesShort: [ "일","월","화","수","목","금","토" ],
 	dayNamesMin: [ "일","월","화","수","목","금","토" ],
 	weekHeader: "주",
-	dateFormat: "yy-mm-dd",
+	dateFormat: "yy. mm. dd.",
```
scottgonzalez: The month and day should be single digit when possible. I'm not sure which form of the year we would use here. @rxaviers, do we display the medium or short date from CLDR?

rxaviers: Hey @scottgonzalez,

Datepicker rewrite using Globalize 1.x uses `{date: 'short'}` https://github.com/jquery/jquery-ui/blob/datepicker-globalize-1.x/ui/calendar.js#L42.

For comparison:

```js
Globalize('en').formatDate(new Date(), {date: 'short'})
// > '9/21/15'

Globalize('en').formatDate(new Date(), {date: 'medium'})
// > 'Sep 21, 2015'

Globalize('en').formatDate(new Date(), {date: 'long'})
// > 'September 21, 2015'

Globalize('en').formatDate(new Date(), {date: 'full'})
// > 'Monday, September 21, 2015'

Globalize('en').formatDate(new Date(), {skeleton: 'yMd'})
// > '9/21/2015'

Globalize('en').formatDate(new Date(2015,0,1), {skeleton: 'yMd'})
// > '1/1/2015'

Globalize('en').formatDate(new Date(2015,0,1), {skeleton: 'yMMdd'})
// > Should give you the above using padded months and days when the PR below is complete:
// https://github.com/jquery/globalize/pull/462
```

fnagel: Please note (from the wiki): "Compared to master, the default date format is different, is that intended? Master has 4 digit years, rewrite only 2"

PR #1596, opened 2015-09-15
Autocomplete for Chrome Fixed

exussum12:
Chrome since Version 34 has not followed the standard "off" setting
setting to false seems to work on other browsers

### DIFF
--- a/tests/unit/autocomplete/options.js
+++ b/tests/unit/autocomplete/options.js
@@ -316,4 +316,11 @@ test( "source, update after init", function() {
 	equal( menu.find( ".ui-menu-item" ).text(), "php" );
 } );
 
+test( "Autocomplete attribute", function() {
+	expect( 1 );
+	var element = $( "#autocomplete" ).autocomplete();
+	equal( element.attr( "autocomplete" ), "false",
+		"Should be false, Off prevents chrome working" );
+} );
+
 } );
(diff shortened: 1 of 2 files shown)


exussum12:
Not sure how to run the bot again, Fixed the error

scottgonzalez:
It's working just fine for me. Do you have a link to a bug report or anything else to back up your statement?

exussum12:
The chrome issue
https://code.google.com/p/chromium/issues/detail?id=468153

Second answer here
http://stackoverflow.com/questions/15738259/disabling-chrome-autofill
(78 upvotes currently)

No jQueryUI bug currently, but I can create one ?

Technically it should be a chrome only thing, But as false seems to work in other places, it doesnt seem harmful to change

scottgonzalez:
I'm skeptical of this fix. Looking at http://jsfiddle.net/4wq5cozo/, which is a test case from https://code.google.com/p/chromium/issues/detail?id=527408 (closed as a duplicate of the ticket you linked to), changing `firstname2` to have `autocomplete="false"` disables all autofill on that page, not just for that field. That doesn't seem like a viable solution in jQuery UI.

exussum12:
Hmm I see that, What about having an option passed on autocomplete to allow this ? Have the default as "off" but an option to set as false ?

scottgonzalez:
Setting an option would be just as easy as doing `.attr("autocomplete", "false");`, so I don't see how that would help.

scottgonzalez:
I'm going to close this PR since I don't think there's anything we can do. However, I am talking to Chrome developers to see if we can get some updates in the browser. I'll post updates here.

PR #1598, opened 2015-09-16
All: Delegate to base `_getCreateOptions()`.

scottgonzalez:
Ensures that any extensions to the base widget will be handled properly by
individual widgets.

### DIFF
--- a/tests/unit/widget/core.js
+++ b/tests/unit/widget/core.js
@@ -267,7 +267,7 @@ test( "merge multiple option arguments", function() {
 } );
 
 test( "._getCreateOptions()", function() {
-	expect( 3 );
+	expect( 4 );
 	$.widget( "ui.testWidget", {
 		options: {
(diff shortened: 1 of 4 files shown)


jzaefferer:
Can we extend the common widget tests to make sure all `_getCreateOptions` overrides call `_super`?

Otherwise looks good.

scottgonzalez:
That's tricky because we would need to create an instance of the widget to guarantee that the properties are set to reasonable values for the individual widget's implementation of `_getCreateOptions()`. Creating an instance requires knowing what markup to use, and widgets like checkboxradio don't have `defaultElement` set.

We can check for `this._super()` in the `.toString()` of `._getCreateOptions()`, but it seems a little ghetto.

jzaefferer:
I can't think of a non-ghetto approach to deal with this, so lets leave it as it is.

PR #1599, opened 2015-09-17
Widget: Call `._setOptionDisabled()` on init if the widget is disabled

scottgonzalez:
Fixes #9151

The full test suite passes, but I haven't gone through demos to make sure everything actually looks and behaves properly. We should probably also go through the `._create()` method for each widget and remove any logic that isn't necessary as a result of this change.

### DIFF
--- a/tests/unit/widget/core.js
+++ b/tests/unit/widget/core.js
@@ -729,6 +729,39 @@ test( ".disable()", function() {
 	$( "<div>" ).testWidget().testWidget( "disable" );
 } );
 
+test( "._setOptionDisabled()", function() {
+	expect( 3 );
+
+	var method;
+	var widget;
+
+	$.widget( "ui.testWidget", {
+		_setOptionDisabled: function( value ) {
+			method( value );
+		}
+	} );
+
+	method = function() {
+		ok( false, "._setOptionDisabled() called on init when not disabled" );
+	};
+	$( "<div>" ).testWidget();
+
+	method = function( value ) {
+		strictEqual( value, true, "._setOptionDisabled called on init when disabled" );
+	};
+	widget = $( "<div>" ).testWidget( { disabled: true } );
+
+	method = function( value ) {
+		strictEqual( value, false, "._setOptionDisabled called when enabling" );
+	};
+	widget.testWidget( "enable" );
+
+	method = function( value ) {
+		strictEqual( value, true, "._setOptionDisabled called when disabling" );
+	};
+	widget.testWidget( "option", "disabled", true );
+} );
+
 test( ".widget() - base", function() {
 	expect( 2 );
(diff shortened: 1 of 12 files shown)


jzaefferer:
Looking through uses of `disabled` options in widgets and testing with `{disabled: true}` on create:

* Accordion doesn't respect the `disabled` option on create. Patched in be67a39
* Autocomplete works fine
* Skipping button and datepicker, see rewrite
* Dialog can't be disabled, need to ensure that on create, patched in e58ed3c
* Draggable's adding of the disabled class on create is now redundant, patched in 337b307
* Droppable works fine
* Menu's handling of the disabled option on create is now redundant, patched in eb5374a
* Mouse doesn't deal with disabled at all
* Progressbar, like accordion, doesn't respect the `disabled` option on create. Patched in 92adcb2
* Resizable and Selectable work fine
* Selectmenu's handling of the disabled option on create is now redundant, patched in 0296e9c
* Sliders's handling of the disabled option on create is now redundant, patched in cab1363
* Sortable works fine
* Spinners's handling of the disabled option on create is now redundant, patched in c2a9aa6
* Tabs is not working correctly, for example it ends up with the `ui-tabs-disabled` class when only setting specific tabs, like `disabled: [1, 2]`. Due to the array variant I haven't yet figured out what to do here.
* Tooltip has special behaviour for `disabled`, it already disable regular style changes to the element, 68273a4 applies the same for create.

None of the interactions have a visual disabled state, that's on purpose, right?

scottgonzalez:
> None of the interactions have a visual disabled state, that's on purpose, right?

Correct. Interactions don't get visually disabled states because they represent actions, not controls. Applying a disabled state to something like draggable would make the contents of the draggable seem disabled, which isn't the case.

scottgonzalez:
All of the commits that Jörn referenced are from https://github.com/scottgonzalez/jquery-ui/pull/1, if anyone else wants to take a look. I've already reviewed it, and once everything is updated over there, I'll land all of this together.

scottgonzalez:
@jzaefferer's commits have been merged into this PR.

jzaefferer:
I actually failed to verify passing tests (and your repo doesn't have Travis enabled...), so I missed on failing assertion. This fixes it, and seems like the right approach to me: https://gist.github.com/jzaefferer/3c06668c1ed37129edd9

That is, if all individual tabs are disabled, it should have the `ui-tabs-disabled` class, just like it would have when disabling the whole widget.

zhump:
How to make the jQuery UI Accordion widget multiple?

scottgonzalez:
@zhump This is not an appropriate place to ask support questions. Please use the forums, Stack Overflow, or IRC.

zhump:
ok sorry thanks

PR #1601, opened 2015-09-18
Changed to allow new Globalize functions

sasimon:
There are some problems with your Spinner widget that prevent them from working with the latest version of Globalizer, a major release (0.1 -> 1.0). I have identified and fixed some of them.


scottgonzalez:
We're not switching to Globalize 1.x yet.

scottgonzalez:
To expand on that, we'll be updating to Globalize 1.x in 1.13.0. The work has already started in the datepicker branch. Spinner will actually be changed to include options for how to parse and format value. See [the roadmap](http://wiki.jqueryui.com/Roadmap) for more information about the Globalize update.

PR #1602, opened 2015-09-18
Autocomplete: Fix HTML markup

monojp:
There is an accessibility tag appended before the closing body tag that contain divs inside a span which is invalid HTML markup and the W3C Validator reports it as 'Element “div” not allowed as child of element “span” in this context.'. This fixes the markup by using a div instead of a span.

Fixes #14587


scottgonzalez:
Thanks. Unfortunately as soon as I saw the bug report, I pushed a fix. Sorry for not seeing this PR first.

PR #1603, opened 2015-09-21
Demos: Deal with missing disable-selection dependency

jzaefferer:
Noticed this while looking at demos for gh-1599.


scottgonzalez:
Looks good.

PR #1605, opened 2015-09-25
Focusable: Fix handling of `visibility: inherit`

scottgonzalez:
Ref #14596
Ref gh-1583

### DIFF
--- a/ui/focusable.js
+++ b/ui/focusable.js
@@ -42,8 +42,17 @@ $.ui.focusable = function( element, hasTabindex ) {
 		"a" === nodeName ?
 			element.href || hasTabindex :
 			hasTabindex ) &&
-		$( element ).is( ":visible" ) && $( element ).css( "visibility" ) === "visible";
+		$( element ).is( ":visible" ) && visible( $( element ) );
 };
+
+function visible( element ) {
+	var visibility = element.css( "visibility" );
+	while ( visibility === "inherit" ) {
+		element = element.parent();
+		visibility = element.css( "visibility" );
+	}
+	return visibility !== "hidden";
+}
 
 $.extend( $.expr[ ":" ], {
 	focusable: function( element ) {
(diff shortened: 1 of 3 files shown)


jzaefferer:
Tests?

jzaefferer:
Nevermind, saw your comment about IE8. Looks good to me.

scottgonzalez:
I can add a test that explicitly assigned `visibility: inherit` which would cause all browsers to hit this code path.

jzaefferer:
That seems useful.

ui/focusable.js
```diff
@@ -42,9 +42,18 @@ $.ui.focusable = function( element, hasTabindex ) {
 		"a" === nodeName ?
 			element.href || hasTabindex :
 			hasTabindex ) &&
-		$( element ).is( ":visible" ) && $( element ).css( "visibility" ) === "visible";
+		$( element ).is( ":visible" ) && visible( $( element ) );
 };
 
+function visible( element ) {
+	var visibility = element.css( "visibility" );
+	while ( visibility === "inherit" ) {
```
westonruter: Should this also check for `element.length !== 0` to avoid a possible infinite loop?

scottgonzalez: There is no possible infinite loop here. If the element doesn't exist, it can't have a `visibility` of `"inherit"`.

westonruter: Of course. :+1:

scottgonzalez:
Tests added. It turns out that even with these tests, only IE 8 will hit the new code path. All other browsers resolve `inherit` to either `hidden` or `visible` when getting the computed value (which makes sense). As such, I've marked the `visible()` method as supporting IE 8, but not the tests. The test cases are still valid, and if the implementation were to ever change to checking `element.style` instead of the computed style, the tests would catch the failure (unless the tree walking logic still existed).

jzaefferer:
Looks good. Somewhat related: https://github.com/medialize/ally.js/issues/41

PR #1606, opened 2015-09-28
Menu: Check that there is an active item in _activate function

RandScullard:
If there was no active item, _activate was throwing an exception.

### DIFF
--- a/ui/widgets/menu.js
+++ b/ui/widgets/menu.js
@@ -270,7 +270,7 @@ return $.widget( "ui.menu", {
 	},
 
 	_activate: function( event ) {
-		if ( !this.active.is( ".ui-state-disabled" ) ) {
+		if ( this.active && !this.active.is( ".ui-state-disabled" ) ) {
 			if ( this.active.children( "[aria-haspopup='true']" ).length ) {
 				this.expand( event );
 			} else {


jzaefferer:
Can you provide a test case where this is failing?

RandScullard:
Is that really necessary for such a trivial fix? If so, I'll create one, but I was hoping I could skip that step to save time.

jzaefferer:
Yeah, it is. Even trivial changes can lead to regressions. We need to at least know how this occurs.

RandScullard:
Here you go:
https://jsbin.com/cugujaxiha/edit?html,css,js,output

To reproduce the error, click in the scroll bar of the menu, then hit the space bar (you might have to repeat these steps a few times). You'll see in the console:

    TypeError: this.active is null jquery-ui.js:2381:7

(This was in Firefox 40.0.3)

jzaefferer:
Thanks, I can reproduce that. The error also happens without clicking around on the scrollbar, just hit space on a focused menu with overflow.

Based on that, I don't think the fix is correct. As long as space scrolls the menu, it should also activate whatever item scrolls into view, like it does when using the page up/down keys. Space should behave like page down.

What do you think? If that makes sense, could you look into implementing that, along with unit tests?

RandScullard:
I view it as two separate issues: 1. The menu widget should not throw an exception in this case and 2. the behavior of the menu widget could be enhanced to better handle the space key. The fix I provided fixes (1) with no negative impact on the behavior of the widget. For (2), I'm afraid I don't have the time to go that deep into this -- I'm pretty busy on a project right now.

jzaefferer:
A defensive check just to suppress the exception isn't very helpful, it still leaves the widget in an invalid state. Suppressing the space key should also requires only one or two lines of code, but would be a more thorough workaround.

RandScullard:
From my reading of the code, this is not an invalid state. It seems to be perfectly normal for there to be no active item -- this particular function just wasn't checking for that case. Meanwhile, there is code to explicitly handle the space key, so I'm not sure why you're suggesting that it be suppressed.

If you want to implement a more thorough change, please go ahead. I thought I would try to contribute a simple and effective fix that resolved an issue in our application, in the hope that it would help others. It is not my intention to tell you how to run your project, so please feel free to deal with this as you like. I don't have time for further involvement in this right now.

jzaefferer:
Okay, thanks for the contribution and the test case. Will get back to this later.

PR #1609, opened 2015-09-30
Tests: Use demo bootstrap for visual tests

jzaefferer:
Adds data-base-url and data-composite options and fixes a missing entry for effects. Otherwise just applies the same conversion as already applied to demos, sometimes moving a `<style>` element to the right place.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -147,15 +147,12 @@ grunt.initConfig({
 	},
 	uglify: minify,
 	htmllint: {
-		good: [ "tests/**/*.html" ].concat( htmllintBad.map( function( file ) {
-			return "!" + file;
-		} ) ),
-		demos: {
+		good: {
 			options: {
 				ignore: [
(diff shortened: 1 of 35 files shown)


jzaefferer:
Fixing build issue...

scottgonzalez:
Since the base URL only ever has two possible values, can we just check the path for the `script` and figure out the base inside the bootstrap code?

Also, some of these lines became very long, can we break the attributes onto multiple lines when they're long?

jzaefferer:
> Since the base URL only ever has two possible values, can we just check the path for the script and figure out the base inside the bootstrap code?

Done, replaced with a check for `window.location.pathname.indexOf( "demos/" ) !== -1`

> Also, some of these lines became very long, can we break the attributes onto multiple lines when they're long?

Without the `data-base-url` it got better. Also inserted some linebreaks on the remaining very-long lines. I also finally figured out why `data-modules="effects-all"` wasn't working. Now it works for both demos and visual tests.

scottgonzalez:
That's much better.

PR #1612, opened 2015-10-02
Fix: Autocomplete component invisible or visible on back

didijudo:
This class **ui-front** doesn't work on autocomplete module with modal box.
When a modal box are on focus, the autocomplete is invisible or is visible on back modal!

### DIFF
--- a/themes/base/core.css
+++ b/themes/base/core.css
@@ -54,7 +54,7 @@
 }
 
 .ui-front {
-	z-index: 100;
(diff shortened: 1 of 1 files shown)


scottgonzalez:
That's not a viable solution. Please provide a reduced test case showing your problem since we have a test that shows this works fine: http://view.jqueryui.com/master/tests/visual/compound/dialog_widgets.html

PR #1613, opened 2015-10-07
Tooltip: Adds the check to add timers only when both track and…

manimis902:
As explained in the raised ticket number.
This fix is to add timers only when both track and delay options are present.
The original issue of high CPU usage on mac/chrome when adding timers is a separate concern.

Fixes [#14682](http://bugs.jqueryui.com/ticket/14682)


jzaefferer:
Looks good to me. That check should've been there in the first place.

manimis902:
So should we merge this change ?

jzaefferer:
@scottgonzalez are we missing anything?

jzaefferer:
Landed with an updated commit message.

PR #1614, opened 2015-10-15
Autocomplete: Close the menu on any outside interactions

scottgonzalez:
This ensures that the menu will close if the user interacts with a
draggable, resizable, etc. element since those interactions don't
change focus.

Ref [#6642](http://bugs.jqueryui.com/ticket/6642)

-----

While testing this, I noticed that clicking on the scrollbar arrows (on an overflow menu) in IE doesn't work properly. Since we can't stop IE from moving focus (https://github.com/jquery/jquery-ui/blob/52d9ec6c4d7402d5ef04b8c7959f812f8d095cef/ui/widgets/autocomplete.js#L227-L234), when the user interacts with the scrollbar, the input blurs and then gains focus again. When the field receives focus, the menu jumps back up. As a result, clicking on the scrollbar arrows always results in the menu scrolling to the top. This happens while using the scroll box as well, but since that sets an explicit scroll amount, the menu ends up in the right location after jumping to the top.

We should try to fix that, but it's not related to this change.

### DIFF
--- a/tests/unit/autocomplete/core.js
+++ b/tests/unit/autocomplete/core.js
@@ -398,4 +398,28 @@ asyncTest( "Search if the user retypes the same value (#7434)", function() {
 	} );
 } );
 
+asyncTest( "Close on click outside when focus remains", function() {
+	expect( 2 );
+
+	var element = $( "#autocomplete" ).autocomplete( {
+		source: [ "java", "javascript" ],
+		delay: 0
+	} );
+	var menu = element.autocomplete( "widget" );
+
+	$( "body" ).on( "mousedown", function( event ) {
+		// event.preventDefault();
+	} );
+
+	element.val( "j" ).autocomplete( "search", "j" );
+	setTimeout(function() {
+		ok( menu.is( ":visible" ), "menu displays initially" );
+		$( "body" ).mousedown();
+		setTimeout(function() {
+			ok( menu.is( ":hidden" ), "menu closes after clicking elsewhere" );
+			start();
+		} );
+	} );
+} );
+
 } );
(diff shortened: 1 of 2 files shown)


jzaefferer:
Fix looks good. Let's file a separate ticket for the scrollbar issue. This seems easy enough to test (trigger mousedown inside and outside while menu is open).

scottgonzalez:
Test added.

tests/unit/autocomplete/core.js
```diff
@@ -398,4 +398,28 @@ asyncTest( "Search if the user retypes the same value (#7434)", function() {
...
+	expect( 2 );
+
+	var element = $( "#autocomplete" ).autocomplete( {
+		source: [ "java", "javascript" ],
+		delay: 0
+	} );
+	var menu = element.autocomplete( "widget" );
+
+	$( "body" ).on( "mousedown", function( event ) {
+		// event.preventDefault();
```
jzaefferer: What's this? Build failed due to unused `event` argument.

tests/unit/autocomplete/core.js
```diff
@@ -398,4 +398,28 @@ asyncTest( "Search if the user retypes the same value (#7434)", function() {
...
+	} );
+
+	element.val( "j" ).autocomplete( "search", "j" );
+	setTimeout(function() {
+		ok( menu.is( ":visible" ), "menu displays initially" );
+		$( "body" ).mousedown();
```
jzaefferer: We shouldn't use event shortcuts anymore, replace with `trigger( "mousedown" )`

tests/unit/autocomplete/core.js
scottgonzalez: Oops. That's from debugging. The call shouldn't be commented out.

tests/unit/autocomplete/core.js
scottgonzalez: Also from debugging when my browser kept caching the files. This is supposed to go through `.simulate()`.

PR #1616, opened 2015-10-15
Accordion: Correct height calculated when closed

scottgonzalez:
Fixes #11938
Closes gh-1536

This is just the code form gh-1536 squashed, rebased, and updated to the new style guide. It's passing locally for me, so I want to see if it will pass on Travis.


scottgonzalez:
Well that didn't work. I even changed Travis to use container based infrastructure and switched to node 0.12. I have no idea what else we can try to make Travis pass.

jzaefferer:
We need an evergreen headless browser that actually works. At this point we have to consider just not running any tests against PhantomJS.

PR #1618, opened 2015-10-17
Draggable: Fix options tests

jzaefferer:
These failed when running through tests/unit/all.html, due to the smaller
iframe. Reducing a browser window enough triggered the same issue. Making
the top offset much smaller fixes that.

The rest is fixing bad indent and adds a missing module call.

I ran tests through browserstack-runner locally, on these, everything passes:

```
"chrome_current",
"firefox_current",
"opera_current",
"safari_current",
"ie_8",
"ie_9",
"ie_10",
"ie_11",
"edge_latest"
```

### DIFF
--- a/tests/unit/draggable/draggable.html
+++ b/tests/unit/draggable/draggable.html
@@ -48,7 +48,7 @@
 	}
 	.sortable {
 		position: relative;
-		top: 8000px;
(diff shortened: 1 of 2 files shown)


scottgonzalez:
Looks good.

PR #1619, opened 2015-10-17
Tests: Fix button space assertions on IE8

jzaefferer:
IE8 still doesn't keep the space in the text (or HTML) representation. We don't really care, so adding more trims in these tests as well.

Fixes both of these: http://swarm.jquery.org/job/1478

Tested with BrowserStack on IE8.


scottgonzalez:
Looks good.

PR #1620, opened 2015-10-17
Mouse: Ignore `mousemove` events triggered by key presses in Safari

scottgonzalez:
If the user presses control, meta, shift, or alt during a drag operation,
Safari will trigger an event where `event.which` is `0`. We use that scenario
to detect that a `mouseup` occurred in a different document, so we need to
ignore these events when one of the keys are pressed.

Fixes #14461

### DIFF
--- a/ui/widgets/mouse.js
+++ b/ui/widgets/mouse.js
@@ -146,7 +146,16 @@ return $.widget( "ui.mouse", {
 
 			// Iframe mouseup check - mouseup occurred in another document
 			} else if ( !event.which ) {
-				return this._mouseUp( event );
+
+				// Support: Safari 8
+				// Safari sets which to 0 if you press any of the following keys
+				// during a drag (#14461)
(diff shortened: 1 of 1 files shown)


ui/widgets/mouse.js
```diff
@@ -146,7 +146,16 @@ return $.widget( "ui.mouse", {
 
 			// Iframe mouseup check - mouseup occurred in another document
 			} else if ( !event.which ) {
-				return this._mouseUp( event );
+
+				// Support: Safari 8
```
arschmitz: Support comment style also should make sure to check 9 since its out

jzaefferer:
Looks good.

arschmitz:
looks good to me other then on minor support comment issue

ui/widgets/mouse.js
scottgonzalez: I just tested Safari 9 and the bug exists there as well, so this will be updated to `// Support: Safari 8+` but I'm not sure what you think is wrong with this comment.

PR #1621, opened 2015-10-18
Mouse: Don't stop propagation of mouseup

hmethvin:
Fixes #10818


jzaefferer:
Looks good to me!

jzaefferer:
Thank you!

jzaefferer:
The new test is failing on IE8: http://swarm.jquery.org/result/740908 "Expected 1 assertions, but 0 were run"

scottgonzalez:
Heh. Welcome to 2009 where DOM Level 2 Events is the current spec and events don't bubble up to the window.

I'll send a PR to skip the test in IE 8 since the tested behavior actually shouldn't occur in that browser.

PR #1623, opened 2015-10-23
All: Fix manifest issues for categories and CSS dependencies

jzaefferer:
Collapses "UI Core" and "Core" into just "Core".

Fixes bad paths for CSS dependencies. Regressed when moving widgets into the widgets subfolder.

### DIFF
--- a/ui/core.js
+++ b/ui/core.js
@@ -9,7 +9,7 @@
  */
 
 //>>label: Core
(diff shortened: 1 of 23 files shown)


scottgonzalez:
Looks good.

arschmitz:
:+1:

PR #1624, opened 2015-10-25
Controlgroup: Fix create and destroy for controlgroupLabel

jzaefferer:
Creating the label broke while doing a refactoring, the lack of tests allowed that to go unnoticed. Shouldn't happen again with the extra tests, covering both controlgroupLabel and spinner.

### DIFF
--- a/tests/unit/controlgroup/core.js
+++ b/tests/unit/controlgroup/core.js
@@ -9,12 +9,10 @@ define( [
 module( "Controlgroup: Core" );
 
 test( "selectmenu: open/close corners", function( assert ) {
-	expect( 1 );
+	expect( 12 );
 	var element = $( ".controlgroup" ).controlgroup(),
 		selects = element.find( "select" ),
 		selectButton = selects.eq( 0 ).selectmenu( "widget" );
-
-	expect( 12 );
 
 	selects.eq( 0 ).selectmenu( "open" );
 	assert.hasClasses( selectButton, "ui-corner-tl",
@@ -64,6 +62,15 @@ test( "selectmenu: open/close corners", function( assert ) {
 	selects.eq( 2 ).selectmenu( "close" );
 	assert.hasClasses( selectButton, "ui-corner-bottom",
 		"vertical: Last selectmenu gets ui-corner-bottom when closed" );
+} );
+
+test( "selectmenu: controlgroupLabel", function( assert ) {
+	expect( 2 );
+	var element = $( ".controlgroup" ).controlgroup();
+	var label = element.find( ".ui-controlgroup-label" );
+
+	assert.hasClasses( label, "ui-widget ui-widget-content ui-state-default ui-controlgroup-item" );
+	assert.hasClasses( label.find( "span" ), "ui-controlgroup-label-contents" );
 } );
 
 } );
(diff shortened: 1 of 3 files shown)


tests/unit/controlgroup/core.js
```diff
@@ -66,4 +65,12 @@ test( "selectmenu: open/close corners", function( assert ) {
...
 } );
 
+test( "selectmenu: controlgroupLabel", function( assert ) {
+	expect( 2 );
+	var element = $( ".controlgroup" ).controlgroup();
+	var label = element.find( ".ui-controlgroup-label" );
+	assert.hasClasses( label, "ui-widget ui-widget-content ui-state-default ui-controlgroup-item" );
+	assert.hasClasses( label.find( "span" ), "ui-controlgroup-label-contents" );
+} );
+
```
arschmitz: this should also add `controlgroup-label` to `methods.js` to fully test it along with the other widgets in different positions to ensure proper classes. https://github.com/jquery/jquery-ui/blob/master/tests/unit/controlgroup/methods.js#L38

ui/widgets/controlgroup.js
```diff
@@ -61,6 +61,9 @@ return $.widget( "ui.controlgroup", {
 		this._callChildMethod( "destroy" );
 		this.childWidgets.removeData( "ui-controlgroup-data" );
 		this.element.removeAttr( "role" );
+		if ( this.options.items.controlgroupLabel ) {
+			this.element.find( ".ui-controlgroup-label-contents" ).contents().unwrap();
```
arschmitz: i wonder if we should be more explicit and do `this.element.find( this.options.items.controlgroupLabel ).find( ".ui-controlgroup-label-contents" ).contents().unwrap();`

tests/unit/controlgroup/core.js
```diff
@@ -66,4 +65,12 @@ test( "selectmenu: open/close corners", function( assert ) {
 		"vertical: Last selectmenu gets ui-corner-bottom when closed" );
 } );
 
+test( "selectmenu: controlgroupLabel", function( assert ) {
+	expect( 2 );
+	var element = $( ".controlgroup" ).controlgroup();
+	var label = element.find( ".ui-controlgroup-label" );
+	assert.hasClasses( label, "ui-widget ui-widget-content ui-state-default ui-controlgroup-item" );
```
arschmitz: blank line after var statements

tests/unit/controlgroup/core.js
jzaefferer: That test relies on too many properties that only widgets have, like the [widget method](https://github.com/jquery/jquery-ui/blob/master/tests/unit/controlgroup/methods.js#L74) and [pseudo selectors](https://github.com/jquery/jquery-ui/blob/master/tests/unit/controlgroup/methods.js#L120).

jzaefferer:
@arschmitz wrote:

> this should also add `controlgroup-label` to `methods.js` to fully test it along with the other widgets in different positions to ensure proper classes. https://github.com/jquery/jquery-ui/blob/master/tests/unit/controlgroup/methods.js#L38

To which I replied:

> That test relies on too many properties that only widgets have, like the [widget method](https://github.com/jquery/jquery-ui/blob/master/tests/unit/controlgroup/methods.js#L74) and [pseudo selectors](https://github.com/jquery/jquery-ui/blob/master/tests/unit/controlgroup/methods.js#L120).

That should be the only remaining issue, I've addressed the rest.

arschmitz:
:+1:

arschmitz:
except the jscs issue of course :-)

PR #1626, opened 2015-10-26
Selectmenu: Handle form resets

jzaefferer:
Fixes [#12236](http://bugs.jqueryui.com/ticket/12236)
Ref gh-1550


scottgonzalez:
Looks good.

On a related note, @arschmitz: Do you remember what I was working on related to form reset before my laptop died? I feel like I was working on *something* inside the mixin itself, but I have no recollection of any details.

PR #1628, opened 2015-10-26
Tabs: Handle overly-precise values in tests in IE

scottgonzalez:
Fixes the failures shown in http://swarm.jquery.org/index.php?action=result&item=747545


jzaefferer:
Looks good.

PR #1629, opened 2015-10-27
Spinner: Adjust styles for new button CSS

scottgonzalez:


scottgonzalez:
before:
screen shot before change

after:
screen shot after change

arschmitz:
LGTM

PR #1631, opened 2015-10-27
[WIP] Checkboxradio: Properly find radio groups from the associated form

scottgonzalez:
Fixes #9973

Still need to write a test for this.

### DIFF
--- a/tests/unit/checkboxradio/checkboxradio.html
+++ b/tests/unit/checkboxradio/checkboxradio.html
@@ -64,6 +64,13 @@
 <label>
 	<input type="checkbox" id="label-with-no-for"/>
 </label>
+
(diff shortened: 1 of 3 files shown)


jzaefferer:
As discussed, test needs to be limited to browsers with native support for `form` property, using `QUnit.skip` for the others.

scottgonzalez:
So...umm...this test passes in IE 8 because the browser just sees the controls as grouped within the body. So that's cool. Running this test without the fix does result in a failure in browsers that support the `form` attribute, so this test is actually correct and IE 8 *should* pass the test.

jzaefferer:
Nice. As long as we don't have to care about actually submitting the form...

PR #1632, opened 2015-10-28
Button: Escape label option as text

jzaefferer:
Fixes jquery/api.jqueryui.com#281

### DIFF
--- a/tests/unit/button/options.js
+++ b/tests/unit/button/options.js
@@ -86,6 +86,21 @@ test( "label, explicit value", function() {
 	deepEqual( button.button( "option", "label" ), "xxx" );
 } );
 
+test( "label, escape html", function() {
+	expect( 4 );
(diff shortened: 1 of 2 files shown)


scottgonzalez:
Button labels have always been HTML and this is sort of a requirement since you can initialize a button that already has a label which contains HTML. A change like this would require discussion, proper ticket, upgrade guide, etc.

scottgonzalez:
Note that the bug that was actually reported is legit. The `closeText` option on dialog has to be treated as text. But that doesn't mean that the more generic application of button labels must be limited as well.

jzaefferer:
Okay, we can discuss that in the meeting today. If we stick with html labels, we need to add a test for that.

jzaefferer:
Closing this, will create a PR to add tests for the current behaviour instead, and patch dialog to escape the `closeText` option.

PR #1634, opened 2015-10-28
Button: Add tests for button with html markup

jzaefferer:
Ref gh-1632
Ref jquery/api.jqueryui.com#281


scottgonzalez:
The markup needs to be added to the deprecated fixture as well.

jzaefferer:
Updated. Now also passing in IE8, by comparing only lowercase strings. `<SPAN>` anyone?

scottgonzalez:
Looks good.

PR #1635, opened 2015-10-28
Dialog: Escape closeText option before passing it to button

jzaefferer:
Ref gh-1632
Fixes jquery/api.jqueryui.com#281


scottgonzalez:
Looks good.

PR #1636, opened 2015-10-28
Theme: Fix position of some icons

scottgonzalez:
The problem is that the icons are on a 16x16 grid, but many of the arrows use an odd number of pixels, so they can't be exactly centered. This change makes sure that arrows that are likely to be placed together vertically actually align vertically.

----

Before:

screen shot 2015-10-28 at 3 39 29 pm

After:

screen shot 2015-10-28 at 3 39 09 pm

### DIFF
--- a/themes/base/spinner.css
+++ b/themes/base/spinner.css
@@ -50,9 +50,3 @@
 .ui-spinner-down {
 	bottom: 0;
 }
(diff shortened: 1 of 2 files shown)


jzaefferer:
Looks good.

PR #1638, opened 2015-10-29
fix memory leak

yakimko:


jzaefferer:
Why did you close the PR? Is there an issue with the CLA?

yakimko:
There was an issue with CLA, my information and build was failed

PR #1639, opened 2015-10-29
fix memory leak

yakimko:
Maybe, it should be done in the draggable widget too


yakimko:
I signed CLA, but there is still an issue

arschmitz:
@yakimko Sorry about that I do see your signature and everything looks fine we will look into the bug with out bot Sorry!

jzaefferer:
Sorry about that, there's nothing wrong with your signature. We're trying to figure out why the automatic check failed.

PR #1640, opened 2015-10-29
Release fixes

jzaefferer:
Two release blockers. There's probably more, tracking DB issue in https://github.com/jquery/download.jqueryui.com/issues/282


scottgonzalez:
Looks good.

PR #1643, opened 2015-11-02
Menu: Honour aria-label tags as alternative text in ui-menu-item

andrew-oxenburgh:
I would like to be able to create a horizontal list of icons, similar to gmail, or facebook, for really basic functionality.

Currently this is prevented because ui-menu-items without text are made into dividers.

This solution checks for the existence of aria-label text, and will allow that through.

It has passed a full grunt run.

### DIFF
--- a/tests/unit/menu/methods.js
+++ b/tests/unit/menu/methods.js
@@ -45,7 +45,7 @@ test( "enable/disable", function( assert ) {
 } );
 
 test( "refresh", function() {
-	expect( 5 );
+	expect( 8 );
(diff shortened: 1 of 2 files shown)


scottgonzalez:
Thanks for the patch! However, this would be a new feature. As such, we need to have a discussion about this. That discussion should happen in a ticket (http://bugs.jqueryui.com/newticket). Some of the new assertions seems suspicious as well, which is why the discussion is important. It's not clear exactly what use cases you're trying to support. I'm going to close this PR for now so we can have the appropriate discussions before reviewing any code.

PR #1644, opened 2015-11-05
Resizable: Don't use CSS shorthand properties with `.css()`

scottgonzalez:
Fixes #7766

### DIFF
--- a/ui/widgets/resizable.js
+++ b/ui/widgets/resizable.js
@@ -100,6 +100,7 @@ $.widget( "ui.resizable", $.ui.mouse, {
 	_create: function() {
 
 		var n, i, handle, axis, hname,
+			marginTop, marginRight, marginBottom, marginLeft,
 			that = this,
 			o = this.options;
 		this._addClass( "ui-resizable" );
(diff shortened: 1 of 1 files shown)


ui/widgets/resizable.js
```diff
@@ -131,11 +132,16 @@ $.widget( "ui.resizable", $.ui.mouse, {
 
 			this.elementIsWrapper = true;
 
+			marginTop = this.originalElement.css( "marginTop" );
```
dmethvin: This could be DRYed out a bit by creating an object here with the margins and using it for the two calls below.

ui/widgets/resizable.js
```diff
@@ -131,11 +132,16 @@ $.widget( "ui.resizable", $.ui.mouse, {
...
-				marginLeft: this.originalElement.css( "marginLeft" ),
-				marginTop: this.originalElement.css( "marginTop" ),
-				marginRight: this.originalElement.css( "marginRight" ),
-				marginBottom: this.originalElement.css( "marginBottom" )
+				marginLeft: marginLeft,
+				marginTop: marginTop,
+				marginRight: marginRight,
+				marginBottom: marginBottom
 			} );
 			this.originalElement.css( {
```
dmethvin: Setting (as opposed to getting) margins with a shortcut is usually safe, could this just be `{ margin: 0 }`?

scottgonzalez:
@dmethvin updated

dmethvin:
LGTM!

PR #1645, opened 2015-11-10
Create datepicker-de-at.js

stieler-it:
Localisation for Austria is slightly different from German in Germany. This adds a country specific translation for January, so date picker behaves like AngularJS or fullcalendar does.

### DIFF
--- a/ui/i18n/datepicker-de-at.js
+++ b/ui/i18n/datepicker-de-at.js
@@ -1,0 +1,38 @@
+/* Austrian initialisation for the jQuery UI date picker plugin. */
+/* Written by Michael Stieler (github.com/stieler-it), only slightly adapted from
+   the German version by Milian Wolff (mail@milianw.de). */
+( function( factory ) {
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Thanks for the pull request, but we're no longer accepting new locales for datepicker. We're in the process of rewriting datepicker to use CLDR for all locale information so that we are no longer maintaining any of this information ourselves.

stieler-it:
Thanks, I see, so I won't have to repair the PR.

scottgonzalez:
The good news is that de-AT exists in CLDR, along with hundreds of other locales that we don't currently support. You can take a look at http://unicode.org/cldr/trac/browser/trunk/common/main/ to see the list of locales currently in CLDR. You can check out the rewrite in the [datepicker branch](https://github.com/jquery/jquery-ui/tree/datepicker) and see notes about the new API in the [wiki](http://wiki.jqueryui.com/Datepicker).

PR #1646, opened 2015-11-12
Build: Use jquery-git in place of jquery-compat-git

mzgol:
jQuery Compat is not going to get released after all; jQuery UI should be tested
against jquery-git instead of jquery-compat-git.

---

Should I report a ticket about that? cc @scottgonzalez @jzaefferer 

Once it's ready to get merged, please notify me so that I update the Jenkins setup to use the `git` version instead of `compat-git`.


scottgonzalez:
No need for a ticket. What are the supported browsers?

mzgol:
> What are the supported browsers?

* Desktop
  * Chrome: (Current - 1) and Current
  * Edge: Current
  * Firefox: (Current - 1) and Current
  * Internet Explorer: 9+
  * Safari: (Current - 1) and Current
  * Opera: Current
  * Yandex.Browser: Current
* Mobile
  * Android: 4.0+ (plus limited 2.3 support)
  * iOS: 7+

mzgol:
Right, so the problem will be you're testing on IE8, it would require using a different browser set for the `git` part...

mzgol:
I'd go with the way jQuery TestSwarm is configured (by accepting [externally passed `browserSets` parameter](https://github.com/jquery/jquery/blob/88046440da8f5433b510ea705255d1df12c2963e/build/tasks/testswarm.js#L15)) and creating a new `jquery-ui+jquery-git` browser set that would be jquery-ui` minus browsers not supported by jQuery Git (currently just IE 8).

Does that sound OK?

scottgonzalez:
Yup. And we already use configured browser sets: https://github.com/jquery/jquery-ui/blob/0cc197a421202b676d8d61e629f27bf7c68af027/build/tasks/testswarm.js#L63

mzgol:
@scottgonzalez These are the browserSets from the JSON config on the TestSwarm server. Core uses it by default by allows to provide a browserSet name as a parameter to the testswarm Grunt task, this is required since the "UI + Core Git" job will need a different browserSet. So IMO a modification is still needed.

mzgol:
@scottgonzalez I submitted PR #1648; this is a pre-requisite to make this one work. PTAL.

mzgol:
PR updated. I modified the Jenkins jobs so that they use this branch so after merging this PR tests should run correctly.

scottgonzalez:
Thanks @mzgol! This looks good to me, but I'd like to have @jzaefferer chime in as well before merging.

mzgol:
> I'd like to have @jzaefferer chime in as well before merging.

Should I disable the "jQuery UI - git core" Jenkins job until he approves it? It will currently fail as it now looks for the "git" key which is not present in [the versions variable](https://github.com/jquery/jquery-ui/blob/cd7f8f02cf84d395bf0f06b752fe8572d4871482/build/tasks/testswarm.js#L5-L12).

scottgonzalez:
You can, but if we end up with one or two failures in the meantime, I don't think anyone will be alarmed :-)

mzgol:
OK, fine. :) I'll wait then.

jzaefferer:
Looks good to me.

PR #1647, opened 2015-11-13
Refactors _updateAlternate() method

benjamin-albert:
As far as i'm aware `.each(function() { $(this).val( ... ) });` is not required to update the value of all elements in the set of matched element.


benjamin-albert:
I just signed the CLA with my full name Benjamin Albert but github made the commit with the name benjamin-albert (I just updated my profile but I don't know how to amend the commit from the browser).

scottgonzalez:
> I just updated my profile but I don't know how to amend the commit from the browser

You can't. You'd have to clone the repo locally to amend the commit. But if you're not comfortable doing that, we can fix this when we merge.

benjamin-albert:
I amended the commit locally

scottgonzalez:
Thanks!

PR #1648, opened 2015-11-13
Build: Allow to pass custom browserSets to the testswarm* tasks

mzgol:
This is needed to be able to test UI with Core Git that doesn't support IE 8.

Ref gh-1646

### DIFF
--- a/build/tasks/testswarm.js
+++ b/build/tasks/testswarm.js
@@ -34,11 +34,13 @@ var versions = {
 		"Widget": "widget/widget.html"
 	};
 
-function submit( commit, runs, configFile, extra, done ) {
+function submit( commit, runs, configFile, extra, browserSets, done ) {
 	var testName,
 		testswarm = require( "testswarm" ),
 		config = grunt.file.readJSON( configFile ).jqueryui,
 		commitUrl = "https://github.com/jquery/jquery-ui/commit/" + commit;
+
+	browserSets = browserSets || config.browserSets;
 
 	if ( extra ) {
 		extra = " (" + extra + ")";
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Can we just manage the browser set for git in the task directly? I'd prefer having that logic contained in the task instead of having to manage the external build settings.

mzgol:
We can but it will be spread anyway as a browserSet with this name has to be defined in the infrastructure repository. Also, as new jQuery versions may drop further browsers, you may end up needing to hardcode more than one browserSet and keep it in sync with the infra repo.

Do you want me hardcode the browserSet for Core Git?

scottgonzalez:
> We can but it will be spread anyway as a browserSet with this name has to be defined in the infrastructure repository.

That's fine. Creating the browser set will be a one time thing. I don't see this expanding to more and more sets.

> Also, as new jQuery versions may drop further browsers, you may end up needing to hardcode more than one browserSet and keep it in sync with the infra repo.

I don't think we'd ever want to handle many browser sets. To be honest, I wouldn't be surprised to see this entire thing go away in just a few months.

scottgonzalez:
> Do you want me hardcode the browserSet for Core Git?

Yes, please.

> Is the name jquery-ui+jquery-3 OK? The regular UI browserSet is named jquery-ui.

Let's just use `jquery-ui-future`.

mzgol:
> Let's just use `jquery-ui-future`.

OK, I'll do it. I'll just note that when we release jQuery 3.0.0 in a few months, `jquery-ui-future` will have to be used for that as well.

mzgol:
I'm closing this PR; everything can be handled directly in #1646.

PR #1649, opened 2015-11-16
_blockFrames() called 2 times

maxazan:
On touch devices method _blockFrames() called 2 times when begin drag element, but unblock only once when drop. So iframe stay blocked and cannot be scrolled.

https://www.youtube.com/watch?v=OOW7qPaflQU


scottgonzalez:
Thanks, but this seems like a treatment of the symptom, not a prevention of the problem. Can you please start by filing a [ticket](http://bugs.jqueryui.com/newticket) with a reduced test case showing the problem? We'll need to track down what causes the frames to be blocked twice and stop that from happening.

PR #1653, opened 2015-11-19
fixed memory leak

yakimko:


fnagel:
Please provide some description about the bug and probably a reduced test case. Is there an existing bug in our bugtracker linked to this?

ui/widgets/draggable.js
```diff
@@ -98,6 +98,10 @@ $.widget( "ui.draggable", $.ui.mouse, {
 		}
 		this._removeHandleClassName();
 		this._mouseDestroy();
+
+		if ( $.ui.ddmanager ) {
```
scottgonzalez: This doesn't seem right. A check for `$.ui.ddmanager.current` being the draggable that's getting destroyed could work, but this implementation is error prone.

scottgonzalez:
Closing due to inactivity.

PR #1654, opened 2015-11-25
fix placeholder's  container is mistaked

xwartz:
Never consider a container that's `display:none`


scottgonzalez:
Thanks for the patch! However, we need a bug report explaining what issue you're trying to solve, along with a reduced test case that demonstrates the problem. Please read through the [contribution guidelines](https://github.com/jquery/jquery-ui/blob/master/CONTRIBUTING.md) for more details, and feel free to ask any questions about the process here or in IRC.

PR #1656, opened 2015-11-26
fix #7288, #9923, #7580 showCurrentAtPos bug

Wikinaut:
Pls. bear with me, it's my first commit to jquery-ui.

This PR fixes the weird datepicker behaviour when showing many months (e.g. [4, 3]) and when using the "showCurrentAtPos" option. a bug which is open since more than 5(five) years.

See http://bugs.jqueryui.com/search?q=showcurrentatpos for all **showCurrentAtPos" related bugs and see what I have posted here http://stackoverflow.com/questions/16955134/jquery-ui-datepicker-weird-behavior/33880029#33880029 

The PR fixes in my view: #7288, #9923, #7580 (what I think are all related and duplicates).

I have to admit, that I did not yet run the tests as mentioned in the contribution guidelines. Pls. bear with me, let me know, if you need further infos or changes. I am willing to learn.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist
 node_modules
 .sizecache.json
+*.bak
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -1023,6 +1023,8 @@ $.extend( Datepicker.prototype, {
 
 		if ( period === "M" ) {
 			this._adjustInstDate( inst, this._get( inst, "showCurrentAtPos" ), period );
+		} else {
+			this._notifyChange( inst );
 		}
 		this._updateDatepicker( inst );
 	},
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -862,6 +862,7 @@ $.extend( Datepicker.prototype, {
 				origyearshtml = inst.yearshtml = null;
 			}, 0 );
 		}
+		inst.drawMonth += this._get( inst, "showCurrentAtPos" );
 	},
 
 	// #6694 - don't focus the input if it's already focused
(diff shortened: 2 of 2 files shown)


Wikinaut:
@jquerybot @scottgonzalez I don't know what exactly the CLA check regarded as a problem - please advice. Yes, I did sign the CLA, several times and repushed the amended PR.

Wikinaut:
@scottgonzalez a comment: I am concentrating my work to fix a problem which prevents - when displaying more than one month (e.g. one or two full years) the positioning of the current month at a fixed place. This is solved now.

If you want I can present an example page (not yet a jsfiddle) which uses the modified code, but all the reported problems appear to be fixed.

Let me know, if you or your teams need something else.

.gitignore
```diff
@@ -1,3 +1,4 @@
 dist
 node_modules
-.sizecache.json
\ No newline at end of file
+.sizecache.json
+*.bak
```
scottgonzalez: You should put this in your own `core.excludesfile` setting. Individual projects should not have rules for specific editors.

fnagel:
Sorry for the late response. We're in the process of rewriting datepicker, check out the rewrite in the [datepicker branch](https://github.com/jquery/jquery-ui/tree/datepicker) and see notes about the new API in the [wiki](http://wiki.jqueryui.com/Datepicker).

Anyway, we're willing to merge bugfix PRs but we have to be careful because we've historically had a lot of regressions caused by bug fixes in datepicker.

Can you provide a reduced test case and a step by step guide for this issue?
These should be helpful: http://jsfiddle.net/tj_vantoll/Qp39S/show/
I've found this one by @scottgonzalez  http://jsfiddle.net/Hsy5w/1/ in https://bugs.jqueryui.com/ticket/7580
Is this still up to date?

Tests seems to run without issues, see https://travis-ci.org/jquery/jquery-ui/builds/94939718
Are you able to provide a unit test that checks for the issue you're fixing with this commit? Would be really helpful and makes sure this won't happen again.

Please see the inline comment by scott regarding the gitignore file.

.gitignore
Wikinaut: thanks for this info. Will do so.

Wikinaut:
@fnagel wrote
> Can you provide a reduced test case and a step by step guide for this issue?

I will try to do so! 

At the moment I only can tell, that my fix really (and still) solves all the issue/s regarding the datapicker.

scottgonzalez:
Closing due to inactivity. If you're still interested in working on this, please just push another update and leave a comment so we can reopen.

Wikinaut:
ok. My patch is still working and okay.

PR #1657, opened 2015-11-26
Effects: Refactor transfer back compat to avoid bad nested return

jzaefferer:
Download builder replaces return statements while concatenating a bundle, but fails to deal with this nested return. Since this module is already deprecated, pulling the return to the top level makes more sense.


scottgonzalez:
Looks good.

PR #1659, opened 2015-12-09
Added onRender callback to datepicker.js

canvaspixels:
Currently there is no way of knowing when the datepicker calendar DOM has loaded. Once this is merged, i'll add to documentation and an example.


canvaspixels:
Currently there is no way of knowing when the datepicker calendar DOM has loaded. Once this is merged, i'll add to documentation and an example. The merge adds an onRender option to the datepicker initialization.

scottgonzalez:
Thanks, but we're not accepting any API changes as datepicker is in the process of a complete rewrite. It's also not clear why this is needed and how its use would be different from `beforeShow`. Please read through our [contribution guidelines](https://github.com/jquery/jquery-ui/blob/master/CONTRIBUTING.md) to see the proper workflow for proposing new features.

PR #1661, opened 2015-12-16
Datepicker Ukrainian Locale

olegbul:
Datepicker Ukrainian Locale


scottgonzalez:
Thanks, but we're not accepting any new locales. We're in the process of a complete rewrite for datepicker and the rewrite uses CLDR for all locales so that we're no longer maintaining any locales ourselves.

PR #1663, opened 2016-01-15
Selectors: Fix metadata

scottgonzalez:
I noticed that tabbable had the wrong label on download builder. While I was fixing that, I decided to make the labels more meaningful.


jzaefferer:
Looks good.

PR #1664, opened 2016-01-17
Slider: Fixed max value miscalculation

mukulhase:
Due to floating point precision while dividing floats followed by Math.floor().

Fixes #12852

### DIFF
--- a/tests/unit/slider/options.js
+++ b/tests/unit/slider/options.js
@@ -43,7 +43,7 @@ test( "disabled", function( assert ) {
 } );
 
 test( "max", function() {
-	expect( 5 );
+	expect( 7 );
 	element = $( "<div></div>" );
 
 	options = {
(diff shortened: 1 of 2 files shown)


mukulhase:
DatePicker test failure is independent of my fix.

scottgonzalez:
Please implement the logic from spinner. Arbitrary precision fixing isn't very elegant.

mukulhase:
There is a behavioral difference between the spinner and the slider, if an invalid default value(not divisible by step) is set:-
- in the spinner, it is rounded up.
- but in the slider, it is floored to the previous allowed value.
Hence I added the offset(instead of the spinner logic), to preserve this flooring behavior which was enforced by one of the previous test cases.
The offset solution may not be elegant, but the check after that makes it an absolute solution.
I could change the behavior to rounding-up instead of flooring(and delete the corresponding test cases).
Any suggestions?

scottgonzalez:
I'm not sure what you mean by "invalid default value" since spinner's don't have default values. Are you referring to the scenario where at the time of initialization, the associated text field has a value which results in a step mismatch? In that case, absolutely nothing is done to the value.

I can't really respond to the other part since it's based on the first part, which I don't understand. Perhaps you can clarify with a jsbin?

mukulhase:
Implemented the logic from the spinner in the latest commit. It will now round-off the max value.

mukulhase:
Is this being merged?

jasongrout:
This new logic still has a problem with the range `(0.0, 0.6)` with step `0.1`: http://jsfiddle.net/f551752v/ - 0.6 is not reachable in this case - the slider only goes up to 0.5.

Perhaps instead of subtracting a step from max, we could build in the knowledge that the last step may not be a full step and keep the user-set maximum (since that's what the user intended, since it was explicitly set).

PR #1665, opened 2016-01-18
Tooltip: Remove use of `$.parseJSON()`

scottgonzalez:
Also fixes the module loading for the tooltip animations visual test.

Fixes #14903

### DIFF
--- a/tests/visual/tooltip/animations.html
+++ b/tests/visual/tooltip/animations.html
@@ -12,11 +12,12 @@
 	}
 	</style>
 	<script src="../../../external/requirejs/require.js"></script>
-	<script src="../../../demos/bootstrap.js">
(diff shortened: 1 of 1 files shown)


rxaviers:
The JSON update LGTM. I'm unfamiliar on how the `data-modules` attribute from the script tag  is used by the demo.

scottgonzalez:
> I'm unfamiliar on how the `data-modules` attribute from the script tag is used by the demo.

They're just split on spaces and then injected into a `require()` call.

rxaviers:
> They're just split on spaces and then injected into a `require()` call.

Ok, thanks for the clarification.

jzaefferer:
:+1:

PR #1666, opened 2016-01-21
Resizable: Implementing setOption for handles

kdinev:
The handles could only be changed post initialization, unless the resizable is destroyed and recreated. Related to issues https://bugs.jqueryui.com/ticket/3423 and https://bugs.jqueryui.com/ticket/4310

Fixes: #3423


kdinev:
I ran all the tests before I make any changes and there is a datepicker test that fails. I left the test to keep failing after my commit as it's not affected by it.

scottgonzalez:
Thanks! The test is a bit awkward, so I'm going to rewrite when I merge this.

scottgonzalez:
@kdinev Would you mind looking into https://bugs.jqueryui.com/ticket/15084, which is a regression caused by this change?

kdinev:
@scottgonzalez I'll take a look at it.

scottgonzalez:
Thanks.

kdinev:
@scottgonzalez I will fix the issue. 

One comment however: I don't think the issue should be marked as regressions, as this functionality didn't exist prior to my change. The `_setOption` could not be performed at all before that, so it's a bug in new functionality, but it's not a regression.

scottgonzalez:
That logically makes sense, but somehow this was working before. See https://jsfiddle.net/2cg12b1j/3/ which is the test case from the ticket updated to use 1.11.4. Clicking the button does add the third handle, even though there's no `_setOption()` defined in resizable. I'm really not sure how that test case works with 1.11.4, but since it does, this seems like a regression.

kdinev:
Makes sense. Leave it as is.

PR #1668, opened 2016-01-22
Update datepicker.js to support showSelectorAtPos in multi-month scenario

aquavark:
Added new setting "showSelectorAtPos" to control which position the month and year selectors appear in (zero based, as per showCurrentAtPos). Allows correction of the non-intuitive layout when you have multiple months, but the month you are selecting does not appear in the box that you are making the selections in.

Works by changing the check for "secondary" when building the month/year part of the month box in multiple month operation.

### DIFF
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -1680,9 +1680,9 @@ $.extend( Datepicker.prototype, {
 			maxDate = this._getMinMaxDate( inst, "max" ),
 			drawMonth = inst.drawMonth - showCurrentAtPos,
 			drawYear = inst.drawYear;
-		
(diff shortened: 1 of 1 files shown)


aquavark:
Failure with Travis was present prior to pull request. CLA failure - have now signed it 4 times!

scottgonzalez:
Thanks, but we're not accepting any new features for datepicker as it is in the middle of a rewrite. In the future, please start by filing an issue to have discussion before sending a pull request. Of course you can still file an issue so we can discuss this for the new version. You can read about the process for contributing in CONTRIBUTING.md.

aquavark:
Which I would (as it's an important usability feature for us), but I can't log in to Trac on the jQuery-UI website, and the IRC developers channel appears to be dead for the past 24 hours. How do I log it?

scottgonzalez:
Just to close the loop on this, @aquavark did file the feature request on the planning wiki.

PR #1672, opened 2016-02-11
Droppable: Consider all greedy children before acting on parents

jpallen:
Previously the decision to call _over for a parent element was made as soon as
it was seen we were moving out of a greedy child. However, it's possible we
were moving into another greedy child in which case we should not call _over
for the parent. To fix this we now do a first pass to assess the state of all
of the elements. If any parent elements have left a greedy child without
entering another one we call _over as before.

Fixes #9389


scottgonzalez:

PR #1673, opened 2016-02-16
Autocomplete: Work around `isContentEditable` bug in Chrome

scottgonzalez:
Fixes #14917

### DIFF
--- a/ui/widgets/autocomplete.js
+++ b/ui/widgets/autocomplete.js
@@ -627,9 +627,11 @@ $.widget( "ui.autocomplete", {
 
 		var editable = element.prop( "contentEditable" );
 
-		return editable === "true" ? true :
-			editable === "false" ? false :
-			this._isContentEditable( element.parent() );
+		if ( editable === "inherit" ) {
+		  return this._isContentEditable( element.parent() );
+		}
+
+		return editable === "true";
 	}
 } );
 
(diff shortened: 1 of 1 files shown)


ui/widgets/autocomplete.js
```diff
@@ -614,6 +614,22 @@ $.widget( "ui.autocomplete", {
...
+	_isContentEditable: function( element ) {
+		if ( !element.length ) {
+			return false;
+		}
+
+		var editable = element.prop( "contentEditable" );
+
+		return editable === "true" ? true :
+			editable === "false" ? false :
+			this._isContentEditable( element.parent() );
```
jzaefferer: What value does `editable` have in this case? I think this would be easier to understand (assuming its null, replace as needed):

```js
if ( editable === null ) {
  return this._isContentEditable( element.parent() );
}

return editable === "true";
```

scottgonzalez: The third value is `"inherit"`.

Your proposed logic would traverse before checking for an explicitly non-editable element. Traversing always has to be the last operation for inherited values.

jzaefferer: This seems fine to me:
```js
if ( editable === "inherit" ) {
  return this._isContentEditable( element.parent() );
}

return editable === "true";
```
Assuming "true", "false" and "inherit" are the only values `editable` ever has, that should behave exactly like yours.

scottgonzalez: updated

PR #1674, opened 2016-02-17
Build: Upgrade to grunt-html 6.0.0

scottgonzalez:


scottgonzalez:
Well, that confirms that specifying the JDK doesn't work for node projects.

jzaefferer:
We can try the trusty image, via:

```
sudo: required
dist: trusty
```

See https://github.com/travis-ci/travis-ci/issues/4090#issuecomment-184864134

scottgonzalez:
Tested in #1679, but it didn't seem to load the newer version of Java.

PR #1677, opened 2016-02-19
Update datepicker.js. Improve keyboard support

piotros:
I found the problem with keyboard support feature.
Take a look at this example: https://jsfiddle.net/mye1oy63/5/ It's impossible to go (using Ctrl+arrows) to dates later than 7 April 2016. My little change in code fixes this bug.


ui/widgets/datepicker.js
```diff
@@ -1914,7 +1914,7 @@ $.extend( Datepicker.prototype, {
 	/* Adjust one of the date sub-fields. */
 	_adjustInstDate: function( inst, offset, period ) {
 		var year = inst.drawYear + ( period === "Y" ? offset : 0 ),
```
scottgonzalez: I'm going to update this line as well when I merge this fix.

PR #1681, opened 2016-03-01
1 10 stable

chandra381:
is LiveQuery is depricated in JQuery 1.10.2 version.. because i am receiving .livequery is not a function error after upgrading my application from Jquery 1.9 to 1.10.2 version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,2 @@
 dist
 node_modules
-docs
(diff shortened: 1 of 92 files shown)


scottgonzalez:
I don't know what you're trying to accomplish, but this has nothing to do with jQuery UI.

PR #1682, opened 2016-03-04
Widgets: Add RTL support

AhmedMustafa:
A starting point for UI mirroring support in jquery-ui.

### DIFF
--- a/themes/base/accordion.css
+++ b/themes/base/accordion.css
@@ -13,22 +13,11 @@
 	cursor: pointer;
 	position: relative;
 	margin: 2px 0 0 0;
-	padding: .5em .5em .5em .7em;
+	padding: .5em;
 	font-size: 100%;
 }
 .ui-accordion .ui-accordion-content {
 	padding: 1em 2.2em;
 	border-top: 0;
 	overflow: auto;
-}
-
-/* RTL support */
-.ui-accordion-rtl {
-	direction: rtl;
-}
-.ui-accordion-rtl .ui-accordion .ui-accordion-header {
-	padding: .5em .7em .5em .5em;
-}
-.ui-accordion-header-iconRtl.ui-icon {
-	transform: scaleX(-1);
 }
--- a/themes/base/accordion.css
+++ b/themes/base/accordion.css
@@ -21,3 +21,14 @@
 	border-top: 0;
 	overflow: auto;
 }
+
+/* RTL support */
+.ui-accordion-rtl {
+	direction: rtl;
+}
+.ui-accordion-rtl .ui-accordion .ui-accordion-header {
+	padding: .5em .7em .5em .5em;
+}
+.ui-accordion-header-iconRtl.ui-icon {
+	transform: scaleX(-1);
+}
--- a/ui/widgets/accordion.js
+++ b/ui/widgets/accordion.js
@@ -78,6 +78,9 @@ return $.widget( "ui.accordion", {
 		var options = this.options;
 
 		this.prevShow = this.prevHide = $();
+		if ( this.isRtl() ) {
+			this._addClass( ".ui-accordion-rtl" );
+		}
 		this._addClass( "ui-accordion", "ui-widget ui-helper-reset" );
 		this.element.attr( "role", "tablist" );
 
@@ -108,6 +111,9 @@ return $.widget( "ui.accordion", {
 
 		if ( icons ) {
 			icon = $( "<span>" );
+			if ( this.isRtl() ) {
+				this._addClass( icon, "ui-accordion-header-iconRtl" );
+			}
 			this._addClass( icon, "ui-accordion-header-icon", "ui-icon " + icons.header );
 			icon.prependTo( this.headers );
 			children = this.active.children( ".ui-accordion-header-icon" );
--- a/ui/widgets/accordion.js
+++ b/ui/widgets/accordion.js
@@ -78,9 +78,6 @@ return $.widget( "ui.accordion", {
 		var options = this.options;
 
 		this.prevShow = this.prevHide = $();
-		if ( this.isRtl() ) {
-			this._addClass( ".ui-accordion-rtl" );
-		}
 		this._addClass( "ui-accordion", "ui-widget ui-helper-reset" );
 		this.element.attr( "role", "tablist" );
 
@@ -111,9 +108,6 @@ return $.widget( "ui.accordion", {
 
 		if ( icons ) {
 			icon = $( "<span>" );
-			if ( this.isRtl() ) {
-				this._addClass( icon, "ui-accordion-header-iconRtl" );
-			}
 			this._addClass( icon, "ui-accordion-header-icon", "ui-icon " + icons.header );
 			icon.prependTo( this.headers );
 			children = this.active.children( ".ui-accordion-header-icon" );
--- a/ui/widgets/autocomplete.js
+++ b/ui/widgets/autocomplete.js
@@ -223,6 +223,9 @@ $.widget( "ui.autocomplete", {
 			.menu( "instance" );
 
 		this._addClass( this.menu.element, "ui-autocomplete", "ui-front" );
+		if ( this.isRtl() ) {
+			this._addClass( this.menu.element, "ui-autocomplete-menu-rtl" );
+		}
 		this._on( this.menu.element, {
 			mousedown: function( event ) {
 
@@ -539,9 +542,10 @@ $.widget( "ui.autocomplete", {
 		// Size and position menu
 		ul.show();
 		this._resizeMenu();
+		var pos = this.isRtl() ? { my: "right top", at: "right bottom", collision: "none" } : this.options.position;
 		ul.position( $.extend( {
 			of: this.element
-		}, this.options.position ) );
+		}, pos ) );
 
 		if ( this.options.autoFocus ) {
 			this.menu.next();
--- a/ui/widgets/autocomplete.js
+++ b/ui/widgets/autocomplete.js
@@ -224,7 +224,7 @@ $.widget( "ui.autocomplete", {
 
 		this._addClass( this.menu.element, "ui-autocomplete", "ui-front" );
 		if ( this.isRtl() ) {
-			this._addClass( this.menu.element, "ui-autocomplete-menu-rtl" );
+			this._addClass( this.menu.element, "ui-rtl" );
 		}
 		this._on( this.menu.element, {
 			mousedown: function( event ) {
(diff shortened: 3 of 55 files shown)


themes/base/accordion.css
```diff
@@ -21,3 +21,14 @@
 	border-top: 0;
 	overflow: auto;
 }
+
+/* RTL support */
+.ui-accordion-rtl {
+	direction: rtl;
```
scottgonzalez: Why does this need to exist? The direction is already set or we wouldn't be adding this class.

themes/base/accordion.css
```diff
@@ -21,3 +21,14 @@
...
+}
+.ui-accordion-rtl .ui-accordion .ui-accordion-header {
+	padding: .5em .7em .5em .5em;
+}
+.ui-accordion-header-iconRtl.ui-icon {
+	transform: scaleX(-1);
```
scottgonzalez: This seems wrong. The user has full control over what icons are used via the `icons` option.

ui/widgets/accordion.js
```diff
@@ -78,6 +78,9 @@ return $.widget( "ui.accordion", {
 		var options = this.options;
 
 		this.prevShow = this.prevHide = $();
+		if ( this.isRtl() ) {
+			this._addClass( ".ui-accordion-rtl" );
```
scottgonzalez: This shouldn't have a leading dot. I'm also not sure it should be a keyed class.

themes/base/accordion.css
```diff
@@ -21,3 +21,14 @@
 	border-top: 0;
 	overflow: auto;
 }
+
+/* RTL support */
+.ui-accordion-rtl {
+	direction: rtl;
+}
+.ui-accordion-rtl .ui-accordion .ui-accordion-header {
```
scottgonzalez: This is overly specific (and actually wrong). `.ui-accordion-rtl .ui-accordion-header` is what you actually want.

I'd personally prefer just changing the rule to `padding: .5em` and not having anything for RTL here.

arschmitz: +1

themes/base/accordion.css
```diff
@@ -21,3 +21,14 @@
...
+	padding: .5em .7em .5em .5em;
+}
+.ui-accordion-header-iconRtl.ui-icon {
+	transform: scaleX(-1);
+}
```
arschmitz: missing new line at end of file

arschmitz:
If i'm not mistaken, we don't actually need any changes to the widget or css here. Unless i'm missing something the Demo works with no changes except adding the demo.

I reverted all changes except adding the demo and it does not change the display at all.

themes/base/accordion.css
AhmedMustafa: That is correct for the current case. But it would be required if there is a parent widget with `dir ='rtl'` & its child widget exists somewhere in the DOM with no `dir `attribute (e.g. autocomplete & menu).
I know it doesn't make any sense her, but I added it in case we need it at some point & for the mirroring support to be consistent across all the widgets.

themes/base/accordion.css
AhmedMustafa: The support aims to make the changes required to mirror the app as minimal as possible (just set `dir='rtl'`).
So I think it is better for the user to write his app once & just change one attribute (`dir`) to get his app mirrored.

themes/base/accordion.css
AhmedMustafa: +1

AhmedMustafa:
@arschmitz Accordion & some other widgets don't require extensive modifications to support ui mirroring. However, a little changes should be made & in this case we have to make sure that the icons displayed correctly in rtl layout.

themes/base/accordion.css
scottgonzalez: We would never need this for accordion, so it doesn't make sense to add it.

themes/base/accordion.css
scottgonzalez: I very much disagree with this. This penalizes every user who uses RTL and a proper non-symmetrical icon. The documentation would have to tell users to intentionally build their icons backward or override our CSS.

arschmitz: Also many icons should not be mirrored for example play icons fastforward rewind etc all should never be mirrored mirroring is specific to the icon chosen.

AhmedMustafa: that makes sense

AhmedMustafa:
Based on your comments, Accordion doesn't need special handling to support ui-mirroring (except a minor css change).
So, I did the required changes & I am going to step to the next widget.

ui/widgets/autocomplete.js
```diff
@@ -539,9 +542,10 @@ $.widget( "ui.autocomplete", {
 		// Size and position menu
 		ul.show();
 		this._resizeMenu();
+		var pos = this.isRtl() ? { my: "right top", at: "right bottom", collision: "none" } : this.options.position;
```
scottgonzalez: This can't be hard-coded like this. It makes the option impossible to use.

ui/widgets/autocomplete.js
```diff
@@ -223,6 +223,9 @@ $.widget( "ui.autocomplete", {
 			.menu( "instance" );
 
 		this._addClass( this.menu.element, "ui-autocomplete", "ui-front" );
+		if ( this.isRtl() ) {
+			this._addClass( this.menu.element, "ui-autocomplete-menu-rtl" );
```
scottgonzalez: This seems like it should just be a generic `.ui-rtl` class.

arschmitz: agreed on generic but i wonder should we even use a class or should we set the attribute?

scottgonzalez: Well, I guess it depends on whether this will be used exclusively on generated content, which it might. If it is, then just setting the attribute is fine since we don't need to clean it up. Cleaning up the class would be easier.

jzaefferer:
Btw. the CLA check failed because the commit author (AhmedMustafa) doesn't match the CLA signature (Ahmed Mohamed Mustafa).

ui/widgets/autocomplete.js
AhmedMustafa: Should I create `_getPositionRtl: function( position )` which replace every `left` with `right` & vice versa when `this.isRtl() == true` ?
If so, should I put it in widget.js as it might be used within others widgets ?

scottgonzalez: No, you should assume the user has the option set to what they actually want.

scottgonzalez: If we do anything with position, we would change the default to `null`, and on create, we would set a value based on the direction if the value is still `null`.

arschmitz: That would be consistent with our `null` value options approach in other places

arschmitz: @AhmedMustafa thats right for default, but not if you explicitly change the position. With what we suggested for a default of `null` by default it would show just as you describe in you pic if you did not explicitly set it. but would not override an explicitly set value.

arschmitz: That would make the option unusable if you actually want it based on your design to be the same in either direction. We should never second guess users they are the ones that control both the direction an the option.

arschmitz: a better parallell would be position absolute elements which are NOT mirrored by default. Browsers will never change
```css
position: absolute;
top:0;
left:0;
```
to
```css
position: absolute;
top:0;
right:0;
```

AhmedMustafa: The browser will not change that, but someone have to change it to be acceptable for bidi-users.
You are suggesting that the user should make the changes himself. But I prefer the changes to be made by the library.
So instead of letting every user writing another version of his app for rtl-layout, the library will do that for him by just setting `dir=rtl`. And he still have the ability to disable or override that behavior.

AhmedMustafa: I agree with you, we have to respect the user's choice & that is what we actually do in ltr-layout.
The user will always get what he expects in ltr-layout. But when he choose to enable mirroring(by setting `dir=rtl`), the library should do the magic for him with no or with minimal modification from his side.
The pros of this approach is:
1- The user don't have to write another app or make any major changes to enable ui mirroring.
2- Users don't have to have a deep knowledge about mirroring to make their apps mirrored correctly.
3- The user still have the ability to disable or override that behavoir.

arschmitz: > The browser will not change that, but someone have to change it to be acceptable for bidi-users.

That not always true you can choose to have something in the same position regardless of your layout this is application specific and gives developers a way to force position regardless of direction if thats what their app requires.

> The user still have the ability to disable or override that behavoir.

Please explain to me with your approach how you would set the position to be the same REGARDLESS of direction. As far as i can tell this is not possible with your approach. Also if someone was making a RTL first app do you really expect them to set position left when they really want right? this seems completely wrong an counter intuitive and will only lead to confusion.

scottgonzalez: It's also very easy to write a plugin that does the mirroring for you. Widget functionality is never a black and white choice of the built-in or implemented as a one-off for every single user. Creating a position mirroring extension would allow users to have the mirroring logic without writing it themselves, but I still don't agree that automatic mirroring for position is correct.

AhmedMustafa: You are right. I should revert it back.

AhmedMustafa:
@arschmitz @scottgonzalez  I am done here & waiting your feedback.

ashensis:
I found some GUI mirroring issues while preparing demo for https://bugs.jqueryui.com/ticket/14477.
Specifically in demos/widget/textDir.html, the Menu, Select menu and Tabs doesn't seemto be properly mirrorred
in right column that has wrapping 'div' element with "rtl" dir attribute that apparently should suffice
to ensure the proper GUI mirroring of corresponding widgets.

ashensis:
gui_mirroring_jjquery_widgetspng

ashensis:
gui_mirroring_jquery_widgets

AhmedMustafa:
It has been a while since I receive any feedback from your side regarding this PR. 
If you still convinced that some widgets don't need such a support and that the library is flexible enough to render correctly in RTL pages, then it will be great to mention that somewhere in jQueryUI documentation. This will provide the bidirectional-languages users with the support they need to mirror their pages easily & quickly.
For example, This could be part of the i18n section.

AhmedMustafa:
And for the other widgets (e.g. The Slider), I think you have to consider reviewing the code I have submitted above

mrashad2017:
Dears, 
Based on AhmedMustafa comment above can you advise to add RTL  documentation. This will provide  bidirectional-languages users with the support they need to mirror their pages easily & quickly. And will avoid to do the same effort and same investigation.

mrashad2017:
Dears @arschmitz  @jzaefferer ,

We do some investigation in mirroring, and we have valuable feedback to help the bidirectional-languages users with the support they need to mirror their pages easily & quickly.

This will minimize any effort to do mirroring as we can support them with code details. 

Appreciate your guide to make the document in the proper format and the proper place, and may be after your review, you can reference to it in case of mirroring for bidirectional-languages users.

Attached file contain sample of our change to make mirroring
[jQueryUI_mirror_document.docx](https://github.com/jquery/jquery-ui/files/1473924/jQueryUI_mirror_document.docx)

dmethvin:
Closing as a stale branch.

PR #1683, opened 2016-03-11
Menu: Remove active class from top-level item when menu is blurred

jzaefferer:
This issue was introduced by 0bbd1569182bc03e8dc4f5f8aa203e8edbe15f99, which reduced the use of ui-state-focus and ui-state-active to using only ui-state-focus. This introduced the issue addressed here.

The fix is more of a workaround. With test test in place, we can investigate a better solution in the future.

Fixes [#14919](https://bugs.jqueryui.com/ticket/14919)

### DIFF
--- a/tests/unit/menu/core.js
+++ b/tests/unit/menu/core.js
@@ -73,4 +73,34 @@ asyncTest( "#9532: Need a way in Menu to keep ui-state-active class on selected 
 	} );
 } );
 
+asyncTest( "active menu item styling", function( assert ) {
+	expect( 5 );
+	function isActive( item ) {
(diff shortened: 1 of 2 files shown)


tests/unit/menu/core.js
```diff
@@ -73,4 +73,34 @@ asyncTest( "#9532: Need a way in Menu to keep ui-state-active class on selected
 	} );
 } );
 
+asyncTest( "", function( assert ) {
```
scottgonzalez: Missing name.

scottgonzalez:
One minor issue, but this looks good.

PR #1685, opened 2016-03-23
Selectmenu: Use button markup removing the text span

arschmitz:
Fixes #14938


scottgonzalez:
This caused test failures and a lint error.

PR #1688, opened 2016-03-25
Update selectmenu.js

obaidnaqvi:
Updated :  Select Menu now have default maximum height which would enable option-list scroll , not requiring to scroll complete page
Reference
https://bugs.jqueryui.com/ticket/11606


obaidnaqvi:
I have already signed CLA on
http://contribute.jquery.org/CLA/

please review

scottgonzalez:
Thanks, but please read the [description of what should actually be implemented](https://bugs.jqueryui.com/ticket/11606#comment:2) to fix this. Also note that inline styles are not allowed.

PR #1691, opened 2016-04-02
Add locales

siebrand:
Add de-AT and de-CH locales. Originally submitted as [part of an older version of jQuery UI in MediaWiki core](https://gerrit.wikimedia.org/r/#/c/279605/), but put on hold there until it's submitted and merged in the upstream.


siebrand:
I signed the CLA. I've asked the second author to do the same.

scottgonzalez:
Thanks, but we're not accepting any new locales. We're in the process of rewriting datepicker to use CLDR for locale information so that we don't have to manage all of this data.

PR #1694, opened 2016-04-05
Dialog: 	fix html enconding on dialog title 

Gread:


scottgonzalez:
This is not a bug. The titles are text.

Gread:
the problem is that the text function does not support accentuation, try putting a title in Portuguese for example:
"Não," which in English means "No".
I am not saying that the title must be an html, I mean that the function 'text' does not support encoding for show the accentuation, and accents in texts exist in other languages.
with html() function, the text is displayed correctly

scottgonzalez:
> the problem is that the text function does not support accentuation

I'm not sure how you got this impression. It works just fine and always has: http://jsbin.com/hiyubelili/edit?html,js,output

Gread:
funny
I made the mistake of not having tested without further interference, any reference here in my solution should is contributing to my local problem.
thank's for your time.

PR #1695, opened 2016-04-12
Sortable: fix incorrect top containment for document

maksimr:


scottgonzalez:
Thanks for the pull request. Before we can accept this, we need a few more actions from you. Please file a [bug report](https://bugs.jqueryui.com/newticket) with an explanation of the problem along with a reduced test case that shows the problem. We also need you to sign our [CLA](http://contribute.jquery.org/CLA/).

Thanks.

maksimr:
@scottgonzalez 


I haven't created new issue because there is [issue](https://bugs.jqueryui.com/ticket/14927) about this.

Thanks

PR #1696, opened 2016-04-12
Menu: Don't reset active element until after triggering `blur` event

scottgonzalez:
This ensures that `ui.item` actually exists in the `blur` event.

Fixes #14945


jzaefferer:
:+1:

PR #1698, opened 2016-04-16
Calendar / Datepicker: a11y improvements, classes and icons options

fnagel:
This PR adds some missing a11y functionality to match the specs in wiki and updates the globalize / cldr dependencies.

Besides, I started implementing the classes option using `this._addClass` and `this._removeClass` methods. This required to change the HTML rendering from string concatenation to jQuery elements. For now I changed this for all elements except for the day rendering. For elements created once or a few times this should be no problem at all (header, header buttons, table header, etc.), but the day elements are rendered quite often so this could have some performance impact. 

Please give me some feedback on how to handle day elements and how to improve the classes implementation in general.

### DIFF
--- a/ui/widgets/calendar.js
+++ b/ui/widgets/calendar.js
@@ -54,6 +54,10 @@ return $.widget( "ui.calendar", {
 		},
 		dateFormat: { date: "short" },
 		eachDay: $.noop,
+		icons: {
+			prevButton: "ui-icon-circle-triangle-w",
+			nextButton: "ui-icon-circle-triangle-e"
+		},
 		labels: {
 			"datePickerRole": "date picker",
 			"nextText": "Next",
@@ -309,20 +313,21 @@ return $.widget( "ui.calendar", {
 	_buildHeaderButtons: function() {
 		var buttons = $( "<div>" );
 
-		this.prevButton = $( "<button>", {
-			html: "<span class='ui-icon ui-icon-circle-triangle-w'></span>"
-		} );
-		this.nextButton = $( "<button>", {
-			html: "<span class='ui-icon ui-icon-circle-triangle-e'></span>"
-		} );
-
-		this._addClass( buttons, "ui-calendar-header-buttons" )
-			._addClass( this.prevButton, "ui-calendar-prev" )
-			._addClass( this.nextButton, "ui-calendar-next" );
+		this._addClass( buttons, "ui-calendar-header-buttons" );
 
 		return buttons
-			.append( this.prevButton )
-			.append( this.nextButton );
+			.append( this.prevButton = this._buildIconButton( "prev" ) )
+			.append( this.nextButton = this._buildIconButton( "next" ) );
+	},
+
+	_buildIconButton: function( key ) {
+		var button = $( "<button>" ),
+			icon = $( "<span>" );
+
+		this._addClass( button, "ui-calendar-" + key )
+			._addClass( icon, null, "ui-icon " + this.options.icons[ key + "Button" ] );
+
+		return button.append( icon );
 	},
 
 	_buildHeader: function() {
--- a/ui/widgets/calendar.js
+++ b/ui/widgets/calendar.js
@@ -45,6 +45,13 @@ return $.widget( "ui.calendar", {
 	version: "@VERSION",
 	options: {
 		buttons: [],
+		classes: {
+			"ui-calendar": "ui-corner-all",
+			"ui-calendar-header-first": "ui-corner-left",
+			"ui-calendar-header-last": "ui-corner-right",
+			"ui-calendar-prev": "ui-corner-all",
+			"ui-calendar-next": "ui-corner-all"
+		},
 		dateFormat: { date: "short" },
 		eachDay: $.noop,
 		labels: {
@@ -119,7 +126,7 @@ return $.widget( "ui.calendar", {
 	},
 
 	_hover: function( event ) {
-		$( event.currentTarget ).toggleClass( "ui-state-hover" );
+		this._addClass( $( event.currentTarget ), null, "ui-state-hover" );
 	},
 
 	_handleKeydown: function( event ) {
@@ -193,14 +200,15 @@ return $.widget( "ui.calendar", {
 	},
 
 	_updateDayElement: function( state ) {
-		var id = this._getDayId( this.date );
+		var id = this._getDayId( this.date ),
+			button = this._getDateElement( id ).children( "button" );
 
-		this.grid
-			.attr( "aria-activedescendant", id )
-			.find( "button." + state )
-			.removeClass( state );
+		this.grid.attr( "aria-activedescendant", id );
 
-		return this._getDateElement( id ).children( "button" ).addClass( state );
+		this._removeClass( this.grid.find( "button." + state ), null, state );
+		this._addClass( button, null, state );
+
+		return button;
 	},
 
 	_getDateElement: function( id ) {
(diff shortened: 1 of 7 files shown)


ui/widgets/calendar.js
```diff
@@ -232,146 +239,157 @@ return $.widget( "ui.calendar", {
...
 			this.gridId = this.id + "-" + i;
+			labelledBy.push( this.gridId + "-title" );
 
-			labelledby.push( this.gridId + "-title" );
-			headerClass = "ui-calendar-header ui-widget-header ui-helper-clearfix";
+			html = $( "<div>" );
+			this._addClass( html, "ui-calendar-group" );
+
+			header = this._buildHeader();
+			this._removeClass( header, "ui-corner-all" );
```
fnagel: I guess there is a better way to handle this using classes options?

fnagel:
@scottgonzalez @arschmitz @jzaefferer Please take a look!

ui/widgets/calendar.js
```diff
@@ -45,8 +45,18 @@ return $.widget( "ui.calendar", {
 	version: "@VERSION",
 	options: {
 		buttons: [],
+		classes: {
```
fnagel: Please help what else classes we want to add here!

ui/widgets/calendar.js
scottgonzalez: You'll need to use different classes for single and multi-picker instances. This will allow the default values for those to apply the correct corner classes. See `ui-selectmenu-button-open` and `ui-selectmenu-button-closed` for a similar example.

scottgonzalez: Or more likely what we need in this case is three classes: first, middle, last. And we'll just apply first and last when it's single.

ui/widgets/calendar.js
```diff
@@ -230,146 +243,158 @@ return $.widget( "ui.calendar", {
...
-				this._getTranslation( "datePickerRole" ) +
-			"</span>" +
-		"</div>";
+	_buildHeader: function() {
+		var header = $( "<div>" ),
+			title = $( "<div>", { role: "header", id: this.gridId + "-title" } ),
+			notice = $( "<span>" ).text( ", " + this._getTranslation( "datePickerRole" ) );
+
+		this._addClass( header, "ui-calendar-header", "ui-widget-header ui-helper-clearfix" )
+			._addClass( notice, "ui-helper-hidden-accessible" );
```
scottgonzalez: This is an extra class, not a keyed class.

ui/widgets/calendar.js
```diff
@@ -230,146 +243,158 @@ return $.widget( "ui.calendar", {
...
+			month = $( "<span>" ).text( this.viewDate.monthName() ),
+			year = $( "<span>" ).text( this.viewDate.year() );
+
+		this._addClass( title, "ui-calendar-title" )
+			._addClass( month, "ui-calendar-month" )
+			._addClass( year, "ui-calendar-year" );
+
+		return title
+			.append( month )
+			.append( " " )
```
scottgonzalez: @timmywil Is this a safe use of `.append()`?

ui/widgets/calendar.js
```diff
@@ -230,146 +243,158 @@ return $.widget( "ui.calendar", {
...
-			this._buildGridHeading() +
-			this._buildGridBody() +
-		"</table>";
+		var table = $( "<table>", {
+				role: "grid",
+				tabindex: 0,
+				"aria-readonly": true,
+				"aria-labelledby": this.gridId + "-month-label",
+				"aria-activedescendant": this._getDayId( this.date )
+			} );
```
scottgonzalez: Outdent one level.

ui/widgets/calendar.js
fnagel: Info: needed to add some space between those elements

ui/widgets/calendar.js
fnagel: Fixed.

ui/widgets/calendar.js
fnagel: Changed.

ui/widgets/calendar.js
fnagel: @scottgonzalez Added three classes as advised. Please take a look!

fnagel:
@scottgonzalez @jzaefferer Updated branch, please take a look!

fnagel:
I've got some more commits with added and improved unit tests and some minor bugfixes. I will add a new PR as soon this one has been merged.

@scottgonzalez @jzaefferer 
Question: in wiki there is a "Add change event (datepicker only)" todo item, but we never added this to the specs. Do we want this event? And do we really want to add it datepicker only when it needs to be implemented in calendar anyway (I'm pretty sure about this)?

Related: "Think about parameter for callback events" -- another one from the todo list which needs some more specification.

scottgonzalez:
> Question: in wiki there is a "Add change event (datepicker only)" todo item, but we never added this to the specs. Do we want this event? And do we really want to add it datepicker only when it needs to be implemented in calendar anyway (I'm pretty sure about this)?

We have `select` which exists in both and is triggered only on a click or ENTER (actually, on `mousedown`, but I'm not sure why that is, seems like it should be on `click`)<sup>1</sup>. This will inform the user of all changes on the calendar. However, these events are not filtered, so selecting the currently selected date will result in another `select` event being triggered.

For datepicker, you can change the date without selecting a date from the calendar by typing a new date into the text input. On `blur`, we want to capture that the value has changed and trigger a `change` event. I assume this is why `change` is marked as datepicker only. I'm not strongly opposed to adding `change` to calendar, but it would be duplicative of `select` and would not have the additional useful cases that datepicker has.

On the other hand, we do have behavioral differences across widgets regarding programmatic value changes and triggering a `change` event. A widget like autocomplete which has no internal concept of a value will not trigger `change` on programmatic change of the input (because the value is changed outside of the widget). However, something like slider which has no concept of an external value will (because the value is changed inside the widget). So you could easily make the argument that calendar **should** have a `change` event and we should be triggering it whenever the value is changed programmatically. And if we're doing that, we should be triggering consistently when the value changes via any method.

SO IN CONCLUSION, let's add `change` to both :-)

_<sup>1</sup> We should also change ENTER to not trigger `mousedown`, but to instead just invoke a function which does the right thing._

fnagel:
Thanks for your feedback! I will implement this soon.

> We should also change ENTER to not trigger mousedown, but to instead just invoke a function which does the right thing.

I actually already changed this in my latest local version due to inconsistent event return value and, well, because it looked just wrong :-D

ui/widgets/calendar.js
fnagel: @timmywil Would you mind to give some feedback to the comment above? Thanks!

fnagel:
@scottgonzalez @jzaefferer @arschmitz Any chance to get another review on this one?

And: any idea why there are no CI checks for this PR?

scottgonzalez:
> And: any idea why there are no CI checks for this PR?

I think this is because @jzaefferer had accidentally turned Travis off for a while when he intended to turn it off for jqueryui.com. I bet it will work again on the next push.

ui/widgets/calendar.js
timmywil: I don't see a problem. This creates a text node and appends it.

ui/widgets/calendar.js
scottgonzalez: This is fine if these are the only places we have style classes. We just need to document the full list of structural classes in the API docs.

ui/widgets/calendar.js
```diff
@@ -230,146 +244,162 @@ return $.widget( "ui.calendar", {
...
+			labelledBy.push( this.gridId + "-title" );
 
-			labelledby.push( this.gridId + "-title" );
-			headerClass = "ui-calendar-header ui-widget-header ui-helper-clearfix";
-			if ( months[ i ].first ) {
-				headerClass += " ui-corner-left";
-			} else if ( months[ i ].last ) {
-				headerClass += " ui-corner-right";
-			}
+			html = $( "<div>" );
```
scottgonzalez: This variable name doesn't really make sense anymore.

ui/widgets/calendar.js
```diff
@@ -230,146 +244,162 @@ return $.widget( "ui.calendar", {
...
+			this._addClass( html, "ui-calendar-group" );
 
-			html += "<div class='ui-calendar-group'>";
-			html += "<div class='" + headerClass + "'>" +
-				this._buildTitlebar() + "</div>";
-			html += this._buildGrid() + "</div>";
+			header = this._buildHeader();
+			this._addClass( header, "ui-calendar-header-" + (
```
scottgonzalez: That last opening paren should probably be on the next line.

ui/widgets/calendar.js
```diff
@@ -230,146 +244,162 @@ return $.widget( "ui.calendar", {
...
 
 	_buildGridHeaderCell: function( day ) {
-		return "<th role='columnheader' abbr='" + day.fullname + "' aria-label='" + day.fullname + "'>" +
-			"<span title='" + day.fullname + "'>" +
-				day.shortname +
-			"</span>" +
-		"</th>";
+		return $( "<th role='columnheader' abbr='" + day.fullname + "' aria-label='" + day.fullname + "'>" +
+			"<span title='" + day.fullname + "'>" + day.shortname + "</span>" +
+		"</th>" );
```
scottgonzalez: Seems like you converted all the other multi-attribute elements to use `$(elem, props)` instead of lots of concatenation for HTML.

ui/widgets/datepicker.js
```diff
@@ -279,6 +275,10 @@ var widget = $.widget( "ui.datepicker", {
 
 		this.isOpen = true;
 		this._trigger( "open", event );
+
+		this._delay( function() {
+			this.calendarInstance.grid.focus();
+		}, 1 );
```
scottgonzalez: Drop the duration parameter, since this is the default behavior.

ui/widgets/calendar.js
fnagel: I've converted all string concatenations where it was needed in order to use the classes option. I tried to convert as little as possible due to performance considerations. 

For example: day cell rendering is called quite a lot so this still uses string concatenation and has no support for `classes` option yet.

Do we want to use `$(elem, props)` for _all_ elements?

ui/widgets/datepicker.js
fnagel: Fixed.

ui/widgets/calendar.js
fnagel: Fixed.

ui/widgets/calendar.js
fnagel: Fixed.

ui/widgets/calendar.js
scottgonzalez: > Do we want to use $(elem, props) for all elements?

No, we don't have a mandate on either approach. This just stuck out in the diff. It's fine to leave as is.

fnagel: Ok, ready to merge then? And how to get rid of this "Changes requested" message?

ui/widgets/calendar.js
```diff
@@ -230,146 +244,162 @@ return $.widget( "ui.calendar", {
...
-		this.prevButton = this.element.find( ".ui-calendar-prev" );
-		this.nextButton = this.element.find( ".ui-calendar-next" );
 		this._refreshHeaderButtons();
-
 		this._createButtonPane();
 
 		this.grid = this.element.find( ".ui-calendar-calendar" );
 	},
 
+	_builSinglePicker: function() {
```
scottgonzalez: typo

fnagel: Fixed.

PR #1699, opened 2016-04-20
Tooltip: Fix re-enabling of delegated tooltips

scottgonzalez:
Fixes #14950

### DIFF
--- a/ui/widgets/tooltip.js
+++ b/ui/widgets/tooltip.js
@@ -114,6 +114,8 @@ $.widget( "ui.tooltip", {
 			} )
 			.appendTo( this.document[ 0 ].body );
 		this._addClass( this.liveRegion, null, "ui-helper-hidden-accessible" );
+
+		this.disabledTitles = $( [] );
 	},
 
 	_setOption: function( key, value ) {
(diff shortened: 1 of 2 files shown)


ui/widgets/tooltip.js
```diff
@@ -143,25 +145,29 @@ $.widget( "ui.tooltip", {
...
 		// Remove title attributes to prevent native tooltips
-		this.element.find( this.options.items ).addBack().each( function() {
-			var element = $( this );
-			if ( element.is( "[title]" ) ) {
-				element
-					.data( "ui-tooltip-title", element.attr( "title" ) )
-					.removeAttr( "title" );
-			}
-		} );
+		this.disabledTitles = this.disabledTitles.add(
```
jzaefferer: Why the `.add()` call (which tends to be rather slow)?

scottgonzalez: I wanted to cover `.disable().disable()` and makes sure the collection wasn't overwritten.

jzaefferer: `this.disabledTitles = this.element.find( this.options.items ).addBack() [...]` would be simpler/faster, could also skip allocating the empty object.

scottgonzalez: That would result in `this.disabledTitles` being an empty set after `.disable().disable()` using default settings because `this.options.items` looks for elements with titles, but the titles are removed after the first call.

PR #1700, opened 2016-04-24
_on,_off method handing the "In the widget" event name

ckimdotcn:
_on
_off
These method handing the full events name should be generated automatically (like _trigger)
And I can do..
in  jquery.button.js  I binded the '**press**' event
`this._on(this.element,{'press',btnStartPress})`

call the '**press**' event outside
`$('.btn').button().trigger('buttonpress')`

also like 
bind the '**press**' event outside
`$('.btn').button().bind('buttonpress',function(){console.log('hello world'});`

in jquery.button.js i trigger '**press**' event
`this._trigger('press')`

### DIFF
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -581,7 +581,10 @@ $.Widget.prototype = {
 			}
 
 			var match = event.match( /^([\w:-]*)\s*(.*)$/ );
-			var eventName = match[ 1 ] + instance.eventNamespace;
+			//var eventName = match[ 1 ] + instance.eventNamespace;
+			var eventName = ( match[1] === instance.widgetEventPrefix ? match[1]
+				: instance.widgetEventPrefix + match[1]).toLowerCase() + instance.eventNamespace,
+			//like _trigger(). The full event type will be generated automatically
 			var selector = match[ 2 ];
 
 			if ( selector ) {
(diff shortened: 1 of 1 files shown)


scottgonzalez:
This should definitely not happen. The use case doesn't even make sense. A widget shouldn't be listening to its own events.

ckimdotcn:
Thank you for your reply. as you can see i'm learning english.
I think 

> _trigger method can't trigger 'click', _trigger only trigger its own events.
> so on and so forth. _on, _off method can't bind 'click', _on, _off only bind its own events.

to bind 'click' event, here is the better way that i think
```
this.element.bind('click',function(){
	console.log('element clicked');
});
```

scottgonzalez:
> _trigger method can't trigger 'click', _trigger only trigger its own events.

Correct, because all triggered events must be isolated from native events.

> so on and so forth. _on, _off method can't bind 'click', _on, _off only bind its own events.

Absolutely they can, and your change makes it so they can't.

> to bind 'click' event, here is the better way that i think

This completely defeats the purpose of `._on()` and `._off()` which is to manage clean up.

It seems like you don't understand why these three methods exist. If you're looking for information, please read the [API docs](http://api.jqueryui.com/jQuery.widget/#jQuery-Widget2) and [articles](http://learn.jquery.com/jquery-ui/widget-factory/). If you still have questions, please use support forums such as IRC, the jQuery forums, or Stack Overflow.

PR #1701, opened 2016-05-01
Update datepicker-pt-BR.js

renatomsteiner:
Shortening dayNamesMin to respect layout width.


fnagel:
@renatomsteiner Do you have any sources / references for those shortcuts?

rxaviers:
As a PT-BR spearker, it seems quite right to me. Anyway, always having a reference seems good to me. I believe CLDR should be the reference in such cases and it doesn't define such length: https://github.com/unicode-cldr/cldr-dates-modern/blob/master/main/pt/ca-gregorian.json#L106

scottgonzalez:
Unfortunately, we cannot deviate from CLDR, even if the changes make sense. We are moving away from maintaining our own locale files to using CLDR directly, so deviating in any way will just result in a temporary change since we'll go back to the longer version when the transition to CLDR is complete.

PR #1702, opened 2016-05-11
Performance Improvement

nurolopher:
I would like to suggest using native for loop over $.each since pre-calculated for loop is the fastest way of looping through array elements. Attached an image for the speed demonstration purposes below.

for vs $.each

### DIFF
--- a/ui/widgets/autocomplete.js
+++ b/ui/widgets/autocomplete.js
@@ -566,8 +566,8 @@ $.widget( "ui.autocomplete", {
 
 	_renderMenu: function( ul, items ) {
 		var length = items.length, index = 0;
-		for (; index < length; index++) {
-			this._renderItemData( ul, items[index] );
+		for ( ; index < length; index++ ) {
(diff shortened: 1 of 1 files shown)


nurolopher:
Currently, I am having a performance issue on autocomplete. :worried:

jzaefferer:
@nurolopher can you help us reproduce your performance issue? Then we can figure out the right solution. Filing a ticket on https://bugs.jqueryui.com/ should be the first step.

nurolopher:
@jzaefferer  :ok_hand: I will do it a bit later.

scottgonzalez:
I'm going to close this for now because I assume your menu contains many more items than we think is reasonable. We can reopen after discussion in an issue, if necessary.

nurolopher:
@scottgonzalez yes, you are right, looks like I used more items than the reasonable. Changing $.each to for didn't give much performance improvement. Thanks!

PR #1703, opened 2016-05-11
Controlgroup fixes

arschmitz:
Fixes various issues with controlgroups

### DIFF
--- a/tests/unit/controlgroup/core.js
+++ b/tests/unit/controlgroup/core.js
@@ -92,12 +92,12 @@ var assertSanatized = function( assert, initClasses, expectedClasses, message ) 
 
 QUnit.test( "_resolveClassesValues", function( assert ) {
 	assert.expect( 6 );
-	assertSanatized( assert, "bar ui-corner-bottom", "bar", "Single Corner Class Removed end" );
-	assertSanatized( assert, "ui-corner-bottom bar", "bar", "Single Corner Class Removed beginning" );
-	assertSanatized( assert, "bar ui-corner-bottom ui-corner-left", "bar", "Multiple Corner Class Removed end" );
-	assertSanatized( assert, "ui-corner-bottom ui-corner-left bar", "bar", "Multiple Corner Class Removed beginning" );
-	assertSanatized( assert, "bar ui-corner-bottom ui-corner-left foo", "bar foo", "Multiple Corner Class Removed Middle" );
-	assertSanatized( assert, "bar", "bar", "No corner Class" );
+	assertSanatized( assert, "bar ui-corner-bottom", "bar", "Single corner class removed end" );
+	assertSanatized( assert, "ui-corner-bottom bar", "bar", "Single corner class removed beginning" );
+	assertSanatized( assert, "bar ui-corner-bottom ui-corner-left", "bar", "Multiple corner classes removed end" );
+	assertSanatized( assert, "ui-corner-bottom ui-corner-left bar", "bar", "Multiple corner classes removed beginning" );
+	assertSanatized( assert, "bar ui-corner-bottom ui-corner-left foo", "bar foo", "Multiple corner class removed middle" );
+	assertSanatized( assert, "bar", "bar", "No corner classes" );
 } );
 
 } );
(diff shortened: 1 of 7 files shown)


ui/widgets/controlgroup.js
```diff
@@ -180,22 +186,31 @@ return $.widget( "ui.controlgroup", {
...
 
 			}[ position ]
 		};
 	},
 
+	_resolveClassesValues: function( classes, instance ) {
+		$.each( classes, function( key ) {
+			var current = instance.options.classes[ key ] || "";
+			current = current.replace( removeClassRegex, "" ).trim();
+			classes[ key ] = ( current + " " + classes[ key ] ).replace( / +/g, " " );
```
scottgonzalez: Can you switch the space literal to `\s` in the regex?

ui/widgets/controlgroup.js
```diff
@@ -30,6 +30,7 @@
 		factory( jQuery );
 	}
 }( function( $ ) {
+var removeClassRegex = /ui-corner-([a-z]){2,6}/g;
```
scottgonzalez: If this stays out of the prototype, then it needs to be prefixed with `controlgroup`. The factory goes away during builds and all variables get put in the same scope.

tests/unit/controlgroup/core.js
```diff
@@ -74,4 +74,29 @@ QUnit.test( "selectmenu: controlgroupLabel", function( assert ) {
...
+}
+
+QUnit.test( "_resolveClassesValues", function( assert ) {
+	assert.expect( 6 );
+	assertSanatized( assert, "bar ui-corner-bottom", "bar", "Single Corner Class Removed end" );
+	assertSanatized( assert, "ui-corner-bottom bar", "bar", "Single Corner Class Removed beginning" );
+	assertSanatized( assert, "bar ui-corner-bottom ui-corner-left", "bar", "Multiple Corner Class Removed end" );
+	assertSanatized( assert, "ui-corner-bottom ui-corner-left bar", "bar", "Multiple Corner Class Removed beginning" );
+	assertSanatized( assert, "bar ui-corner-bottom ui-corner-left foo", "bar foo", "Multiple Corner Class Removed Middle" );
+	assertSanatized( assert, "bar", "bar", "No corner Class" );
```
scottgonzalez: Update the casing in the messages so only the first word is capitalized. Fix plurality of class in "multiple corner class".

ui/widgets/controlgroup.js
arschmitz: Oh right ill fix that

ui/widgets/controlgroup.js
```diff
@@ -112,6 +112,11 @@ return $.widget( "ui.controlgroup", {
 				.each( function() {
 					var element = $( this );
 					var instance = element[ widget ]( "instance" );
+
+					// If the button is the child of a spinner ignore it
+					if ( widget === "button" && element.parent( ".ui-spinner" ).length ) {
```
scottgonzalez: Seems like this should really be more generic. Something along the lines of if the found widget is inside any other widget which is also inside this controlgroup, then ignore it. The problem is the "which is also inside any other widget" part. I'm not sure there's anything we can do about this right now, but I really don't like that we're writing logic around specific widgets.

ui/widgets/controlgroup.js
```diff
@@ -87,8 +87,11 @@ return $.widget( "ui.controlgroup", {
 			if ( widget === "controlgroupLabel" ) {
 				labels = that.element.find( selector );
 				labels.each( function() {
-					$( this ).contents()
-						.wrapAll( "<span class='ui-controlgroup-label-contents'></span>" );
+					var element = $( this );
+					if ( !element.children( ".ui-controlgroup-label-contents" ).length ) {
```
scottgonzalez: You can invert this and return to avoid nested even deeper.

ui/widgets/controlgroup.js
arschmitz: I agree I just was not sure how else to do this unless we force enhancment within a controlgroup to be more explicit ala JQM data-role

scottgonzalez:
Looks good.

PR #1704, opened 2016-05-25
Checkboxradio: icons inside dialogs

ChaseWagoner:
Limit a CSS `background` rule that only sets color to `background-color`
to avoid losing the `background-position`.

Fixes #14955 (https://bugs.jqueryui.com/ticket/14955)

Also fixed some JS errors in the other checkboxradio visual test; that commit doesn't necessarily satisfy the "avoid unrelated changes" requirement & can be removed if necessary.

Before, then after this change (screenshots from the added visual test):
14955-pre-post

### DIFF
--- a/themes/base/theme.css
+++ b/themes/base/theme.css
@@ -134,7 +134,7 @@ a.ui-button:active,
 .ui-widget-content .ui-state-highlight,
 .ui-widget-header .ui-state-highlight {
 	border: 1px solid #dad55e/*{borderColorHighlight}*/;
-	background: #fffa90/*{bgColorHighlight}*/ /*{bgImgUrlHighlight}*/ /*{bgHighlightXPos}*/ /*{bgHighlightYPos}*/ /*{bgHighlightRepeat}*/;
+	background-color: #fffa90/*{bgColorHighlight}*/ /*{bgImgUrlHighlight}*/ /*{bgHighlightXPos}*/ /*{bgHighlightYPos}*/ /*{bgHighlightRepeat}*/;
 	color: #777620/*{fcHighlight}*/;
 }
 .ui-state-highlight a,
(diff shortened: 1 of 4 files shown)


themes/base/theme.css
```diff
@@ -134,7 +134,7 @@ a.ui-button:active,
 .ui-widget-content .ui-state-highlight,
 .ui-widget-header .ui-state-highlight {
 	border: 1px solid #dad55e/*{borderColorHighlight}*/;
-	background: #fffa90/*{bgColorHighlight}*/ /*{bgImgUrlHighlight}*/ /*{bgHighlightXPos}*/ /*{bgHighlightYPos}*/ /*{bgHighlightRepeat}*/;
+	background-color: #fffa90/*{bgColorHighlight}*/ /*{bgImgUrlHighlight}*/ /*{bgHighlightXPos}*/ /*{bgHighlightYPos}*/ /*{bgHighlightRepeat}*/;
```
ChaseWagoner: Should the comments about unenforced rules be removed? i.e. this line would read

```css
	background-color: #fffa90/*{bgColorHighlight}*/;
```

scottgonzalez: This change will break ThemeRoller.

ChaseWagoner: @scottgonzalez how can I test that? Is there a way to make ThemeRoller produce individual properties, rather than consolidated?

scottgonzalez: > @scottgonzalez how can I test that?

Testing it is a bit tricky. It would require setting up https://github.com/jquery/download.jqueryui.com/ locally and then using your updated theme in ThemeRoller.

> Is there a way to make ThemeRoller produce individual properties, rather than consolidated?

You'd have to list all the properties. ThemeRoller just does token replacement. But won't producing individual properties cause the same problem?

PR #1706, opened 2016-05-25
Add jQuery 1.12/2.2 to the test configuration + minor updates

mgol:
I've noticed jQuery 1.12/2.2 are not currently tested; I added proper config. Once this lands I can create a new Jenkins job for UI + Core 1.12.

I have a few questions:

1. Are there any plans to test against jQuery Core 2.x? While they're supposed to be API-compatible with 1.x counterparts the internal code is different so they may be subtle bugs. It seems weird to me to test all 4 patch releases of 1.11.x but not to test even one 2.1.x.
2. What's the purpose of the `jquery-1.x` entry in bower.json & in the `external` directory?

I've also added a few improvements to build-related stuff in the followup commits.


scottgonzalez:
> Are there any plans to test against jQuery Core 2.x? While they're supposed to be API-compatible with 1.x counterparts the internal code is different so they may be subtle bugs.

We have the ability to test locally, but we don't test against 2.x in CI. This was done so long ago, I honestly don't remember the reasoning. I can't think of any problems with adding it, we'll just need to use different browser sets.

> It seems weird to me to test all 4 patch releases of 1.11.x but not to test even one 2.1.x.

Re: all 4 patch releases; we'll probably never test against only the latest patch version because that's just not how developers actually work.

Re: 2.1.x; As I said above, I'm ok with adding this now.

scottgonzalez:
> What's the purpose of the jquery-1.x entry in bower.json & in the external directory?

I'm not sure what you're asking. Are you asking why we have all the 1.x versions? It's because we support them. Are you asking why they're in `external`? It's because we don't use `bower_components`.

package.json
```diff
@@ -55,6 +55,7 @@
 		"commitplease": "2.3.0",
 		"grunt": "0.4.5",
 		"grunt-bowercopy": "1.2.4",
+		"grunt-cli": "0.1.13",
```
scottgonzalez: I'm not really a fan of this change. It breaks the standard usage of Grunt.

mgol:
> We have the ability to test locally, but we don't test against 2.x in CI. This was done so long ago, I honestly don't remember the reasoning.

I see. From what I remember the browser set used for testing was hardcoded in the UI repo but now different jobs can use different browser sets (I did it so that UI could be tested against jQuery 3.0.0 WIP) so this can be done. I can update this PR & create proper Jenkins jobs or 2.0, 2.1 & 2.2.

> I'm not sure what you're asking. Are you asking why we have all the 1.x versions?

No, I'm asking about [this entry in bower.json](https://github.com/jquery/jquery-ui/blob/fbc79e1bda2b58f980b9c6fedf3f8ddbce294d7b/bower.json#L39) and [this in `bowercopy` config](https://github.com/jquery/jquery-ui/blob/fbc79e1bda2b58f980b9c6fedf3f8ddbce294d7b/Gruntfile.js#L283-L284).

package.json
mgol: How does it break it? You can still use the global `grunt-cli` if you wish so. OTH:

1. In development all the dependencies are fixed so the possibility of a breakage is decreased. npm scripts can use the `grunt` binary without any global installation.
2. People that don't like global packages may not use them here. Without including `grunt-cli` it's simply not possible as the `grunt` package doesn't contain any binary to run.

Most of the projects I contribute to include `grunt-cli` in `devDependencies` for these reasons. I often don't even have `grunt-cli` installed globally until I encounter a project that doesn't include this.

It seems to me that including it as a dependency doesn't break the usage for people that don't want to rely on it but not including it breaks the usage for others.

Even the `grunt-cli` [README](https://github.com/gruntjs/grunt-cli#installing-grunt-cli-locally) mentions this pattern so I don't understand your comment.

scottgonzalez:
> I can update this PR & create proper Jenkins jobs or 2.0, 2.1 & 2.2.

Sounds good.

> No, I'm asking about this entry in bower.json and this in the bowercopy config.

As I said, we don't use `bower_components`. `bower.json` only exists for the `bowercopy` task. The only files we care about live in `external`.

scottgonzalez:
Oh, sorry. Let me explain the 1.x entry. We need a single reference to jQuery for use in demos. To do that, we use 1.x and copy it to a file that doesn't have a version. This allows us to not have to update a ton of references when we change the latest version we're testing against.

mgol:
@scottgonzalez Thanks for the explanation!

bower.json
```diff
@@ -36,6 +36,11 @@
 		"jquery-1.11.1": "jquery#1.11.1",
 		"jquery-1.11.2": "jquery#1.11.2",
 		"jquery-1.11.3": "jquery#1.11.3",
+		"jquery-1.12.0": "jquery#1.12.0",
+		"jquery-1.12.1": "jquery#1.12.1",
+		"jquery-1.12.2": "jquery#1.12.2",
+		"jquery-1.12.3": "jquery#1.12.3",
+		"jquery-1.12.4": "jquery#1.12.4",
 		"jquery-1.x": "jquery#1.11.3",
```
scottgonzalez: Thanks for pointing this out in your comment (and sorry for taking so long to understand what you were asking). This should be updated to point to 1.12.4; it always tracks the latest 1.x version, and then it gets used in demos and some other places.

package.json
scottgonzalez: Ok, let's leave this change.

bower.json
mgol: Changed. I also changed it at https://github.com/jquery/jquery-ui/blob/fbc79e1bda2b58f980b9c6fedf3f8ddbce294d7b/tests/lib/bootstrap.js#L99

mgol:
PR updated.

scottgonzalez:
This looks good. Is this ready to be merged or do you still need to make changes in Jenkins first?

mgol:
> do you still need to make changes in Jenkins first

I've just created the Jenkins jobs for 1.12, 2.0, 2.1 & 2.2. I also added another commit to this PR that makes only Core 1.x jobs to be tested on IE 8, PTAL.

If this looks fine, feel free to merge. I'll then check if I configured the Jenkins jobs correctly.

scottgonzalez:
Merged.

mgol:
I've triggered the jobs manually and they seem to be running correctly:

* 2.2: http://swarm.jquery.org/job/2916
* 2.1: http://swarm.jquery.org/job/2917
* 2.0: http://swarm.jquery.org/job/2918
* 1.12: http://swarm.jquery.org/job/2919

A couple of tests are consistently failing with 2.0.0-2.0.3 in all browsers.

mgol:
OK, Core 2.0.2 is completely unusable on TestSwarm, it drains the available IE pool & needs to be removed as 1.10.1 has for the same reason. See PR #1707 for more details.

PR #1707, opened 2016-05-26
Tests: Stop testing against core 2.0.2 on testswarm

mgol:
jQuery 2.0.3 fixed very few things from 2.0.2, the list is here:
​http://blog.jquery.com/2013/07/03/jquery-1-10-2-and-2-0-3-released/

One of the fixes was ​http://bugs.jquery.com/ticket/13980, though which was
about cross-domain iframe issues. The problem is TestSwarm loads a cross-domain
iframe: the main page is swarm.jquery.org, the frame is on
builds.jenkins.jquery.com so it might be causing issues. You can see jQuery UI
tests are timing out in all IE versions in jQuery 2.0.2:
​http://swarm.jquery.org/job/2918

The problem is that it doesn't just fail, it starves the available IE pool,
making it sometimes harder for other projects to get their tests run on IE.

That's why tests with jQuery 2.0.2 on TestSwarm need to be removed as it's been
done with 1.10.1.

Refs #12745
Refs c2224bf5dc418c84c185844611786b9ccfb869a7
Refs gh-1706


mgol:
An example problematic run: http://swarm.jquery.org/job/2918.

PR #1708, opened 2016-05-27
Slider: parse options.step to int

parzychu:
Slider is used in [dateTimePicker from primefaces.org](http://www.primefaces.org/primeng/#/calendar). Changing stepMinute or stepHour passes  _number like string_  (e.g. "5") to jQuery slider option.step and it causes error.
I don't know how Calendar passes step to slider, but I think it's better to make sure slider option can handle _number like string_ in the future.


scottgonzalez:
Sorry, but we don't do this in jQuery UI. Valid values should be supplied.

PR #1712, opened 2016-06-01
Checkboxradio: Use new ui-state-checked class in checkboxradio

arschmitz:
Using ui-state-highlight caused a conflict with dialog

Fixes #14955

### DIFF
--- a/ui/widgets/checkboxradio.js
+++ b/ui/widgets/checkboxradio.js
@@ -177,8 +177,8 @@ $.widget( "ui.checkboxradio", [ $.ui.formResetMixin, {
 
 		if ( this.options.icon && this.type === "checkbox" ) {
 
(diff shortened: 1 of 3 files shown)


ui/widgets/checkboxradio.js
```diff
@@ -177,8 +177,8 @@ $.widget( "ui.checkboxradio", [ $.ui.formResetMixin, {
 
 		if ( this.options.icon && this.type === "checkbox" ) {
 
-			// We add ui-state-highlight to change the icon color
-			this._toggleClass( this.icon, null, "ui-icon-check ui-state-highlight", checked )
+			// We add ui-state-checked to change the icon color based on ui-state-highlight colors
```
scottgonzalez: I'd drop the reference to `ui-state-highlight` since that's a ThemeRoller-specific thing.

PR #1713, opened 2016-06-14
Controlgroup: Correctly handle non-empty child class key

gabrielschulhof:


tests/unit/controlgroup/non-empty-class-key.js
```diff
@@ -0,0 +1,51 @@
...
+	group.refresh();
+	assert.deepEqual(
+		( ( ( select.options.classes[ "ui-selectmenu-button-closed" ] || "" )
+			.match( /\S+/g ) ) || [] ).sort(),
+		[ "something-custom", "ui-corner-right" ],
+		"Adding a second item to a controlgroup results in the expected value for the class key" );
+
+} );
+
+} );
```
arschmitz: This should not be its own file

tests/unit/controlgroup/non-empty-class-key.js
```diff
@@ -0,0 +1,51 @@
...
+
+QUnit.test( "Controlgroup handles non-empty class key correctly", function( assert ) {
+	expect( 2 );
+	var select, resultingClasses;
+	var group = $( ".controlgroup-non-empty-class-key" )
+		.controlgroup()
+		.controlgroup( "instance" );
+
+	select = $( "<select><option>Option 1</option><option>Option 2</option></select>" )
```
arschmitz: Why are you constructing your html as a string in JS this is inconsistent with other tests in general

tests/unit/controlgroup/non-empty-class-key.js
```diff
@@ -0,0 +1,51 @@
...
+		.appendTo( group.element )
+		.selectmenu()
+		.selectmenu( "instance" );
```
arschmitz: if this was in your html you wouldn't have to go through all this

tests/unit/controlgroup/non-empty-class-key.js
```diff
@@ -0,0 +1,51 @@
...
+	group.refresh();
```
arschmitz: or this

tests/unit/controlgroup/non-empty-class-key.js
```diff
@@ -0,0 +1,51 @@
...
+	assert.deepEqual(
+		( ( ( select.options.classes[ "ui-selectmenu-button-closed" ] || "" )
+			.match( /\S+/g ) ) || [] ).sort(),
+		[ "something-custom", "ui-corner-all" ],
+		"Adding one item to a controlgroup results in the expected value for the class key" );
+
+	select = $( "<select><option>Option 1</option><option>Option 2</option></select>" )
```
arschmitz: same here as above

ui/widgets/controlgroup.js
```diff
@@ -117,15 +117,17 @@ return $.widget( "ui.controlgroup", {
...
 					if ( widget === "button" && element.parent( ".ui-spinner" ).length ) {
 						return;
 					}
 					if ( instance ) {
-						options.classes = that._resolveClassesValues( options.classes, instance );
+						instanceOptions.classes =
+							that._resolveClassesValues( instanceOptions.classes, instance );
 					}
-					element[ widget ]( options );
+					element[ widget ]( instanceOptions );
```
arschmitz: I can't be sure what the problem you're trying to fix is since you didn't include any description but im guessing its that when there are multiple of the same widget the options variable is over-written and re-used. if so i think we should fix this within the `_resolveClassesValues` function.

tests/unit/controlgroup/non-empty-class-key.js
```diff
@@ -0,0 +1,51 @@
...
+		.appendTo( group.element )
+		.selectmenu()
+		.selectmenu( "instance" );
+	group.refresh();
+	assert.deepEqual(
+		( ( ( select.options.classes[ "ui-selectmenu-button-closed" ] || "" )
+			.match( /\S+/g ) ) || [] ).sort(),
+		[ "something-custom", "ui-corner-all" ],
+		"Adding one item to a controlgroup results in the expected value for the class key" );
```
arschmitz: Why aren't you just doing `hasClasses( select, "something-custom ui-corner-all" );` ?

arschmitz:
I don't see where you have addressed the vast majority of my comments and I see more changes with absolutely no explanation.

arschmitz:
Also failing jshint

gabrielschulhof:
@arschmitz good that you told me to add the HTML into the markup! It exposed another problem: If you let the controlgroup instantiate the children, it always instantiates them with options, including classes options, but which are calculated without taking into account the default classes options stored in the prototype, because there is no instance on which to base those options. So, the prototype has to be the instance in that case, but the prototype cannot be gotten from the plugin. So, it has to be tacked onto the plugin in ```$.widget.bridge```.

tests/unit/controlgroup/core.js
```diff
@@ -160,4 +160,39 @@ QUnit.test( "Single controlgroup button - vertical", function( assert ) {
...
+} );
+
+QUnit.test( "Controlgroup handles non-empty class key correctly", function( assert ) {
+	assert.expect( 2 );
+
+	$( ".controlgroup-non-empty-class-key" )
+		.controlgroup()
+		.controlgroup( "instance" );
+
+	assert.deepEqual(
```
arschmitz: Again why are you not just using `hasClasses` to ensure the element has the classes we want

gabrielschulhof: Because it's not the classes on the element that are wrong, but the value of the class key that's wrong.
```.toggleClass()``` de-dupes when it applies classes, but the class key can look like "something-custom something-custom something-custom something-custom ui-corner-left".

Comment (gabrielschulhof) [inline tests/unit/controlgroup/core.js]: So, this is about the JS side, not the resulting markup.

**Comment by gabrielschulhof** [comment]
I'll add more comments.

File: ui/widgets/controlgroup.js [hunk 9]
```diff
@@ -117,15 +117,19 @@ return $.widget( "ui.controlgroup", {
...
+							$.fn[ widget ].widgetConstructor.prototype );
+					var instanceOptions = $.widget.extend( {}, options );
 
 					// If the button is the child of a spinner ignore it
 					if ( widget === "button" && element.parent( ".ui-spinner" ).length ) {
 						return;
 					}
-					if ( instance ) {
-						options.classes = that._resolveClassesValues( options.classes, instance );
+					if ( defaultInstance ) {
```
arschmitz: Instead of all this why not just
```
if ( !instance ) {
   element[ widget ]();
}
```

scottgonzalez:
Please file an issue about whatever is being fixed.

tests/unit/controlgroup/core.js
arschmitz: ok and why do we care if a user does that what does this break?

ui/widgets/controlgroup.js
gabrielschulhof: OK, instantiate with default options first, and then call options. I can do that.

ui/widgets/controlgroup.js
gabrielschulhof: Yes, but it's more than that. Even if I were to fix ```_resolveClassesValues()``` to not modify the value that is being passed in, you'd still need ```$.widget.extend()```, because the new object returned by ```_resolveClassesValues()``` would be stored inside ```options```, and ```options``` would be reused at the next iteration. So, it is ```options``` as well as ```options.classes``` that need to be cloned at every iteration.

tests/unit/controlgroup/core.js
gabrielschulhof: The user doesn't do that. Controlgroup does by the way it computes a value for the classes option.

gabrielschulhof: So, the user innocently sets ```$.ui.selectmenu.prototype.options.classes[ "ui-selectmenu-button-closed" ] to, say, "ui-shadow"``` and we proceed to mangle that value in every instance of a selectmenu that resides inside a controlgroup. Oh, and it's worse, because the length of the string increases with each ```refresh()```.

arschmitz: I still dont think we need to test for that when the bug only existed as part of a greater bug that would have also failed with just a `hasClasses` check.

Alternatively if you really think this is important for you to check make it a separate test and just use `matches` on `instance.options.classes[ 'whatever-key' ]` and check the length of the class you added it should never be more then 1.

gabrielschulhof: Yeah, OK, I see what you're saying. I do believe that we should test both the class value with ```.hasClasses()``` and the class key value.

tests/unit/controlgroup/core.js
```diff
@@ -160,4 +160,42 @@ QUnit.test( "Single controlgroup button - vertical", function( assert ) {
...
+			.selectmenu( "option", "classes" )[ "ui-selectmenu-button-closed" ]
+			.match( /something-custom/ ) || [] ).length, 1,
+		"Class 'something-custom' appears exactly once in the first widget's class key value" );
+
+	assert.strictEqual(
+		( $( "#non-empty-class-key-value-no-dupe-second" )
+			.selectmenu( "option", "classes" )[ "ui-selectmenu-button-closed" ]
+			.match( /something-custom/ ) || [] ).length, 1,
+		"Class 'something-custom' appears exactly once in the second widget's class key value" );
+} );
```
arschmitz: can we shorten all the class and id's here and in the html kind of excessively long

gabrielschulhof: Sure can :)

tests/unit/controlgroup/core.js
```diff
@@ -160,4 +160,42 @@ QUnit.test( "Single controlgroup button - vertical", function( assert ) {
 		" ui-corner-tr ui-corner-tl ui-corner-bl ui corner-br" );
 } );
 
+QUnit.module( "Controlgroup: Non-empty class key", {
+	setup: function() {
+		this.classKey = $.ui.selectmenu.prototype.options.classes[ "ui-selectmenu-button-closed" ];
+		$.ui.selectmenu.prototype.options.classes[ "ui-selectmenu-button-closed" ] =
```
arschmitz: why dont we just hard code a value here to simplify it really does not matter what we set here just that we set something non default

gabrielschulhof: Done.

ui/widgets/controlgroup.js
```diff
@@ -118,14 +118,24 @@ return $.widget( "ui.controlgroup", {
 					var element = $( this );
 					var instance = element[ widget ]( "instance" );
 
+					// We need to clone the default options for this type of widget to avoid
+					// polluting the variable options which has a wider scope than a single widget.
+					var instanceOptions = $.widget.extend( {}, options );
+
 					// If the button is the child of a spinner ignore it
 					if ( widget === "button" && element.parent( ".ui-spinner" ).length ) {
```
scottgonzalez: I know you're not touching this block of code, but since you're modifying a bunch of code around it, can you just add a comment here that says "TODO: Find a more generic solution"? This is something @arschmitz and I discussed before, but we don't have any ideas for how to implement it.

tests/unit/controlgroup/core.js
```diff
@@ -160,4 +160,41 @@ QUnit.test( "Single controlgroup button - vertical", function( assert ) {
...
+} );
+
+QUnit.test( "Controlgroup correctly assigns child widget classes options key", function( assert ) {
+	assert.expect( 2 );
+
+	$( ".controlgroup-class-key-dupe" ).controlgroup();
+
+	assert.strictEqual(
+		( $( "#class-key-dupe-first" )
+			.selectmenu( "option", "classes" )[ "ui-selectmenu-button-closed" ]
```
scottgonzalez: Please use `"classes.ui-selectmenu-button-closed"` instead of indexing into the full `classes` object.

tests/unit/controlgroup/core.js
```diff
@@ -160,4 +160,41 @@ QUnit.test( "Single controlgroup button - vertical", function( assert ) {
...
+			.match( /something-custom/g ) || [] ).length, 1,
+		"Class 'something-custom' appears exactly once in the first widget's class key value" );
+
+	assert.strictEqual(
+		( $( "#class-key-dupe-second" )
+			.selectmenu( "option", "classes" )[ "ui-selectmenu-button-closed" ]
```
scottgonzalez: same here

gabrielschulhof:
@scottgonzalez I have addressed your review comments.

arschmitz:
thank you I will merge this shortly

PR #1714, opened 2016-06-30
[WIP] Draggable: Remove handle class when disabled to allow scrolling

scottgonzalez:
Fixes #14779

I'd like to add a test for this before merging which checks if the `touch-action` property is set. I don't have time to do the necessary testing against IE right now though. We also need to handle sortable and selectable, but I wanted to get this started so I don't forget about it.

### DIFF
--- a/ui/widgets/draggable.js
+++ b/ui/widgets/draggable.js
@@ -78,7 +78,11 @@ $.widget( "ui.draggable", $.ui.mouse, {
 		if ( this.options.addClasses ) {
 			this._addClass( "ui-draggable" );
 		}
-		this._setHandleClassName();
+
+		this.handleElement = $();
+		if ( !this.options.disabled ) {
+			this._setHandleClassName();
+		}
 
 		this._mouseInit();
 	},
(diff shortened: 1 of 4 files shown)


scottgonzalez:
Updated the implementation and added tests. If this approach looks good, I'll update the other widgets.

ui/widgets/draggable.js
```diff
@@ -91,6 +94,15 @@ $.widget( "ui.draggable", $.ui.mouse, {
 		}
 	},
 
+	_setOptionDisabled: function( value ) {
+		// Remove touch-action for disabled draggables
```
arschmitz: blank line before comments

arschmitz: But implementation is good

BenasPaulikas:
Is this related to https://forum.jquery.com/topic/jquery-ui-sortable-disables-scroll-on-android ?

When this will be fixed ?

aldug:
I also have been impacted by this behavior.  Is anyone currently working on finishing this?

PR #1715, opened 2016-06-30
jquery-ui:Base text direction support

ashensis:
As per discussion in https://bugs.jqueryui.com/ticket/14477, current code submit contains Base text direction support for UI widgets.
New option is introduced for relevant widgets 'textDir' with values:
'ltr' - Left-to-right text direction
'rtl' - Right-to-left text direction
'auto' - Contextual text direction that gets resolved to either 'ltr' or 'rtl' according to first strong character ('rtl' if it is Arabic/Hebrew character, 'ltr' otherwise).
Text direction option shouldn't affect the mirroring aspects of widgets having assimmetric GUI, it only impacts the text direction of contained text.
The corresponding demo textDir.html illustrates the issue for all relevant widgets.
When current approach will be approved (or modified according reviewer remaarks) documentation and unitests will be provided.

### DIFF
--- a/tests/unit/tooltip/common.js
+++ b/tests/unit/tooltip/common.js
@@ -18,6 +18,7 @@ common.testWidget( "tooltip", {
 			collision: "flipfit flip"
 		},
 		show: true,
+		textDir: null,
 		track: false,
 
 		// Callbacks
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -697,6 +697,32 @@ $.Widget.prototype = {
 		return !( $.isFunction( callback ) &&
 			callback.apply( this.element[ 0 ], [ event ].concat( data ) ) === false ||
 			event.isDefaultPrevented() );
+	},
+
+	_getTextDir: function( text ) {
+		if ( this.options.textDir === "auto" ) {
+
+			// Look for first strong (either English or Arabic/Hebrew) character
+			// Resolve text direction accordingly ("rtl" for Arabic/Hebrew, "ltr" otherwise)
+			var matcher = /[A-Za-z\u05d0-\u065f\u066a-\u06ef\u06fa-\u07ff\ufb1d-\ufdff\ufe70-\ufefc]/.exec( text );
+			return ( matcher && ( matcher[ 0 ] > "z" ) ) ? "rtl" : "ltr";
+		}
+		return this.options.textDir;
+	},
+
+	_applyTextDir: function( param ) {
+		if ( typeof param === "string" ) {
+			param = param.replace( /[\u202A\u202B\u202C]/g, "" );
+
+			// Unicode directional characters: 202A and 202B used to enforce text direction
+			// 202C - POP formatter closing directional segment
+			return ( this._getTextDir( param ) === "rtl" ? "\u202B" : "\u202A" ) + param + "\u202C";
+		} else if ( param.jquery ) {
+				var isField = param.is( "input" ) || param.is( "textarea" );
+				param.css( "direction", this._getTextDir( isField ? param.val() : param.text() ) );
+		} else if ( param.nodeType === 1 ) {
+				param.style.direction = this._getTextDir( param.textContent );
+		}
 	}
 };
 
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -702,8 +702,8 @@ $.Widget.prototype = {
 	_getTextDir: function( text ) {
 		if ( this.options.textDir === "auto" ) {
 
-			// Look for first strong (either English or Arabic/Hebrew) character
-			// Resolve text direction accordingly ("rtl" for Arabic/Hebrew, "ltr" otherwise)
+			// Look for first strong (either English or Arabic/Hebrew) character.
+			// Resolve text direction accordingly ("rtl" for Arabic/Hebrew, "ltr" otherwise).
 			var matcher = /[A-Za-z\u05d0-\u065f\u066a-\u06ef\u06fa-\u07ff\ufb1d-\ufdff\ufe70-\ufefc]/.exec( text );
 			return ( matcher && ( matcher[ 0 ] > "z" ) ) ? "rtl" : "ltr";
 		}
@@ -714,8 +714,8 @@ $.Widget.prototype = {
 		if ( typeof param === "string" ) {
 			param = param.replace( /[\u202A\u202B\u202C]/g, "" );
 
-			// Unicode directional characters: 202A and 202B used to enforce text direction
-			// 202C - POP formatter closing directional segment
+			// Unicode directional characters: 202A and 202B used to enforce text direction.
+			// 202C - POP formatter closing directional segment.
 			return ( this._getTextDir( param ) === "rtl" ? "\u202B" : "\u202A" ) + param + "\u202C";
 		} else if ( param.jquery ) {
 				var isField = param.is( "input" ) || param.is( "textarea" );
--- a/ui/widgets/accordion.js
+++ b/ui/widgets/accordion.js
@@ -54,6 +54,7 @@ return $.widget( "ui.accordion", {
 			activeHeader: "ui-icon-triangle-1-s",
 			header: "ui-icon-triangle-1-e"
 		},
+		textDir: null,
 
 		// Callbacks
 		activate: null,
@@ -304,6 +305,13 @@ return $.widget( "ui.accordion", {
 			._removeClass( this.active, "ui-accordion-header-collapsed" );
 		this._addClass( this.active.next(), "ui-accordion-content-active" );
 		this.active.next().show();
+
+		if ( this.options.textDir ) {
+			var that = this;
+			this.headers.each( function( i, header ) {
+				header.textContent = that._applyTextDir( header.textContent );
+			} );
+		}
 
 		this.headers
 			.attr( "role", "tab" )
--- a/ui/widgets/autocomplete.js
+++ b/ui/widgets/autocomplete.js
@@ -50,6 +50,7 @@ $.widget( "ui.autocomplete", {
 			collision: "none"
 		},
 		source: null,
+		textDir: null,
 
 		// Callbacks
 		change: null,
@@ -89,6 +90,26 @@ $.widget( "ui.autocomplete", {
 
 		this._addClass( "ui-autocomplete-input" );
 		this.element.attr( "autocomplete", "off" );
+		if ( this.options.textDir ) {
+			var textDir = this._getTextDir( this._value() );
+			this.element.css( "direction", textDir );
+			if ( this.options.textDir === "auto" ) {
+				this.element.css( "text-align", textDir === "rtl" ? "right" : "left" );
+				this._on( this.element, {
+					keyup: function( e ) {
+						var keyCode = $.ui.keyCode;
+						if ( e.keyCode < keyCode.PAGE_UP || e.keyCode >= keyCode.DELETE ) {
+							var textDir = this._getTextDir( this._value() );
+							this.element.css( "direction", textDir )
+								.css( "text-align", textDir === "rtl" ? "right" : "left" );
+						}
+					}
+				} );
+			} else {
+				this.element.css( "text-align",
+					this.element.css( "direction" ) === "rtl" ? "right" : "left" );
+			}
+		}
 
 		this._on( this.element, {
 			keydown: function( event ) {
@@ -288,6 +309,9 @@ $.widget( "ui.autocomplete", {
 
 				if ( false !== this._trigger( "select", event, { item: item } ) ) {
 					this._value( item.value );
+					if ( this.options.textDir === "auto" ) {
+						this.element.css( "direction", this._getTextDir( item.value ) );
+					}
 				}
 
 				// reset the term after the select event
@@ -521,6 +545,15 @@ $.widget( "ui.autocomplete", {
 	_suggest: function( items ) {
 		var ul = this.menu.element.empty();
 		this._renderMenu( ul, items );
+
+		if ( this.options.textDir ) {
+			ul.css( "text-align", ul.css( "direction" ) === "rtl" ? "right" : "left" );
+			var that = this;
+			ul.children().each( function( i, li ) {
+				li.style.direction = that._getTextDir( li.textContent );
+			} );
+		}
+
 		this.isNewMenu = true;
 		this.menu.refresh();
 
--- a/ui/widgets/button.js
+++ b/ui/widgets/button.js
@@ -49,7 +49,8 @@ $.widget( "ui.button", {
 		icon: null,
 		iconPosition: "beginning",
 		label: null,
-		showLabel: true
+		showLabel: true,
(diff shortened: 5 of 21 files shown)


scottgonzalez:
Thanks. See also #1682.

Just FYI: We won't have any time to discuss the approach until after our releases are done, which is why #1682 has been sitting for a long time.

tomerm:
No problem. Will be waiting for your feedback / comments. 

One clarification with your kind permission: 

#1682 is about fixing issues with GUI mirroring (relevant for several widgets). GUI mirroring is strongly associated with translation of UI to Arabic / Hebrew. At the same time, GUI mirroring is mostly concerned with graphical layout of screen and widget. 

As opposed to that, this specific PR is concerned with text only (which can appear as part of widget). Thus this PR is completely orthogonal to GUI direction and #1682 .

Just FYI. My team and our colleagues from Egypt Bidi center (who authored #1682) work together to address variety of Bidi requirements in open source tools / libraries.

arschmitz:
I have not reviewed this at all yet simply read the discription. Just one minor thing i noticed from that is any time we have an "auto" option like this that determines the value used based on DOM etc, we always use the special value `null` to denote it. so in this case the 3 values should be `"ltr"`, `"rtl"`, and `null`. `null` should be the default value. When the value `null` is used and `rtl` or `ltr` chosen based on markup during init, the option value should be set as well. In this way when a developer checks the value of the option it should always return `rtl`  or `ltr` and never `null`

tomerm:
@scottgonzalez may I kindly inquire which release are we waiting for to start the review ? I believe 3.2 was out in March 2017.

scottgonzalez:
Let's focus on a single widget to have our discussions. I'll pick button since it's the first widget in the `textDir.html` demo and is fairly simple. Please explain exactly what the button widget is accounting for with text direction that isn't already handled natively by `dir="rtl"`.

tomerm:
@scottgonzalez button is not the best example to illustrate the benefit of textDir parameter. This is mostly since button has horizontally symmetrical graphics (just a rectangular area without any shade) and thus it looks (from graphical perspective) exactly the same with both LTR and RTL directions. However even for the button there is a minor benefit. Please observe that textDir is meant to affect text only. Currently if you specify dir="ltr" or dir="rtl" for button you don't cover auto direction (aka contextual or first stong) for text label displayed as part of the button. You can specify dir="auto" for button but this won't work at least not on IE. Thus having textDir even in the case of button has some benefit. 

In general textDir has striking benefit for widgets which are horizontally asymmetrical. For example, combo box, list box (with scroll bar), check box, tree, grid etc.  When you want to flip horizontally such widgets you usually specify dir="ltr" or dir="rtl". However, this by itself does not allow you to enforce text direction for text displayed in such widgets independently from GUI direction. For example if you specify dir="rtl" for combo box, all text inside such combo box will be displayed with RTL direction. However if you wish this text to be displayed with LTR direction you have no such option. Having textDir parameter allows you to control text direction of text inside a widget independently from direction of widget graphics. 

For the sake of example, please observe the behavior of native controls on iOS / Android platforms. Text in such controls is displayed by default with textDir = "auto" regardless of direction of control's graphics (dir ="ltr" or dir ="rtl"). By the way majority of desktop technologies (including native mobile libraries) allow to control direction of text independently from widget (or GUI) direction.  For example: 

- Direction of GUI (or widget) is called by **Android** _layoutDirection_ (https://developer.android.com/reference/android/view/View.html#setLayoutDirection%28int%29) . It is equivalent to HTML dir specified for a widget.
- Text direction (for text displayed inside a widget) is called by **Android** _textDirection_ (https://developer.android.com/reference/android/view/View.html#setTextDirection%28int%29). It is equivalent to HTML dir specified on the text inline element itself.

scottgonzalez:
Thanks for the explanation on the two different direction settings. Now that I have a better understanding of the specific goal trying to be accomplished, I will take another look at the changes.

demos/widget/textDir.html
```diff
@@ -0,0 +1,283 @@
+<!doctype html>
```
scottgonzalez: This demo will need to be removed since it's not actually a demo of the widget factory.

tests/unit/tooltip/common.js
```diff
@@ -19,6 +19,7 @@ common.testWidget( "tooltip", {
 		},
 		show: true,
 		track: false,
+		textDir: null,
```
scottgonzalez: These options are alphabetized. All files will need to be updated for the correct order.

ui/widget.js
```diff
@@ -280,6 +280,8 @@ $.Widget.prototype = {
 		classes: {},
 		disabled: false,
 
+//		textDir: null,
```
scottgonzalez: Removed commented out code.

ui/widget.js
```diff
@@ -664,6 +666,27 @@ $.Widget.prototype = {
 		return !( $.isFunction( callback ) &&
 			callback.apply( this.element[ 0 ], [ event ].concat( data ) ) === false ||
 			event.isDefaultPrevented() );
+	},
+
+	_getTextDir: function( text ) {
+		if ( this.options.textDir === "auto" ) {
+			var matcher = /[A-Za-z\u05d0-\u065f\u066a-\u06ef\u06fa-\u07ff\ufb1d-\ufdff\ufe70-\ufefc]/.exec( text );
```
scottgonzalez: Please add a comment about which set of characters this is searching for.

ui/widget.js
```diff
@@ -664,6 +666,27 @@ $.Widget.prototype = {
...
+			return ( matcher && ( matcher[ 0 ] > "z" ) ) ? "rtl" : "ltr";
+		}
+		return this.options.textDir;
+	},
+	_applyTextDir: function( param ) {
```
scottgonzalez: Blank line between methods.

ui/widget.js
```diff
@@ -664,6 +666,27 @@ $.Widget.prototype = {
...
+		if ( typeof param === "string" ) {
+			var dir = this._getTextDir( param );
+			param = param.replace( /[\u202A\u202B\u202C]/g, "" );
```
scottgonzalez: Please add a comment about what these characters are.

ui/widget.js
```diff
@@ -664,6 +666,27 @@ $.Widget.prototype = {
...
+			return ( dir === "rtl" ? "\u202B" : "\u202A" ) + param + "\u202C";
+		} else {
+			if ( param.jquery ) {
+				param.css( "direction", this._getTextDir( param.text() ) );
+			} else if ( param.style ) {
```
scottgonzalez: Use `param.nodeType` to check for elements.

ui/widget.js
```diff
@@ -664,6 +666,27 @@ $.Widget.prototype = {
...
+			return ( matcher && ( matcher[ 0 ] > "z" ) ) ? "rtl" : "ltr";
+		}
+		return this.options.textDir;
+	},
+	_applyTextDir: function( param ) {
+		if ( typeof param === "string" ) {
+			var dir = this._getTextDir( param );
+			param = param.replace( /[\u202A\u202B\u202C]/g, "" );
+			return ( dir === "rtl" ? "\u202B" : "\u202A" ) + param + "\u202C";
+		} else {
```
scottgonzalez: No need to nest conditionals, just use an `else if` here.

ui/widgets/accordion.js
```diff
@@ -89,6 +90,11 @@ return $.widget( "ui.accordion", {
 		}
 
 		this._processPanels();
+		if ( this.options.textDir ) {
```
scottgonzalez: This needs to move to `_refresh()` to handle updates.

ui/widgets/accordion.js
```diff
@@ -89,6 +90,11 @@ return $.widget( "ui.accordion", {
...
+			var that = this;
+			this.headers.each( function( i, header ) {
+				header.textContent = that._applyTextDir( header.textContent );			} );
```
scottgonzalez: Looks like a line break got deleted.

ui/widgets/autocomplete.js
```diff
@@ -89,6 +90,20 @@ $.widget( "ui.autocomplete", {
...
 		this._addClass( "ui-autocomplete-input" );
 		this.element.attr( "autocomplete", "off" );
+		if ( this.options.textDir ) {
+			this.element.css( "text-align",
+				this.element.css( "direction" ) === "rtl" ? "right" : "left" );
+
+			this.element.css( "direction", this._getTextDir( this.element.text() ) );
+			if ( this.options.textDir === "auto" ) {
+				var that = this;
+				this.element.bind( "keyup", function() {
```
scottgonzalez: Use [`._on()`](http://api.jqueryui.com/jQuery.widget/#method-_on).

ui/widgets/autocomplete.js
```diff
@@ -89,6 +90,20 @@ $.widget( "ui.autocomplete", {
...
+					var fld = $( this );
+					fld.css( "direction",
+						that._getTextDir( that.valueMethod.apply( fld ) ) );
```
scottgonzalez: This seems fairly expensive to do on every keystroke, but perhaps it's not.

ui/widgets/autocomplete.js
```diff
@@ -89,6 +90,20 @@ $.widget( "ui.autocomplete", {
...
 		this.element.attr( "autocomplete", "off" );
+		if ( this.options.textDir ) {
+			this.element.css( "text-align",
+				this.element.css( "direction" ) === "rtl" ? "right" : "left" );
+
+			this.element.css( "direction", this._getTextDir( this.element.text() ) );
+			if ( this.options.textDir === "auto" ) {
+				var that = this;
+				this.element.bind( "keyup", function() {
+					var fld = $( this );
```
scottgonzalez: s/fld/element/

But that's not needed once you switch to `._on()` anyway.

ui/widgets/button.js
```diff
@@ -88,6 +89,10 @@ $.widget( "ui.button", {
 
 		this.hasTitle = !!this.element.attr( "title" );
 
+		if ( this.options.textDir ) {
+			this.options.label = this._applyTextDir( this.options.label );
```
scottgonzalez: Doesn't this break HTML inside the button?

ui/widgets/button.js
```diff
@@ -242,6 +248,10 @@ $.widget( "ui.button", {
 				this._updateTooltip();
 		}
 
+		if ( this.options.textDir && ( key === "label" || key === "textDir" ) ) {
```
scottgonzalez: What about handling when `textDir` changes? This seems like a place where we probably need to leverage `_setOptions()` instead of `_setOption()`.

ui/widgets/controlgroup.js
```diff
@@ -92,8 +93,11 @@ return $.widget( "ui.controlgroup", {
 					if ( element.children( ".ui-controlgroup-label-contents" ).length ) {
 						return;
 					}
-					element.contents()
-						.wrapAll( "<span class='ui-controlgroup-label-contents'></span>" );
+					var dir = that.options.textDir ?
+						"dir='" + that._getTextDir( element.text() ) + "' " : "";
+
+					element.contents().wrapAll
```
scottgonzalez: Don't break between method names and their invoking parens.

ui/widgets/tooltip.js
```diff
@@ -265,6 +266,9 @@ $.widget( "ui.tooltip", {
 		tooltipData = this._find( target );
 		if ( tooltipData ) {
 			tooltipData.tooltip.find( ".ui-tooltip-content" ).html( content );
+			if ( this.options.textDir ) {
+				this._applyTextDir( tooltipData.tooltip.find( ".ui-tooltip-content" ) );
```
scottgonzalez: Doesn't this break HTML inside the tooltip?

scottgonzalez:
@ashensis You can just push updates to this branch and keep the PR open.

adir01:
@scottgonzalez Can you please reopen this PR ?

demos/widget/textDir.html
ashensis: Done.

tests/unit/tooltip/common.js
ashensis: Done.

ui/widget.js
ashensis: Done.

ui/widget.js
ashensis: Done.

ui/widget.js
ashensis: Done.

ui/widget.js
ashensis: Done.

ui/widget.js
ashensis: Done.

ui/widget.js
ashensis: Done, 
Added condition telling between use of text() and val()

ui/widgets/accordion.js
ashensis: Done.

ui/widgets/accordion.js
ashensis: Fixed

ui/widgets/autocomplete.js
ashensis: Done.

ui/widgets/autocomplete.js
ashensis: Indeed, changed

ui/widgets/button.js
ashensis: As a matter of fact inserted UCC constitutes just text node in HTML markup and doesn't break it, while having desirable impact.
However I made changes accounting to possibility of HTML content of button widget and now the element 'direction' style gets changed instead of insertion of directional UCC.

ui/widgets/button.js
ashensis: Moved code to '_setOptions'

ui/widgets/controlgroup.js
ashensis: Fixed

ui/widgets/autocomplete.js
ashensis: I filter now the majority of irrelevant key strokes that have no imact on field content

ui/widgets/tooltip.js
ashensis: By no means. In this case the tooltip.find returns JQuery object and '_applyTextDir' sets its 'direction' style

ashensis:
@scottgonzalez  All comments in initial round of review had been addressed, please have a look

adir01:
@scottgonzalez Are you ok with @sashash changes ? Did he address all your comments ? Is anything else needed to allow his code to be merged ?

adir01:
@scottgonzalez - thank you for review
@ashensis  - can you please address Scott comments ?

adir01:
@arschmitz and  @jzaefferer who is in charge to review this PR . @scottgonzales is not answering a long time

PR #1717, opened 2016-07-05
Droppable: Fix problem with droppable not being deactivated.

Jake91:
I have problem with an old bug: https://bugs.jqueryui.com/ticket/8046
This is because the value returned from my "accept" function is different depending on if the "drop" operation have been executed or not.

Since the code is just iterating the droppables once, it will perform the drop directly when it finds the correct droppable (take a look at the code).
If you ask me, that is a weird behavior. I have seperated this into several loops instead.
So that we:
1 - call the "accept" function for all droppables and store the result
2 - call the "drop" function for the droppable
3 - call "deactivate" for all droppables that were "accepted" before the "drop"

Fixes #8046

### DIFF
--- a/ui/widgets/droppable.js
+++ b/ui/widgets/droppable.js
@@ -355,38 +355,38 @@ $.ui.ddmanager = {
 	    var dropped = false;
 
 	    // Create a copy of the droppables in case the list changes during the drop (#9116)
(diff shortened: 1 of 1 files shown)


scottgonzalez:

PR #1718, opened 2016-07-06
Tooltip: Allow tracking tooltips to be off by a fraciton of a pixel

scottgonzalez:


jzaefferer:
Apart from the jshint failure, looks good.

PR #1719, opened 2016-07-11
Controlgroup: Handle child elements that don't have options defined

scottgonzalez:
This was causing failures with jQuery <1.12.0.


arschmitz:
👍

PR #1720, opened 2016-07-11
Checkboxradio: Fix label handling with jQuery 3.x

scottgonzalez:
This needs to have a ticket filed before I merge this, but I wanted to get a review.

This was caused by https://github.com/jquery/jquery/commit/0e2f8f9effd2c95647be534bf9055e099aad7cfd which @gibson042 views as a bug in core.


jzaefferer:
Looks fine to me

scottgonzalez:
And here's the bug for jQuery Core: https://github.com/jquery/jquery/issues/3226

PR #1722, opened 2016-07-15
Autocomplete: Add display block to ui-menu-item-wrapper CSS

jdufresne:
A custom _renderItem function may not return a block element directly
below the li. e.g. an anchor tag. Adding the CSS ensures proper display
and highlighting.

### DIFF
--- a/themes/base/menu.css
+++ b/themes/base/menu.css
@@ -25,6 +25,7 @@
 	list-style-image: url("data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7");
 }
 .ui-menu .ui-menu-item-wrapper {
+	display: block;
 	position: relative;
 	padding: 3px 1em 3px .4em;
 }


scottgonzalez:
I'm really hesitant to add a rule like this because of the implications on horizontal menus.

scottgonzalez:
Can you explain why you would use an anchor tag inside the autocomplete menu?

arschmitz:
I think using `_renderItem` to change the markup also means you are responsible for the css to make the custom item display how you want, which may not be block at all. For example you may want to display a grid of icons as the autocomplete options.

jdufresne:
> Can you explain why you would use an anchor tag inside the autocomplete menu?

I use the autocomplete widget to allow my users to search for objects in the database. The autocomplete lists available objects given the search term. Once a user sees the object they want, they will click the link to go to the object's page.

jdufresne:
> I think using _renderItem to change the markup also means you are responsible for the css to make the custom item display how you want, which may not be block at all. For example you may want to display a grid of icons as the autocomplete options.

Fair enough. This is my current solution.

I made this PR as I noticed something changed when upgrading from 1.11 to 1.12 causing the autocomplete items to render awkwardly. This was the easiest solution to make it work as before and so thought maybe an upstream fix would be appreciated and useful for others.

arschmitz:
@jdufresne i believe the change you mention is https://jqueryui.com/upgrade-guide/1.12/#require-wrappers-for-each-menu-item

jdufresne:
> That's what the select event is for. You shouldn't be using anchors in the menu.

Anchors allow my users to middle click (or Ctrl + click) to open the page in a new tab. I see this is a important browser feature that I prefer not to break.

jdufresne:
@arschmitz Thanks for then link. That is very useful.

jdufresne:
If the documentation states that the wrapper **must** be a block level element, why not enforce this through CSS for these use cases? Presumably horizontal menus must already work with block level wrappers if they're required.

arschmitz:
@jdufresne that requirement is a styling issue and therefor would only apply to default rendering and markup, not custom. If you supply custom markup there are going to be styling consequences which you must take into account.

PR #1724, opened 2016-07-25
Tabs: Don't allow navigating to hidden tabs

scottgonzalez:
Fixes #15013

### DIFF
--- a/ui/widgets/tabs.js
+++ b/ui/widgets/tabs.js
@@ -258,6 +258,7 @@ $.widget( "ui.tabs", {
 	},
 
 	_findNextTab: function( index, goingForward ) {
+		var that = this;
 		var lastTabIndex = this.tabs.length - 1;
 
 		function constrain() {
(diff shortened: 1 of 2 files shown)


ui/widgets/tabs.js
```diff
@@ -270,7 +271,19 @@ $.widget( "ui.tabs", {
 			return index;
 		}
 
-		while ( $.inArray( constrain(), this.options.disabled ) !== -1 ) {
+		function canActivate( index ) {
+			if ( $.inArray( constrain(), that.options.disabled ) !== -1 ) {
+				return false;
+			}
+
+			if ( that.tabs.eq( index ).is( ":hidden" ) ) {
```
jzaefferer: How about...
```
return that.tabs.eq( index ).is( ":visible" )
```

scottgonzalez: When there are multiple failure cases, I generally like to list them out individually as failure cases, then return `true` at the end, but I'm fine making this change.

jzaefferer:
Besides that one thing, 👍

ui/widgets/tabs.js
arschmitz: Just a thought but the `:visible` and `:hidden` selectors are pretty slow and not 100% reliable. We don't document that hidden tabs are automatically not navigable. Would it perhaps be better to leave this change out and document that you should mark hidden tabs as disabled?

scottgonzalez: Actually, for some reason switching it causes the tests to fail.

scottgonzalez: > We don't document that hidden tabs are automatically not navigable.

Why would anyone ever expect a hidden tab to be navigable?

arschmitz: I don't think they would which is why i suggested documenting that hidden tabs should be marked disabled

scottgonzalez: https://bugs.jqueryui.com/ticket/15013#comment:2

scottgonzalez:
I'm going to close this because this will result in an odd state of tabs not being marked as disabled, but also not being able to be activated properly. Instead, we should implement the hidden detection inside `refresh()` and set those tabs as disabled.

PR #1728, opened 2016-08-15
Tabs: Fix _isLocal

flyering:
When page includes base element, this attribute was incorrect.


scottgonzalez:
The base element affects all anchors, even if they contain a relative path with only a fragment identifier.

arschmitz:
We work around this in jQuery Mobile which uses a base element with an extension https://github.com/jquery/jquery-mobile/blob/master/js/widgets/tabs.ajax.js @flyering you can just include this extension ( you dont need the create method ) and it will fix the issue with the base element.

PR #1729, opened 2016-08-15
Create datepicker-arm.js

karenYerevan:
Armenian language fro jQuery Ui datepicker


scottgonzalez:
Thanks, but we're not accepting any new locales. We're in the process of switching over to Unicode's CLDR data as our data source.

PR #1731, opened 2016-08-17
Update bower.json with license information

johannesschaefer:
Add the license to the bower.json file. This is needed by some tools (e.g. http://www.webjars.org/bower).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -7,7 +7,7 @@
 		"external",
 		"tests"
 	],
-+       "license": "MIT",
+        "license": "MIT",
 	"dependencies": {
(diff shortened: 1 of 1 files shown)


mgol:
There's a parse error in this patch.

bower.json
```diff
@@ -7,7 +7,7 @@
 		"external",
 		"tests"
 	],
-+       "license": "MIT",
+        "license": "MIT",
```
mgol: This doesn't respect the code style of the whole file. A single tab should be used.

mgol:
Also, please sign our CLA as the bot tells you.

PR #1732, opened 2016-08-17
Menu: Support number pad keyboard input

fnagel:
Fixes #15031

### DIFF
--- a/tests/unit/selectmenu/core.js
+++ b/tests/unit/selectmenu/core.js
@@ -341,23 +341,39 @@ $.each( [
 	} );
 } );
 
-	QUnit.test( "Selectmenu should reset when its parent form resets", function( assert ) {
-		var ready = assert.async();
-		assert.expect( 2 );
+QUnit.test( "Selectmenu should reset when its parent form resets", function( assert ) {
+	var ready = assert.async();
+	assert.expect( 2 );
 
-		var element = $( "#speed" ).selectmenu(),
-			widget = element.selectmenu( "widget" ),
-			initialValue = element.val(),
-			form = element.closest( "form" );
+	var element = $( "#speed" ).selectmenu(),
+		widget = element.selectmenu( "widget" ),
+		initialValue = element.val(),
+		form = element.closest( "form" );
 
-		element.val( "Slower" );
-		element.selectmenu( "refresh" );
-		assert.equal( $.trim( widget.text() ), "Slower" );
-		form[ 0 ].reset();
-		setTimeout( function() {
-			assert.equal( $.trim( widget.text() ), initialValue );
-			ready();
-		} );
+	element.val( "Slower" );
+	element.selectmenu( "refresh" );
+	assert.equal( $.trim( widget.text() ), "Slower" );
+	form[ 0 ].reset();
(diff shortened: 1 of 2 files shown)


fnagel:
@arschmitz FYI

ui/widgets/menu.js
```diff
@@ -236,9 +236,12 @@ return $.widget( "ui.menu", {
 		default:
 			preventDefault = false;
 			prev = this.previousFilter || "";
-			character = String.fromCharCode( event.keyCode );
 			skip = false;
 
+			// Support number pad values
+			character = event.keyCode >= 96 && event.keyCode <= 105 ?
+				event.keyCode - 96 : String.fromCharCode( event.keyCode );
```
arschmitz: I don't think this really matters in this case but when `event.keyCode >= 96 && event.keyCode <= 105` we get a number here and when it's less we get a string. Do we want to force character to always be a string for consistency?

tests/unit/selectmenu/core.js
```diff
@@ -341,23 +341,40 @@ $.each( [
...
+	assert.equal( $.trim( widget.text() ), "Slower" );
+	form[ 0 ].reset();
+	setTimeout( function() {
+		assert.equal( $.trim( widget.text() ), initialValue );
+		ready();
 	} );
+} );
+
+// http://bugs.jqueryui.com/ticket/15031
+QUnit.test( "#15031: Number pad input should change value", function( assert ) {
```
arschmitz: We don't refrence bugs in tests like this in general any more i don't think

ui/widgets/menu.js
scottgonzalez: I would.

arschmitz: that was my thought

fnagel: So adding something like `( event.keyCode - 9 ).toString();` would be suitable?

scottgonzalez: Yup, that'd be fine.

tests/unit/selectmenu/core.js
fnagel: I've seen both (the prefix in the test title and the comment before the actual test) in the current code base. Which one are you talking about?

ui/widgets/menu.js
fnagel: Fixed.

tests/unit/selectmenu/core.js
arschmitz: Both actually while yes there are references in the code base like this currently this is something i'm pretty sure we don't do any more ( reference issues directly in tests ). 
@scottgonzalez

scottgonzalez: Correct. We only reference functionality now.

fnagel: Sooooo... I remove both ticket number references we're good to go?

scottgonzalez: Yup. The rest looks good :-)

fnagel: Fixed.

scottgonzalez:
Looks good.

PR #1734, opened 2016-08-22
Workarounds IE 10+ input trigger

bergi9:
https://bugs.jqueryui.com/ticket/9796

### DIFF
--- a/ui/widgets/autocomplete.js
+++ b/ui/widgets/autocomplete.js
@@ -188,7 +188,7 @@ $.widget( "ui.autocomplete", {
 				}
 			},
 			input: function( event ) {
-				if ( suppressInput ) {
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Surely this breaks legitimate input.

arschmitz:
This will ignore the first input event regardless of browser or timing. On the very first input event `supressInput` is always `undefined` and is `false` after that. So while this will fix IE10 it will ignore legit first input in other browsers or even in IE10 when there is no umlaut. For this to be generic enough to work we would have to detect if the initial value contained a string that causes issues for ie10 ( sounds like its anything outside the ascii range ) and then also feature detect if we are in IE10 and check all three of these conditions in the second clause of this if statement

I think we should close this since even if the 2 extra checks are added i don't think this is practical

PR #1735, opened 2016-08-30
Slider: Fix drag move at same value

JanGalek:
When we have the same value, for example $100 for hadles, then first can move down, and second up.

### DIFF
--- a/ui/widgets/slider.js
+++ b/ui/widgets/slider.js
@@ -76,6 +76,7 @@ return $.widget( "ui.slider", $.ui.mouse, {
 		this._detectOrientation();
 		this._mouseInit();
 		this._calculateNewMax();
+		this._originalVal;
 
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Thanks for the patch, but it's not clear exactly what behavior is being changed. Are you trying to fix a bug? If so, has it already been reported? See [the bug tracker](https://bugs.jqueryui.com/report/10?P=slider) to find out. If not, please file a bug first.

You can find more information about the contribution process in our [contributing guidelines](https://github.com/jquery/jquery-ui/blob/master/CONTRIBUTING.md). Let us know if you have any questions. Thanks!

scottgonzalez:
After a quick review of what this does, the behavior is not acceptable. This changes which handle is being moved mid-slide, which not only results in confusing behavior, but also results in a broken visual state.

PR #1736, opened 2016-08-30
Tabs: Strip hash from remote content URLs

scottgonzalez:
As of jQuery 3.0.0, hashes are no longer stripped for Ajax requests. This
causes issues in IE <11, so we need to strip this before making the request.

Ref jquery/jquery#1732

----

I found this from TestSwarm. Nobody has filed a bug yet, so I didn't bother doing that, but I'm not opposed to filing a bug if the team would prefer.

### DIFF
--- a/ui/widgets/tabs.js
+++ b/ui/widgets/tabs.js
@@ -889,7 +889,10 @@ $.widget( "ui.tabs", {
 	_ajaxSettings: function( anchor, event, eventData ) {
 		var that = this;
 		return {
-			url: anchor.attr( "href" ),
+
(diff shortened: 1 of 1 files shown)


jzaefferer:
👍 

PS: Should we keep the mention-bot? Seems to me like its just noise.

scottgonzalez:
> PS: Should we keep the mention-bot? Seems to me like its just noise.

Let's drop it. I don't think it's providing value.

jzaefferer:
Okay, removed the webhook.

arschmitz:
👍

PR #1737, opened 2016-08-30
Dialog: Allow for subpixel calculation errors in tests

scottgonzalez:
Tests were failing in IE 10-11 with values that were off by 0.01 pixels.

### DIFF
--- a/tests/unit/dialog/helper.js
+++ b/tests/unit/dialog/helper.js
@@ -33,11 +33,8 @@ return $.extend( helper, {
 		assert.ok( expectedDX - actualDX <= 1 && expectedDY - actualDY <= 1, "dragged[" + expectedDX + ", " + expectedDY + "] " + msg );
 	},
 
-	// TODO switch back to checking the size of the .ui-dialog element (var d)
-	// once we switch to using box-sizing: border-box (#9845) that should work fine
-	// using the element's dimensions to avoid subpixel errors
 	shouldResize: function( assert, element, dw, dh, msg ) {
-		var heightAfter, widthAfter, actual, expected,
+		var actualDH, actualDW, heightAfter, widthAfter
 			d = element.dialog( "widget" ),
 			handle = $( ".ui-resizable-se", d ),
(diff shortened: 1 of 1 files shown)


tests/unit/dialog/helper.js
```diff
@@ -49,9 +46,11 @@ return $.extend( helper, {
...
 
 		msg = msg ? msg + "." : "";
-		actual = { width: widthAfter, height: heightAfter },
-		expected = { width: widthBefore + dw, height: heightBefore + dh };
-		assert.deepEqual( actual, expected, "resized[" + 50 + ", " + 50 + "] " + msg );
+
+		actualDH = heightAfter - heightBefore;
+		actualDW = widthAfter - widthBefore;
+
+		assert.ok( Math.abs(actualDH - dh) <= 1 && Math.abs(actualDW - dw) <= 1, "resized[50, 50] " + msg );
```
jzaefferer: How about using two `assert.close()` calls instead? It already does the [Math.abs and comparison stuff](https://github.com/jquery/jquery-ui/blob/master/external/qunit-assert-close/qunit-assert-close.js#L13-L14).

Something like this (didn't test):
```
assert.close(actualDH, dh, 1, "...");
assert.close(actualDW, dw, 1, "...");
```
Would provide much more useful output for failures.

scottgonzalez: I was going to, and I can, but it means updating every test that uses this helper to adjust the expected count.

scottgonzalez: Also, if we're going to do that, we should update the helper above, and probably many other places. Didn't seem worth it right now.

jzaefferer: I see. Let's add a TODO comment then, for whenever we end up having to adjust all the tests anyway.

scottgonzalez: > I see. Let's add a TODO comment then, for whenever we end up having to adjust all the tests anyway.

Done.

jzaefferer:
👍

PR #1738, opened 2016-08-30
Effects: Adjust animation duration in tests

scottgonzalez:
With jQuery 3 using `requestAnimationFrame()`, the `setTimeout()` timing
for short animations wasn't working consistently. This resulted in infrequent
failures everywhere (but infrequent enough that it's hard to even notice), but
consistent failures in IE and Edge. Bumping up the duration and running the
assertions in the middle seems to give consistent results.

Eventually, we should refactor this to use `requestAnimationFrame()` in the
tests themselves to avoid problems like this.


mikesherov:
:+1:

apsdehal:
:+1:

PR #1740, opened 2016-09-02
Slider: Add demo for custom handle

scottgonzalez:
Fixes #15023

### DIFF
--- a/demos/slider/custom-handle.html
+++ b/demos/slider/custom-handle.html
@@ -1,0 +1,42 @@
+<!doctype html>
+<html lang="en">
+<head>
+	<meta charset="utf-8">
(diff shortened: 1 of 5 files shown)


jzaefferer:
Works well with mouse, doesn't work with keyboard. Do we expect the user to add `tabindex` to the custom handle markup? Or is this a spinner bug?

jzaefferer:
To me it seems like a bug - we should also add a test.

scottgonzalez:
Yeah, we should handle this inside the slider.

scottgonzalez:
Updated to apply `tabIndex` to all handles; includes a test.

jzaefferer:
Updated diff looks good. Haven't tested it again though.

PR #1742, opened 2016-09-08
Performance improvement for big pages when autoRefresh is false

breck7:
Give that selectables are created probably 10x - 100x more than the user uses them, this change delays creations of the selection (an expensive operation), until first use when autoRefresh is set to false.

In a page with 10,000 selectable elements this reduced init time by >50% (~2 seconds). Although the tradeoff is somewhat slower first use when you drag select, it seems reasonable when autoRefresh is set to false.

### DIFF
--- a/ui/widgets/selectable.js
+++ b/ui/widgets/selectable.js
@@ -54,6 +54,7 @@ return $.widget( "ui.selectable", $.ui.mouse, {
 		this._addClass( "ui-selectable" );
 
 		this.dragged = false;
+		this.refreshed = false;
 
 		// Cache selectee children based on filter
 		this.refresh = function() {
@@ -80,8 +81,12 @@ return $.widget( "ui.selectable", $.ui.mouse, {
 					unselecting: $this.hasClass( "ui-unselecting" )
 				} );
 			} );
+
+			this.refreshed = true;
 		};
-		this.refresh();
+		if ( this.options.autoRefresh ) {
+			this.refresh();
+		}
 
 		this._mouseInit();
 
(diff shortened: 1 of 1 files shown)


ui/widgets/selectable.js
```diff
@@ -80,8 +81,13 @@ return $.widget( "ui.selectable", $.ui.mouse, {
 					unselecting: $this.hasClass( "ui-unselecting" )
 				} );
 			} );
+			if (!this.refreshed) {
```
scottgonzalez: This conditional is unnecessary.

scottgonzalez:
Would you mind [filing a feature ticket](https://bugs.jqueryui.com/newticket) for this?

ui/widgets/selectable.js
breck7: Thanks. Agreed but wasn't sure if it was the style guide or not to favor an extra conditional or an extra assignment. Will remove.

scottgonzalez:
This causes tests to fail. There are also style errors that the linters find.

scottgonzalez:
@breck7 Are you still interested in working on this?

breck7:
Sorry @scottgonzalez. My github messages often go to bulk.

I just did a quick attempt to fix. I'm a little out of the flow with PR's so probably will do it a few times until I get it right. :)

scottgonzalez:
Closing due to inactivity. If you want to continue working on this, please just push an update and we'll reopen. Thanks.

breck7:
I took the time to properly setup my dev environment today (really well organized project--very easy to get set up and run the tests!). What I concluded was that a much simpler solution requiring no code changes is to just avoid creating a selectable until it's needed:

`someElement.one("mouseover", makeSelectable)`

So don't think this pull request is worth reopening. Thanks!

PR #1743, opened 2016-09-09
fix the bug when datepicker using with bootstrap dropdown

priscillapan:
return false : stop the propagation of the click event, so in bootstrap dropdown plugin,  the document can't  catch the click event, which cause the dropdown menu can't be closed


scottgonzalez:
Returning `true` in an event handler shouldn't do anything and is the opposite of what your description says. Please start by [filing a bug report](https://bugs.jqueryui.com/newticket) that includes a reduced test case. Note that because you're specifically talking about how this integrates with a third party plugin, we'll need an abstracted case that explains how jQuery UI is the problem, and not the other plugin.

PR #1744, opened 2016-09-12
Widget: Untrack classes elements when they are removed from the DOM

arschmitz:
Fixes #15043

### DIFF
--- a/tests/unit/widget/classes.js
+++ b/tests/unit/widget/classes.js
@@ -146,33 +146,33 @@ QUnit.test( "._add/_remove/_toggleClass()", function( assert ) {
 	elementLacksClasses( widget, "remove", assert );
 } );
 
-QUnit.test( "Classes elements are untracked as they are removed from the dom", function( assert ) {
+QUnit.test( "Classes elements are untracked as they are removed from the DOM", function( assert ) {
 	assert.expect( 9 );
 
 	var widget = $( "#widget" ).classesWidget();
 	var instance = widget.classesWidget( "instance" );
 
-	assert.equal( 3, instance.classesElementLookup[ "ui-classes-span" ].length,
+	assert.equal( instance.classesElementLookup[ "ui-classes-span" ].length, 3,
 		"Widget is tracking 3 ui-classes-span elements" );
-	assert.equal( 3, instance.classesElementLookup[ "ui-core-span-null" ].length,
+	assert.equal( instance.classesElementLookup[ "ui-core-span-null" ].length, 3,
 		"Widget is tracking 3 ui-core-span-null elements" );
-	assert.equal( 3, instance.classesElementLookup[ "ui-core-span" ].length,
+	assert.equal( instance.classesElementLookup[ "ui-core-span" ].length, 3,
 		"Widget is tracking 3 ui-core-span elements" );
 
 	widget.find( "span" ).eq( 0 ).remove();
-	assert.equal( 2, instance.classesElementLookup[ "ui-classes-span" ].length,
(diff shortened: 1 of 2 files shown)


tests/unit/widget/classes.js
```diff
@@ -144,4 +146,34 @@ QUnit.test( "._add/_remove/_toggleClass()", function( assert ) {
 	elementLacksClasses( widget, "remove", assert );
 } );
 
+QUnit.test( "Classes elements are untracked as they are removed from the dom", function( assert ) {
```
scottgonzalez: dom -> DOM

tests/unit/widget/classes.js
```diff
@@ -144,4 +146,34 @@ QUnit.test( "._add/_remove/_toggleClass()", function( assert ) {
...
+	assert.expect( 9 );
+
+	var widget = $( "#widget" ).classesWidget();
+	var instance = widget.classesWidget( "instance" );
+
+	assert.equal( 3, instance.classesElementLookup[ "ui-classes-span" ].length,
```
scottgonzalez: All of these assertions have the actual and expected values inverted.

ui/widget.js
```diff
@@ -514,6 +514,16 @@ $.Widget.prototype = {
 			}
 		}
 
+		this._on( options.element, {
+			"remove": function( event ) {
```
scottgonzalez: Let's move this to a named function on the widget.

ui/widget.js
```diff
@@ -514,6 +514,16 @@ $.Widget.prototype = {
...
+				$.each( that.classesElementLookup, function( key, value ) {
+					if ( $.inArray( event.target ) ) {
```
scottgonzalez: Do you happen to know the performance of `.not()` vs `$.inArray()`? Is this likely to actually improve performance?

cc @jquery/core

arschmitz: Not sure about performance of `.not()` vs `inArray` but the line below also calls `.get` and then creates a new jQuery object which is what i was trying to avoid doing on every iteration since in many cases this object will contain many arrays most of which won't contain the element we are looking for.

scottgonzalez: I'm pretty sure `.get()` and `$( array )` are both trivial.

mgol: `.get()` is just doing the `slice` on the jQuery.object: https://github.com/jquery/jquery/blob/3.1.0/src/core.js#L73. `$( array )` is only a little more involved as it passes through `jQuery.makeArray`.

dmethvin: Isn't this call missing the `array`?

arschmitz: @dmethvin sure is but that is fixed already this comment thread is on an outdated commit https://github.com/jquery/jquery-ui/pull/1744/files#diff-6d5657679729dec110717fc4d667b823R534

ui/widget.js
```diff
@@ -524,6 +528,15 @@ $.Widget.prototype = {
 		return full.join( " " );
 	},
 
+	_untrackClassesElement: function( event ) {
+		var that = this;
+		$.each( that.classesElementLookup, function( key, value ) {
+			if ( $.inArray( event.target, that.classesElementLookup[ key ] ) !== -1 ) {
```
scottgonzalez: You can use `value` here.

scottgonzalez:
Looks good.

ui/widget.js
arschmitz: @scottgonzalez @mgol Looks like doing the `inArray` check first is MUCH faster
http://jsbin.com/zaparohuba/edit?html,js,output

PR #1746, opened 2016-09-14
Sortable: Use an event object for simulated `mouseup` in `cancel()`

scottgonzalez:
Regression caused by a1d69208bad175a27c7b50c27fdc10001563cd4d

Fixes #15042


arschmitz:
👍

PR #1747, opened 2016-09-16
Autocomplete: fix XSS in JSONP demo

wenz:
fixes #15048

### DIFF
--- a/demos/autocomplete/search.php
+++ b/demos/autocomplete/search.php
@@ -2,7 +2,7 @@
 
 sleep( 2 );
 // no term passed - just exit early with no response
-if (empty($_GET['term'])) exit ;
+if (!isset($_GET["term"]) || empty($_GET['term'])) exit ;
 $q = strtolower($_GET["term"]);
 // remove slashes if they were magically added
 if (get_magic_quotes_gpc()) $q = stripslashes($q);
@@ -585,7 +585,7 @@ foreach ($items as $key=>$value) {
 // json_encode is available in PHP 5.2 and above, or you can install a PECL module in earlier versions
 $output = json_encode($result);
 
-if ($_GET["callback"]) {
+if (isset($_GET["callback"]) && preg_match("/^[A-Za-z0-9_]+$/", $_GET["callback"])) {
 	$output = $_GET["callback"] . "($output);";
 }
 
--- a/demos/autocomplete/search.php
+++ b/demos/autocomplete/search.php
@@ -2,7 +2,7 @@
 
 sleep( 2 );
 // no term passed - just exit early with no response
-if (!isset($_GET["term"]) || empty($_GET['term'])) exit ;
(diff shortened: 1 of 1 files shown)


demos/autocomplete/search.php
```diff
@@ -2,7 +2,7 @@
 
 sleep( 2 );
 // no term passed - just exit early with no response
-if (empty($_GET['term'])) exit ;
+if (!isset($_GET["term"]) || empty($_GET['term'])) exit ;
```
scottgonzalez: This `isset()` is duplicative of the `empty()` check.

wenz: agreed. Fixed.

scottgonzalez:
This is too restrictive; it would reject many valid callback function names. Properly validating the callback name is too complex and there's no actual attack vector to worry about here.

wenz:
I respectfully disagree.

1) in the context of the demo, the filtering works fine, since it alows jQuery's default callback name consisting of jQuery, many digits, and one underscore. In addition, which valid callback function name is rejected by the regex? We might consider making $ and . valid characters, but what else is missing?
2) PoC: put the (unpatched) file on a web server and load http://server/path/to/search.php?callback=<script>alert(1)</script>&term=X in a browser without built-in XSS protection (e.g. Firefox). The JavaScript code is executed in the security context of the target server, a classic XSS vector.

scottgonzalez:
> In addition, which valid callback function name is rejected by the regex? We might consider making $ and . valid characters, but what else is missing?

Tons and tons of unicode characters are missing.

> 2) PoC: put the (unpatched) file on a web server and load http://server/path/to/search.php?callback=&term=X in a browser without built-in XSS protection (e.g. Firefox). The JavaScript code is executed in the security context of the target server, a classic XSS vector.

The JavaScript code being an array, which does nothing. How is that a classic XSS vector?

wenz:
GitHub ate the script tag in my URL. Try this:
http://server/path/to/search.php?callback=[script]alert(1)[/script]&term=X
but replace the square brackets with angle brackets.

scottgonzalez:
So this is about linking directly to `search.php` and injecting HTML, not about injecting JS in the JSONP request. Simply passing the callback through `htmlspecialchars()` would prevent someone from pretending to load a real page and won't break any valid callback names.

wenz:
htmlspecialchars() would also work, however I personally prefer the validation approach over escaping here, since a callback name with (for instance) angle brackets should lead to nothing being returned. 

The proposed fix works for the sample code. If someone is changing the example, e.g. by using a callback name with unicode characters, it's probably their responsibility to adapt the PHP script, as well. If you wish, I could amend the PR to use htmlspecialchars() despite my reservations. 

Anyway, this must be fixed asap, since the same file is also residing on the jqueryui.com domain.

PR #1748, opened 2016-09-18
Remove role:"presentation" in tabs.js.

milk54:
According to the ARIA in HTML (https://www.w3.org/TR/html-aria/), The a element with a href can to use ARIA roles is button, checkbox, menuitem, menuitemcheckbox, menuitemradio, radio, tab, switch or treeitem.
So, The a element can't to use role="presentation". (invalid)
Because, It has been reported in bug by jQuery UI, but It has been laid aside for two year.
https://bugs.jqueryui.com/ticket/10122

Please merge this pull request.

### DIFF
--- a/ui/widgets/tabs.js
+++ b/ui/widgets/tabs.js
@@ -431,7 +431,6 @@ $.widget( "ui.tabs", {
 			return $( "a", this )[ 0 ];
 		} )
 			.attr( {
-				role: "presentation",
 				tabIndex: -1
 			} );
 		this._addClass( this.anchors, "ui-tabs-anchor" );


jzaefferer:
Spec compliance is nice, but not our primary goal. Foremost, it needs to work for our users. That said, have you tested this (before and after) in a screenreader? See http://wiki.jqueryui.com/w/page/43766322/Accessibility-Overview for details on what screenreaders we (try to) support.

milk54:
I Knew it. Because,I had do test 4 type screen readers(PC-Talker7+IE11,NVDA2016.3jp+Firefox48.0.2,iOS9.3.5 ViceOver,Andoroid5.0.1 TalkBack5.0.4) too.
Of course, For both if li element has role="presentation" is not the case was the same result.(no problem)
If the same result, I think better li element don't has role="presentation" rather than element has role="presentation".
HTML code of jQuery UI - v1.9.0pre - 2012-05-31 was li element has role="presentation" and a element has role="tab".
Why did that change by version up? Please, Tell me reason.

scottgonzalez:
Perhaps screen readers have changed how they handle this in the past four years. If you're not finding any differences with or without the role, it sounds like it's safe to remove it.

@colinbdclark @hanshillen Do you see any problems with this?

colinbdclark:
I asked a few of the ARIA and accessibility experts at the IDRC, and here is a summary of their responses:

"I would actually suspect that most AT would disregard <code>role=“presentation”</code> in this case, much like VoiceOver ignores empty alt text when it’s used on images serving as links, since it’s impossible for links to be decorative.  So based on what I could gather I’d say remove it, but maybe I’m missing some context."

@klown said: "There is a new section in the ARIA 1.1 draft spec that describes cases where applying role="presentation" has no effect; indeed, where user agents are required to ignore the role and use the native semantic.  I think this case is one of those."

I hope this helps!

klown:
@colinbdclark wrote:

> ﻿﻿@klown said: "There is a new section in the ARIA 1.1 draft spec that describes cases where applying role="presentation" has no effect; indeed, where user agents are required to ignore the role and use the native semantic. I think this case is one of those."

Here is a link:
[Presentational Role Conflict Resolution]https://rawgit.com/w3c/aria/master/aria/aria.html#conflict_resolution_presentation_none)

scottgonzalez:
Thanks. In the future, please make sure to run the tests as this caused failures. I updated the tests before merging.

milk54:
Thank you. I ran the test the latest [updated] data by Travis CI.(commit b9ffc34710212fd910717ab735818ef265c9372e)

Test results was without errors.

build:passed

milk54:
I wrote report "[I verified how to read many ARIA tab interface example by screen readers.](https://white-stage.com/article.php/a11y-tabmenu_en)" Please read it.

PR #1749, opened 2016-09-23
Sortable: fix wrong initial helper position in scrolled container (#15021)

typytype:
move assignment of this.cssPosition above code that calls _getParentPosition and _getRelativeOffset so that this.cssPosition isn't undefined in those methods. 

This is a fix for issue #15021 - sortable initially gets the helper position wrong if the sortable list is in a scrolling element.

### DIFF
--- a/ui/widgets/sortable.js
+++ b/ui/widgets/sortable.js
@@ -218,7 +218,7 @@ return $.widget( "ui.sortable", $.ui.mouse, {
 			top: this.offset.top - this.margins.top,
 			left: this.offset.left - this.margins.left
 		};
-		
+
 		// Only after we got the offset, we can change the helper's position to absolute
 		// TODO: Still need to figure out a way to make relative sorting possible
 		this.helper.css( "position", "absolute" );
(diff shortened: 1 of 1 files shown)


ui/widgets/sortable.js
```diff
@@ -219,6 +219,11 @@ return $.widget( "ui.sortable", $.ui.mouse, {
 			left: this.offset.left - this.margins.left
 		};
 
+		// Only after we got the offset, we can change the helper's position to absolute
```
scottgonzalez: This comment is now wrong.

ui/widgets/sortable.js
```diff
@@ -231,11 +236,6 @@ return $.widget( "ui.sortable", $.ui.mouse, {
 			relative: this._getRelativeOffset()
```
scottgonzalez: Isn't this now broken for relative positioned elements?

typytype: yeah my bad. not sure how the relative position works since we're always setting it to absolute. but i've moved it around so relative offset is set before the position is changed to absolute.

PR #1750, opened 2016-09-23
Resizable: Fix aspectRatio cannot be changed after initialization

sergeir82:
Fixes #4186

### DIFF
--- a/tests/unit/resizable/options.js
+++ b/tests/unit/resizable/options.js
@@ -156,7 +156,7 @@ QUnit.test( "aspectRatio: Resizing can move objects", function( assert ) {
 } );
 
 QUnit.test( "aspectRatio: aspectRatio can be changed after initialization", function( assert ) {
-	assert.expect( 2 );
+	assert.expect( 4 );
 
 	var target = $( "#resizable1" )
 		.resizable( { aspectRatio: 1 } )
@@ -168,6 +168,13 @@ QUnit.test( "aspectRatio: aspectRatio can be changed after initialization", func
 
 	assert.equal( target.width(), 180, "compare width - size change" );
 	assert.equal( target.height(), 100, "compare height - no size change" );
+
+	target.resizable( "option", "aspectRatio", 2 );
+
+	testHelper.drag( handle, -40 );
+
+	assert.equal( target.width(), 140, "compare width - size change" );
+	assert.equal( target.height(), 70, "compare height - size change in proper relation" );
 } );
 
 QUnit.test( "containment", function( assert ) {
--- a/tests/unit/resizable/options.js
+++ b/tests/unit/resizable/options.js
@@ -166,7 +166,7 @@ QUnit.test( "aspectRatio: aspectRatio can be changed after initialization", func
 
 	testHelper.drag( handle, 80 );
 
-	assert.equal( target.width(), 180, "compare width -  size change" );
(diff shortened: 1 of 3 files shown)


tests/unit/resizable/resizable.html
```diff
@@ -44,17 +44,17 @@
 <div id="qunit"></div>
 <div id="qunit-fixture">
 
-<div id="container">
-	<div id="resizable1">I'm a resizable.</div>
-	<div id="resizer1" class="ui-resizable-handle ui-resizable-s"></div>
-</div>
+	<div id="container">
```
scottgonzalez: Please put the indentation back to how it was.

ui/widgets/resizable.js
```diff
@@ -226,6 +226,9 @@ $.widget( "ui.resizable", $.ui.mouse, {
 			this._removeHandles();
 			this._setupHandles();
 			break;
+		case "aspectRatio":
+			this._aspectRatio = !!value;
```
scottgonzalez: You cannot cast to a boolean here, numbers are used to define the ratio.

tests/unit/resizable/options.js
```diff
@@ -155,6 +155,21 @@ QUnit.test( "aspectRatio: Resizing can move objects", function( assert ) {
...
+
+	var target = $( "#resizable1" )
+		.resizable( { aspectRatio: 1 } )
+		.resizable( "option", "aspectRatio", false );
+
+	var handle = ".ui-resizable-e";
+
+	testHelper.drag( handle, 80 );
+
+	assert.equal( target.width(), 180, "compare width -  size change" );
```
scottgonzalez: Extraneous space after dash.

ui/widgets/resizable.js
sergeir82: We have two properties `this.aspectRatio` and `this._aspectRatio`.
In aspectRatio we have numbers are used to define the ratio. It is set in `_mouseStart` method https://github.com/jquery/jquery-ui/blob/1.12.1/ui/widgets/resizable.js#L397. 
We have numeric `aspectRatio` even in cases when we do not set  aspectRatio in options, because we must know it when resize with `shiftKey` pressed.
We use `_aspectRatio` to know save aspect ratio without pressed `shiftKey` (anytime) or not.

tests/unit/resizable/resizable.html
sergeir82: Commit https://github.com/jquery/jquery-ui/pull/1750/commits/2edcbe30a8524cef11a0db795ef2de00f5e5657a

tests/unit/resizable/options.js
sergeir82: Commit https://github.com/jquery/jquery-ui/pull/1750/commits/ffeb5aba1271ac39262c88230acc8e0d5545200f

ui/widgets/resizable.js
sergeir82: Add tests for setting `aspectRatio` numeric value. Commit https://github.com/jquery/jquery-ui/pull/1750/commits/02796d6dac21c088f30003c11c331cf3499f67ca

scottgonzalez: Ah, thanks for pointing that out. Looking at just the diff, I forgot about the two.

PR #1752, opened 2016-09-28
Datepicker: yearOrder option

kociolekt:
Added yearOrder option for ascending/descending year select ordering. It always was ascending. Created demo for this option,


scottgonzalez:
Thanks, but datepicker is in the middle of a complete rewrite (see the datepicker branch), so we're not accepting any new features right now.

PR #1753, opened 2016-09-29
Theme fixes

arschmitz:
Fixes jquery/download.jqueryui.com#335
Ref https://forum.jquery.com/topic/checkboxradio-widget-checkbox-click-doesn-t-work-with-ui-lightness-theme-22-9-2016

### DIFF
--- a/themes/base/theme.css
+++ b/themes/base/theme.css
@@ -119,7 +119,7 @@ a.ui-button:active,
 .ui-icon-background,
 .ui-state-active .ui-icon-background {
 	border: #003eff/*{borderColorActive}*/;
-	background-color: #ffffff/*{fcActive}*/;
+	background-color: #ffffff/*{bgColorContent}*/;
 }
 .ui-state-active a,
 .ui-state-active a:link,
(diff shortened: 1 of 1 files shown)


jzaefferer:
Diff looks good to me.

scottgonzalez:
This now also fixes https://bugs.jqueryui.com/ticket/15059.

themes/base/theme.css
```diff
@@ -119,7 +119,7 @@ a.ui-button:active,
 .ui-icon-background,
 .ui-state-active .ui-icon-background {
 	border: #003eff/*{borderColorActive}*/;
-	background-color: #ffffff/*{fcActive}*/;
+	background-color: #ffffff/*{bgColorContent}*/;
```
scottgonzalez: Why are we pairing the content background with the active border? Why is this not using the active background?

scottgonzalez: Ok. I think it makes sense for these to match. Though it's a shame the active colors don't work here.

PR #1754, opened 2016-09-29
Correct Traditional Chinese translation

asika32764:
Use `週` instead `周`, the previous word is the correct usage in zh_TW


scottgonzalez:
According to [CLDR](https://github.com/unicode-cldr/cldr-dates-full/blob/master/main/zh/dateFields.json), it seems like what we have is correct.

asika32764:
Please see [CLDR zh-Hant](https://github.com/unicode-cldr/cldr-dates-full/blob/master/main/zh-Hant/dateFields.json), it uses `週` not `周`

`周` was used by `zh`, `zh-CN` or `Simplified Chinese`

And `週` was used by `zh-TW`, `zh-Hant`, `zh-HK` or we called `Traditional Chinese`

PR #1755, opened 2016-09-30
Calendar / Datepicker: improve tests and add missing events

fnagel:
- Fix datepicker min / max attribute parsing
- Add change event incl. basic tests
- Add unit tests for calendar select event
- Add unit tests for calendar value option
- Add unit tests for datepicker min / max option
- Introduce dateEqual assertion as proposed in wiki
- Make use of setup / teardown methods and remove helper file as proposed in wiki


fnagel:
@scottgonzalez @jzaefferer FYI

fnagel:
Unit tests work fine in browser and cli when running locally. Any ideas?

ui/widgets/calendar.js
```diff
@@ -133,12 +125,31 @@ return $.widget( "ui.calendar", {
...
+		}
+	},
+
 	_handleKeydown: function( event ) {
 		var pageAltKey = ( event.altKey || event.ctrlKey && event.shiftKey );
 
 		switch ( event.keyCode ) {
-		case $.ui.keyCode.ENTER:
-			this.activeDescendant.mousedown();
+			case $.ui.keyCode.ENTER:
```
jzaefferer: The indent shouldn't change.

ui/widgets/calendar.js
```diff
@@ -133,12 +125,31 @@ return $.widget( "ui.calendar", {
 		this._addClass( $( event.currentTarget ), null, "ui-state-hover" );
 	},
 
+	_select: function( event ) {
+		var oldValue = this.options.value ? this.options.value.getTime() : "";
+
+		this._setOption( "value", new Date( $( event.currentTarget ).data( "timestamp" ) ) );
```
jzaefferer: Could this access `this.activeDescendant` instead? Would make the ENTER key event handler simpler, since it wouldn't need to pass the event along.

ui/widgets/datepicker.js
```diff
@@ -61,12 +62,13 @@ var widget = $.widget( "ui.datepicker", {
 
 	_create: function() {
 		this.suppressExpandOnFocus = false;
+		this._parse = new Globalize( this.options.locale ).dateParser( this.options.dateFormat );
 
 		if ( $.type( this.options.max ) === "string" ) {
-			this.options.max = Globalize.parseDate( this.options.max, { raw: "yyyy-MM-dd" } );
+			this.options.max = this._parse( this.options.max );
```
jzaefferer: According to http://wiki.jqueryui.com/w/page/12137778/Datepicker both `min` and `max` should be `date`, not `string`. Is that document out of date?

If so, wasn't it on purpose that we used a fixed iso format for min/max as string, instead of matching the `dateFormat` option? With this change, the developer would need to update min/max whenever the (user facing) `dateFormat` option changes.

jzaefferer: With the parsing happening in `_getCreateOptions`, I'm even more confused, since that uses a different format.

Since I doubt these changes are happening by accident, let's update the spec.

ui/widgets/datepicker.js
```diff
@@ -172,6 +176,9 @@ var widget = $.widget( "ui.datepicker", {
 		},
 		blur: function() {
 			this.suppressExpandOnFocus = false;
+		},
+		change: function( event ) {
```
jzaefferer: Is the native change event reliable enough? Usually the answer to that is "no", may need to do a manual dirty-check in `blur` and trigger `change` from there.

tests/unit/calendar/events.js
```diff
@@ -0,0 +1,104 @@
...
+	// This should not trigger another event
+	function step3() {
+		setTimeout( function() {
+			element.calendar( "disable" );
+			element.find( "table button:eq(10)" ).simulate( "mousedown" );
+			start();
+		}, 50 );
+	}
+
+	step1();
```
scottgonzalez: We do this a lot because it makes it easier to read the order of the test from top to bottom.

ui/widgets/datepicker.js
fnagel: Not sure. Had massive issues creating reliable tests for this use case.

Checking the blur events is komplex due to the timeout handling. I wasn't able to prevent additional events from firing.

ui/widgets/datepicker.js
fnagel: The current implementation is:

1) `min / max` option passed as date
2) `min / max` option passed as string: format value according to `dateFormat` option
3)  `min / max` attributes on input element: format value as `yyyy-MM-dd` (according to HTML specs)

I'm not able to find the reference in the GH comments, but we wanted it this way and I guess we need to update the wiki specs.

ui/widgets/calendar.js
fnagel: You're suggesting to replace `new Date( $( event.currentTarget ).data( "timestamp" ) )` with `this.activeDescendant` right?

Don't think so as `this.activeDescendant` would be the current day, not the selected one.  `this.activeDescendant` will be updated in the next line, in `_updateDayElement`.

ui/widgets/datepicker.js
scottgonzalez: I recall the discussions about this, but I don't recall where the discussions occurred. The main thing we want to support is `Date` instances. I believe what we wanted to do for strings was parse the string into a date and store the date, not the string. For attributes, in order to support valid HTML5, we must follow the spec, and not support custom formatting; we'd still store the options as `Date` instances.

fnagel: >I believe what we wanted to do for strings was parse the string into a date and store the date, not the string. For attributes, in order to support valid HTML5, we must follow the spec, and not support custom formatting; we'd still store the options as Date instances.

That's what we have now. We only store date objects and attributes formatting is hard coded.

Changing the wiki specs then, right?

ui/widgets/calendar.js
jzaefferer: Okay, makes sense.

ui/widgets/datepicker.js
jzaefferer: The above makes sense. Let's update the spec with that.

fnagel:
> It would be a bit easier to get these landed with a smaller scope. Mixing refactoring unit tests and implementation changes isn't necessary, as far as I can tell.

Totally agree. I will make sure next PRs are more specific. 

> As for the failing tests, that looks like a timezone issue. I can reproduce it locally by running tests in a different timezone [...]

Mhh, as always, it's not that easy when using windows. Not sure if it's even possible without changing system time globally... :-/

More important would be how to deal with this for the travis tests? Do we want to test multiple zones? Or just setting a specific one for the tests?

ui/widgets/datepicker.js
fnagel: Done.

jzaefferer:
> Mhh, as always, it's not that easy when using windows. Not sure if it's even possible without changing system time globally... :-/

Does Git Bash help? https://git-for-windows.github.io/

> More important would be how to deal with this for the travis tests? Do we want to test multiple zones? Or just setting a specific one for the tests?

On another project where I ran into a similar problem, we decided to run the test suite twice, once on the CI machine's local timezone, once on a specific, different one. That way we make sure that the tests pass in different timezones. Not necessarily in all of them, but testing 2 is much better than just the default.

Its also possible that these issues only affect the testsuite, in which case it might be much more pragmatic to just set a fixed timezone.

tests/unit/calendar/events.js
```diff
@@ -0,0 +1,104 @@
...
+var element;
+
+module( "calendar: events", {
+	setup: function() {
+		element = $( "#calendar" ).calendar();
+	}
+} );
+
+test( "change", function() {
+	expect( 6 );
```
scottgonzalez: This is using globals. Should use `assert.expect()` (`assert` gets passed as a parameter to the `test()` callback).

Looks like the datepicker branch is missing some updates to not allow the use of the QUnit globals. I guess we'll address that separately.

tests/unit/calendar/events.js
```diff
@@ -0,0 +1,104 @@
...
+			ok( shouldFire, "change event fired" );
+			equal(
+				event.type,
+				"calendarchange",
+				"change event"
+			);
+			equal(
+				event.originalEvent.type,
+				eventType,
+				"select originalEvent on calendar button " + eventType
```
scottgonzalez: s/select/change/

tests/unit/calendar/events.js
```diff
@@ -0,0 +1,104 @@
...
+		}
+	} );
+
+	function step1() {
+		setTimeout( function() {
+			eventType = "mousedown";
+			message = "on calendar button " + eventType;
+			element.find( "table button:eq(1)" ).simulate( eventType );
+			step2();
+		}, 50 );
```
scottgonzalez: Unwrap the timeouts inside the steps. Use the timeout for invoking the next step:

```js
function step1() {
	eventType = "mousedown";
	message = "on calendar button " + eventType;
	element.find( "table button:eq(1)" ).simulate( eventType );
	setTimeout( step2, 50 );
}
```

tests/unit/calendar/options.js
```diff
@@ -270,46 +268,66 @@ test( "min / max", function() {
 
 test( "numberOfMonths", function() {
 	expect( 6 );
-	var date = new Date( 2015, 8 - 1, 1 ),
-		input = $( "#calendar" ).calendar( {
+
+	var date = new Date( 2015, 8 - 1, 1 );
+
+	// Number of month option does not work after init
```
scottgonzalez: This needs to work. Can you just make this a TODO for now?

tests/unit/calendar/options.js
```diff
@@ -270,46 +268,66 @@ test( "min / max", function() {
...
+test( "value", function( assert ) {
+	expect( 4 );
+
+	var date = new Date( 2016, 5 - 1, 23 );
+
+	assert.equal( element.calendar( "option", "value" ), null, "Initial value" );
+
+	element.calendar( "option", "value", date );
+	assert.dateEqual( element.calendar( "option", "value" ), date, "Value set" );
+	equal( widget.find( "table button.ui-state-active" ).data( "timestamp" ), 1463954400000, "Active button timestamp" );
```
scottgonzalez: We need to prefix our data keys, so this should be `ui-calendar-timestamp`.

tests/unit/calendar/options.js
```diff
@@ -270,46 +268,66 @@ test( "min / max", function() {
...
+
+	element.calendar( "option", "value", "invalid" );
+	assert.dateEqual( element.calendar( "option", "value" ), date, "Value after invalid parameter" );
```
scottgonzalez: Perhaps we've discussed this before, but I don't remember talking through this. Is there a reason we leave the previous value instead of reverting to `null`? Native date inputs will clear their value if you try to set an invalid value.

tests/unit/datepicker/datepicker.html
```diff
@@ -13,8 +13,7 @@
 <div id="qunit"></div>
 <div id="qunit-fixture">
 
-<input type="text" id="datepicker">
-<input type="text" id="datepicker2">
+<input type="text" id="datepicker" />
```
scottgonzalez: Removing the trailing slash. We don't use XHTML.

ui/widgets/datepicker.js
```diff
@@ -79,30 +81,33 @@ var widget = $.widget( "ui.datepicker", {
 	_getCreateOptions: function() {
 		var max = this.element.attr( "max" ),
 			min = this.element.attr( "min" ),
+			parser = new Globalize( "en" ).dateParser( { raw: "yyyy-MM-dd" } ),
```
scottgonzalez: Not directly related to this PR, but does Globalize always contain the `en` data or do we need to tell users to always include this to handle the attribute parsing?

ui/widgets/datepicker.js
scottgonzalez: The native `change` event should be fine for user input. It's only programmatic value changes that would cause the event to not be fired.

ui/widgets/datepicker.js
```diff
@@ -334,10 +341,16 @@ var widget = $.widget( "ui.datepicker", {
 	},
 
 	_getParsedValue: function() {
-		return this.calendarInstance._parse( this.element.val() );
+		return this._parse( this.element.val() );
 	},
 
 	_setOption: function( key, value ) {
+		if ( key === "max" || key === "min" ) {
+			if ( $.type( value ) === "string" ) {
```
scottgonzalez: Just use `typeof`.

fnagel: Sure. Just for my curiosity: Why not using the internals?

scottgonzalez: jQuery's type detection is mostly unneeded in modern browsers. For strings, it was never needed. `$.type()` was just an abstraction away from the `$.is*()` methods which were limited to types that mattered.

fnagel:
> Does Git Bash help? https://git-for-windows.github.io/

@jzaefferer Already using git bash, but tried power shell as well :-/

tests/unit/calendar/options.js
fnagel: No, no reason afaik and I don't remember talking trough this. I'll see if this can be done easily within the `_setOption` method.

tests/unit/calendar/events.js
fnagel: Using `assert` seems already possible, see: https://github.com/jquery/jquery-ui/pull/1755/commits/53f39f5008a46a332e3a5416875bcd683eb9a1eb#diff-4b5efcb2dfc23f659687625a1565c9b3R140

I guess we want to use this anywhere?

tests/unit/calendar/options.js
fnagel: Done.

fnagel:
Added timezone / travis functionality as a todo task in wiki.

ui/widgets/datepicker.js
fnagel: Thanks for the info.

Done.

tests/unit/calendar/events.js
fnagel: Done.

tests/unit/calendar/events.js
fnagel: Done.

fnagel:
Oh man, changing the system timezone via the UI works (but does not using similar console command). Seems this messes with some of local tools but for a quick test this seems to work.

fnagel:
> Have you tried running tests inside a Docker container? If not, I can help set that up.

I'm using VirtualBox / Vagrant VMs for other projects but always tried to get around this for jQuery UI.
And hey, perhaps it's a good idea to have at least a one developer building the project on Windows, right?

fnagel:
@jzaefferer @scottgonzalez Updated the branch. Let my know what you think!

ui/widgets/calendar.js
```diff
@@ -646,7 +646,8 @@ return $.widget( "ui.calendar", {
 		if ( arguments.length ) {
 			this.valueAsDate( this._parse( value ) );
 		} else {
-			return this._format( this.option( "value" ) );
+			return this.option( "value" ) ===  null ?
```
jzaefferer: How about `return this.option( "value" ) ===  null ? null : ...`?

ui/widgets/datepicker.js
jzaefferer: For this raw pattern, the locale might not be required, since there's nothing locale-specific, I think. In general, Globalize doesn't bundle any locales.

tests/unit/calendar/methods.js
```diff
@@ -56,7 +56,7 @@ test( "value", function( assert ) {
 	equal( this.element.calendar( "value" ), "1/1/14", "getter" );
 
 	this.element.calendar( "value", "abc" );
-	equal( this.element.calendar( "value" ), "1/1/14", "Setting invalid values should be ignored." );
+	equal( this.element.calendar( "value" ), null, "Setting invalid values." );
```
fnagel: We are currently using this for `assert.equalDate` and `assert.expect` only.

jzaefferer:
Updates look (mostly) good. Still need to fix the CI-failing test.

fnagel:
@scottgonzalez  

> Those get reverted to null as well.
https://github.com/jquery/jquery-ui/pull/1755#discussion_r82991504

Seems I'm no longer able to see or answer this comment. Can you add this again on the current code base?

scottgonzalez:
> Seems I'm no longer able to see or answer this comment. Can you add this again on the current code base?

https://github.com/jquery/jquery-ui/pull/1755#discussion_r82991504

scottgonzalez:
@fnagel I'm not sure what's going on with GitHub's links, but the comment was about `min` and `max` reverting to `null` if the dates provided are invalid.

fnagel:
@scottgonzalez 

> was about min and max reverting to null if the dates provided are invalid

Sorry, still missing the context. You wan't me to reset the `min / max` option when trying to change their values to an invalid one? So basically the same functionality for `min / max` option, we now have for `value` option and method. Correct?

ui/widgets/datepicker.js
rxaviers: @jzaefferer is correct

fnagel: This particular line is about parsing the input min / max attributes. So this pattern (`yyyy-MM-dd`) should never change, regardless of the used locale.

Anyway, if we want to keep it this way, we will need to tell users to add it, like done here (for multiple locales):  https://github.com/jquery/jquery-ui/pull/1755/commits/ea986f2808684c0112d56776d76ea35e9cdb77da

Using a simple split seems interesting as we could remove one of two (direct, not within Calendar widget) usages of Globalize in datepicker. In addition we could revert additions made above and keep the dependencies to a minimum.

So, +1 from me for using a simple split.

rxaviers: Nice that you +1 the idea of using split. Emphasizing it: using Globalize would make sense if parsing user input (which could be subject to locale differences), not developer input. :)

scottgonzalez:
> You wan't me to reset the min / max option when trying to change their values to an invalid one? So basically the same functionality for min / max option, we now have for value option and method. Correct?

Correct.

ui/widgets/datepicker.js
fnagel: When using something like

````
````

Comment (jzaefferer) [inline ui/widgets/datepicker.js]: I think you need to set the correct offset for the month, since it is defined as "Integer value representing the month, beginning with 0 for January to 11 for December."

**Comment by jzaefferer** [comment]
I've looked at the timezone-dependent failure. When running the calendar tests with Sao Paulo timezone, the actual/expected values are only 4 seconds apart:
```js
>> calendar: options - value
>> Message: Active button timestamp
>> Actual: 1463972400000
>> Expected: 1463954400000

> Date(1463972400000)
'Tue Oct 25 2016 12:07:23 GMT+0200 (CEST)'
> Date(1463954400000)
'Tue Oct 25 2016 12:07:27 GMT+0200 (CEST)'
```
Still not sure what's causing this. If we can't figure it out, we can probably make the test a little less precise. We don't care about the time component here, only the date, right?

ui/widgets/datepicker.js
jzaefferer: Like tests are littered with `x - 1` for the month everywhere.

fnagel: No idea why, but I was focused on the missing hour not the month. Stupid me.

Pushed a new commit with this.

fnagel:
> We don't care about the time component here, only the date, right?

Correct, calendar and datepicker don't care about time at the moment, but we should keep in mind that timepicker addon is a widely used one.

fnagel:
@scottgonzalez Added invalid min / max reset  incl. tests

jzaefferer:
> Correct, calendar and datepicker don't care about time at the moment, but we should keep in mind that timepicker addon is a widely used one.

Let's keep the timestamp then, but update the test to only look at the date, not the time.

fnagel:
> Let's keep the timestamp then, but update the test to only look at the date, not the time.

So, using our dateequal helper everywhere should help, as we already check for year / month / date only:
https://github.com/fnagel/jquery-ui/blob/datepicker-working/tests/lib/qunit-assert-dateequal.js

fnagel:
@jzaefferer @scottgonzalez 

Fixed tests by replacing a timestamp compare with a dateEqual assrert.

Ready to merge?

fnagel:
CLA error?

tests/unit/calendar/options.js
```diff
@@ -327,9 +327,9 @@ test( "value", function( assert ) {
 
 	this.element.calendar( "option", "value", date );
 	assert.dateEqual( this.element.calendar( "option", "value" ), date, "Value set" );
-	equal(
-		this.widget.find( "table button.ui-state-active" ).data( "ui-calendar-timestamp" ),
-		1463954400000,
+	assert.dateEqual(
+		new Date( this.widget.find( "table button.ui-state-active" ).data( "ui-calendar-timestamp" ) / 1000 ),
```
jzaefferer: This is odd - why is the timestamp data attribute in a different format than the one accepted by `new Date`? Shouldn't those match?

jzaefferer:
The new Foundation is still working on a new CLA check. Until that is ready we can ignore the check at least for team members.

tests/unit/calendar/options.js
fnagel: You'r right, both should be milliseconds. Changed,

fnagel:
@jzaefferer So, merge ready now, right? I would merge master in datepicker and push that right away, ok?

jzaefferer:
Yep

PR #1759, opened 2016-10-11
Create datepicker-en-CA.js

EricManiez:


scottgonzalez:
Thanks, but we're not accepting any new locales. Datepicker is in the middle of a rewrite in which we're switching to CLDR for the locales.

PR #1765, opened 2016-10-31
Build: Stop using the jquery-ui-future browser set

mgol:
jQuery UI now doesn't support browsers not suspported by latest jQuery
so separating the browser sets no longer makes sense.


mgol:
I'll remove the `jquery-ui-future` browser set from TestSwarm once this lands.

PR #1769, opened 2016-11-16
Widget: Avoid tracking remove event handlers for classes options

scottgonzalez:
Fixes #15082
Fixes #15095

### DIFF
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -493,11 +493,29 @@ $.Widget.prototype = {
 			classes: this.options.classes || {}
 		}, options );
 
+		function bindRemoveEvent() {
+			options.element.each( function( _, element ) {
+				var isTracked = $.map( that.classesElementLookup, function( elements ) {
+					return elements;
+				} )
+					.some( function( elements ) {
+						return elements.is( element );
+					} );
+
+				if ( !isTracked ) {
+					that._on( $( element ), {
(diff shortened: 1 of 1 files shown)


arschmitz:
👍

scottgonzalez:
Updated to avoid leaking event listeners.

wangchangshan:
I also find this issue, thanks for your works ~~

scottgonzalez:
We might need to add a `remove` event listener to the element that is having classes added so that we untrack the element when it gets removed. This would catch scenarios like autocomplete menus having items created and destroyed constantly.

scottgonzalez:
Updated to actually work. I confirmed that this fixes all reported issues with the `remove` event and have updated the commit message to reference all tickets.

ui/widget.js
```diff
@@ -493,11 +493,29 @@ $.Widget.prototype = {
 			classes: this.options.classes || {}
 		}, options );
 
+		function bindRemoveEvent() {
+			options.element.each( function( _, element ) {
+				var isTracked = $.map( that.classesElementLookup, function( elements ) {
+					return elements;
+				} )
+					.some( function( elements ) {
```
arschmitz: This is a single element, so the name should be singular, though element is already taken

scottgonzalez: This is a collection of elements for a single class.

PR #1771, opened 2016-11-25
Added support for touch events in jquery sortable etc.

danielgindi:
I've tested this with `sortable` only, but it should work with the other plugins as well.


scottgonzalez:
Thanks, but we don't want to add support for Touch Events. Use [touch punch](http://touchpunch.furf.com/) for that. Support for Pointer Events is planned for the next major release.

PR #1772, opened 2016-11-29
Optimize PNG files with zopflipng

joliss:
I optimized the PNG files with zopflipng by running

```bash
for i in `find demos themes tests -name \*.png`; do zopflipng -y "$i" "$i"; done
```

PNG uses DEFLATE internally, and the zopfli algorithm often finds better DEFLATE representations than zlib for the same text. Most files were already optimized, but for some, zopflipng found better representations, saving us a few hundred bytes.


jzaefferer:
We landed #1763 a little while ago, which, I guess, wasn't using the best available tool for the job? Hopefully this is the one.

scottgonzalez:
And before that one, we landed #1710 which did use zopflipng!

PR #1773, opened 2016-12-02
[1.13] Feature/remove some jquery 3 migrate warnings

eirslett:
There are some migration errors we get when using jQuery 3:

> JQMIGRATE: jQuery.expr[":"] is now jQuery.expr.pseudos

and

> jQuery.unique is deprecated, use jQuery.uniqueSort

Both `jQuery.expr.pseudos` and `jQuery.uniqueSort` were introduced after 1.7.0 (which is the oldest supported jQuery version for jQuery UI), so this PR uses a polyfill solution.

The unit tests appear to be passing.

Is this a change that would be interesting to merge into jQuery UI? In that case, are there any changes to the implementation, or the commit messages, that have to be done?

1) I'm not sure about where to put the polyfills; now I put them right above the code that had to be polyfilled. Another option would be to put the required polyfills in top of every file (module declaration) that needs them. Putting polyfills in a separate file is one option, but more involved, and it requires the file to be imported; I don't think that is a good idea.

2) Git commit messages should be per component, but this is really a cross-component change. Should I split the commits into separate ones per component?


scottgonzalez:
> Is this a change that would be interesting to merge into jQuery UI? In that case, are there any changes to the implementation, or the commit messages, that have to be done?

Yes, this is something we would be interested in merging. Historically, we just used the deprecated methods as long as possible since that's the easiest solution for supporting so many versions of jQuery Core. However, we have switched to a different model where we upgrade to using the new APIs without any checks and then load in compatibility layers to support older versions of jQuery. You can see how we handle this in [/ui/jquery-1-7.js](https://github.com/jquery/jquery-ui/blob/cc37fe76aac4a09c37de83178e2c818c4400dff6/ui/jquery-1-7.js).

I'd suggest adding `/ui/jquery-1-9.js` for `jQuery.expr.pseudos` and `/ui/jquery-1-11.js` for `.uniqueSort()`. The conditionals and assignments you have seem fine. You'll need to add the two new files as dependencies for `/ui/core.js` which will cause them to be included in the full build, but can easily be excluded in a custom build.

Thanks.

eirslett:
Ok, I moved the polyfills into their own files now. I also removed the surrounding checks.

eirslett:
Sorry to nag you; are you still interested in this PR? (And eventually #1774)

ui/jquery-1-11.js
```diff
@@ -0,0 +1,35 @@
...
+	} else {
+
+		// Browser globals
+		factory( jQuery );
+	}
+}( function( $ ) {
+
+// Support: jQuery 1.11.x or older
+var versions = $.fn.jquery.split( "." );
+if ( parseInt( versions[ 0 ], 10 ) === 1 && parseInt( versions[ 1 ], 10 ) <= 11 ) {
```
scottgonzalez: I know you based this on `jquery-1-7.js`, but you'll notice there's a big comment about why we're using version sniffing instead of feature detection. The feature detection here is very simple, so this condition should change to `if (!$.uniqueSort)`

ui/jquery-1-11.js
```diff
@@ -0,0 +1,35 @@
...
+
+	// $.unique has been renamed to uniqueSort
```
scottgonzalez: uniqueSort -> $.uniqueSort

ui/jquery-1-9.js
```diff
@@ -0,0 +1,35 @@
...
+		factory( jQuery );
+	}
+}( function( $ ) {
+
+// Support: jQuery 1.9.x or older
+var versions = $.fn.jquery.split( "." );
+if ( parseInt( versions[ 0 ], 10 ) === 1 && parseInt( versions[ 1 ], 10 ) <= 9 ) {
+
+	// $.expr [ ":" ] has been renamed to $.expr.pseudos
+	$.expr.pseudos = $.expr[ ":" ];
```
scottgonzalez: Same thing here. You can just check for the existence of `$.expr.pseudos`.

eirslett:
I have updated the PR after code review/feedback, fixing the issues that were raised.

scottgonzalez:
Thanks. These changes look good. I just wanted to give you a heads up that it will be a little bit before this gets merged. Since it introduces new modules, it needs to be part of the 1.13.0 release. We're getting close to the 1.12.2 release, after which this can be merged.

eirslett:
If the introduction of new modules is a breaking change, then maybe another change could be made? Instead of a 1.7 polyfill, a 1.9 polyfill, a 1.12 polyfill etc., there could be a 1.x polyfill for all jQuery versions 1.x. (Considering jQuery 2.0 was released in April 2013, it shouldn't be a big burden for people on jQuery 1.12.3 to include the whole 1.x polyfill)

scottgonzalez:
Introducing a 1.x module still requires 1.13.0, though I'm not against that idea. 1.7 is the heaviest, but it's really not a big price to pay for 1.x support in general, despite the fact that very few sites (comparatively) actually use 2.x or 3.x.

Using statistics from builtwith.com, 1.7.x accounts for about 6.5% of jQuery usage, at ~4.4M sites. All 2.x releases only amount to a quarter of that. 3.x releases are so unpopular that they don't even report usage statistics for it yet. However, these stats don't say anything about rate of release and update.

Barring any objections from @jquery/ui, I'd accept a PR that merged all of the files together (you can even just update this PR).

jzaefferer:
A single module for jQuery "adapters" seems fine to me. These new ones are just one liners and don't need their own files.

scottgonzalez:
How about we just use the existing `jquery-1-7` module as the generic jQuery 1.x support module during 1.12? The name is a bit strange, but we can just say it's defining the lowest version that the module adds support for. We can rename it in 1.13 so that there is no version number (or make it just `jquery-1`).

@eirslett Would you like to make that change? We're getting ready to release 1.12.2 and I'd be happy to get this in there.

eirslett:
Ok, I'll have a look at it.

eirslett:
I moved all the polyfills into jquery-1-7 now, and rebased the commits on top of the master branch.

ui/data.js
```diff
@@ -23,7 +23,7 @@
 		factory( jQuery );
 	}
 } ( function( $ ) {
-return $.extend( $.expr[ ":" ], {
+return $.extend( jQuery.expr.pseudos, {
```
scottgonzalez: This has to stay as `$` instead of `jQuery` for `noConflict()` support. `$` is a local reference via the parameter.

I'll make the change when I land this.

jaredbeck:
Thank you for your work on this, Eirik and Scott. Looking forward to the release of 1.13. Will it fix all jquery 3 deprecation warnings?

PR #1774, opened 2016-12-04
Add ability to run tests with jQuery Migrate

eirslett:
This is inspired by #1773 which fixes some JQMIGRATE warnings; but it only fixed a couple ones I stumbled over. In this PR, I added the ability to run all unit tests with jquery-migrate loaded, so that all warnings show up in the console. This will make it easier to systematically fix all migrate warnings, and possibly catch any new ones that are introduced.

Example: Open the browser and run tests: http://localhost:7777/tests/unit/all.html?jquery=3.1.0&migrate=3.0.0

You can then see warnings that appear in the console.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -395,7 +395,13 @@ grunt.initConfig({
 				"jquery-3.1.0/LICENSE.txt": "jquery-3.1.0/LICENSE.txt",
 
 				"jquery-3.1.1/jquery.js": "jquery-3.1.1/dist/jquery.js",
-				"jquery-3.1.1/LICENSE.txt": "jquery-3.1.1/LICENSE.txt"
+				"jquery-3.1.1/LICENSE.txt": "jquery-3.1.1/LICENSE.txt",
+
+				"jquery-migrate-1.4.1/jquery-migrate.js": "jquery-migrate-1.4.1/dist/jquery-migrate.js",
+				"jquery-migrate-1.4.1/LICENSE.txt": "jquery-migrate-1.4.1/LICENSE.txt",
+
+				"jquery-migrate-3.0.0/jquery-migrate.js": "jquery-migrate-3.0.0/dist/jquery-migrate.js",
+				"jquery-migrate-3.0.0/LICENSE.txt": "jquery-migrate-3.0.0/LICENSE.txt"
 			}
 		}
 	},
--- a/tests/lib/bootstrap.js
+++ b/tests/lib/bootstrap.js
@@ -5,6 +5,7 @@ requirejs.config( {
 		"globalize": "../../../external/globalize/globalize",
 		"globalize/ja-JP": "../../../external/globalize/globalize.culture.ja-JP",
 		"jquery": jqueryUrl(),
+		"jquery-migrate": migrateUrl(),
 		"jquery-simulate": "../../../external/jquery-simulate/jquery.simulate",
 		"jshint": "../../../external/jshint/jshint",
 		"lib": "../../lib",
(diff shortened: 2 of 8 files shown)


tests/lib/bootstrap.js
```diff
@@ -102,6 +107,19 @@ function jqueryUrl() {
...
 }
 
+function migrateUrl() {
+	var version = parseUrl().migrate;
+	var url;
+
+	if ( version === "git" ) {
+		url = "http://code.jquery.com/jquery-migrate-" + version;
+	} else {
+		url = "../../../external/jquery-migrate-" + ( version || "1.12.4" ) + "/jquery-migrate";
```
scottgonzalez: There is no `jquery-migrate-1.12.4`. But we also don't need a default version here since the migrate plugin isn't loaded by default and you have to explicitly specify the version to include.

This also needs to be registered with QUnit so that a UI element is generated. See `tests/lib/qunit.js` and search for `QUnit.config.urlConfig`. Since there's a 1:1 mapping between the jQuery version and the jQuery Migrate version, I'd prefer if the `migrate` config was a boolean. This function should look at the jQuery version to determine the jQuery Migrate version to load.

jQuery 1.x and 2.x -> jQuery Migrate 1.4.1
jQuery 3.x -> jQuery Migrate 3.0.0
jQuery Git -> jQuery Migrate Git

Gruntfile.js
```diff
@@ -395,7 +395,9 @@ grunt.initConfig({
 				"jquery-3.1.0/LICENSE.txt": "jquery-3.1.0/LICENSE.txt",
 
 				"jquery-3.1.1/jquery.js": "jquery-3.1.1/dist/jquery.js",
-				"jquery-3.1.1/LICENSE.txt": "jquery-3.1.1/LICENSE.txt"
+				"jquery-3.1.1/LICENSE.txt": "jquery-3.1.1/LICENSE.txt",
+
+				"jquery-migrate-3.0.0/jquery-migrate.js": "jquery-migrate-3.0.0/jquery-migrate.js"
```
scottgonzalez: We'll want 1.4.1 as well.

scottgonzalez: Missing the license file.

tests/lib/bootstrap.js
eirslett: Updated with your suggestion to resolve jquery-migrate version depending on jquery version.

Gruntfile.js
eirslett: Fixed

eirslett: Fixed

eirslett:
I updated the PR now.
I had to install jquery-migrate from the npmjs tgz files, because the default bower repositories were referring to some [third party repository](https://github.com/appleboy/jquery-migrate), which neither had a license file or version 1.4.1 published.
Trying to install `jquery/jquery-migrate#1.4.1` with  bower downloaded the individual source files of jquery-migrate, but not the built distribution.
Looks like jquery-migrate is configured to be published to npm primarily.

tests/lib/bootstrap.js
```diff
@@ -102,6 +107,21 @@ function jqueryUrl() {
...
 
+function migrateUrl() {
+	var jqueryVersion = parseUrl().jquery || "1.12.4";
+	var url;
+
+	if ( jqueryVersion === "git" ) {
+		url = "http://code.jquery.com/jquery-migrate-git";
+	} else if ( jqueryVersion[ 0 ] === "3" ) {
+		url = "../../../external/jquery-migrate-3.0.0/jquery-migrate";
+	} else {
```
scottgonzalez: Let's add a condition here for `jqueryVersion[ 0 ] === "1" || jqueryVersion[ 0 ] === "2"` and then make the `else` block throw an error so that when jQuery 4 is released, we'll know that we need to check the appropriate jQuery Migrate version.

tests/lib/bootstrap.js
```diff
@@ -68,6 +69,10 @@ function requireTests( dependencies, noBackCompat ) {
 		dependencies.push( "testswarm" );
 	}
 
+	if ( parseUrl().migrate ) {
```
scottgonzalez: This isn't actually working because `parseUrl()` doesn't support booleans right now.

PR #1775, opened 2016-12-07
Removes outline on and after clicking on tab

jhpratt:
I just noticed this, and it's quite annoying. It's a quick and simple fix, so I figured I'd throw a quick pull request together.


scottgonzalez:
This is bad for accessibility, so we will not remove focus outlines.

PR #1776, opened 2016-12-08
Tabs: Add option forceLocal

tjmonk15:
Adds an option to Tabs to have them treat all hrefs as local selectors

### DIFF
--- a/ui/widgets/tabs.js
+++ b/ui/widgets/tabs.js
@@ -52,6 +52,7 @@ $.widget( "ui.tabs", {
 		heightStyle: "content",
 		hide: null,
 		show: null,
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Sorry, but we won't add options that lie about what's happening.

tjmonk15:
How does this option lie about anything?

scottgonzalez:
It says that the tabs are pointing to local content when they're not. This has been discussed ad nauseam and will not be changed. See https://bugs.jqueryui.com/ticket/7822 for more information.

scottgonzalez:
You can of course make your own extension to add such an option, it just won't be accepted upstream.

PR #1780, opened 2016-12-25
Datepicker: I18N for de-CH

MartinGantenbein:
Copied from datepicker-de.js and changed spelling of closeText to Swiss Standard German spelling "Schliessen"


scottgonzalez:
Thanks, but we're in the process of rewriting datepicker and the new version uses CLDR for all of the locale data. As a result, we're not accepting any new locale files.

PR #1781, opened 2017-01-04
Datepicker: Added Philippines Filipino locale

marilouranoy:


scottgonzalez:
Thanks, but we're in the process of rewriting datepicker and the new version uses CLDR for all of the locale data. As a result, we're not accepting any new locale files.

marilouranoy:
Oh I see. Thank you and looking forward for the new version soon.

PR #1782, opened 2017-01-10
Fixed : https://bugs.jqueryui.com/ticket/7822

OzoneGrif:
The "base" tag, if it exists on the page, was not handled correctly for local tabs.

### DIFF
--- a/ui/widgets/tabs.js
+++ b/ui/widgets/tabs.js
@@ -64,11 +64,15 @@ $.widget( "ui.tabs", {
 		var rhash = /#.*$/;
 
 		return function( anchor ) {
-			var anchorUrl, locationUrl;
+			var anchorUrl, locationUrl, baseTag;
 
 			anchorUrl = anchor.href.replace( rhash, "" );
-			locationUrl = location.href.replace( rhash, "" );
-
+			baseTag = document.getElementsByTagName("base");
+			if (baseTag.length > 0)
+				locationUrl = baseTag[0].href.replace( rhash, "" );
+			else
+				locationUrl = location.href.replace( rhash, "" );
+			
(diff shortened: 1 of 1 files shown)


OzoneGrif:
I signed your form but it doesn't recognize my professional email although I attached it to my GitHub account.

scottgonzalez:
That's not how the base element works.

OzoneGrif:
The "base" tag specifies the base URL/target for all relative URLs in a document.
I know perfectly how the "base" tag works. I use it everyday. 
You obviously misunderstood the ticket here.

When you try to use <a href="#localTab"> with the tabs module, it fails because the location doesn't match the new internal href. If there's a "base" tag, your tabs points to : <a href="baseUrl#localTab">, and this fails.

I patch fixes the issue which many people have when you see the number of tickets open on the subject.

scottgonzalez:
> The "base" tag specifies the base URL/target for all relative URLs in a document.

Correct, but it does nothing to the current location. Remove the tabs widget, click the link, and see if you stay on the same page.

OzoneGrif:
I see your point, but then the documentation is wrong :
"Each tab panel may be any valid element but it must have an id which corresponds to the hash in the anchor of the associated tab."

The ID corresponds, yet it doesn't work because of the "base" tag.

Also, if I try to load a distant page with an ID hash, it won't load just that DIV.
The whole system is currently flawed by design. And your argument is not better than my fix.

scottgonzalez:
> The whole system is currently flawed by design.

It's not.

> And your argument is not better than my fix.

My argument is that we're following web standards. I'm sure this sounds harsh, but nothing you say will convince the team to make any changes here. I will not continue discussing this issue.

PR #1783, opened 2017-01-10
Datepicker: Fix parseDate (invoked by public getDate) when there isn't day

wzomer:
When I invoke the getDate method on a Datepicker with the options: { dateFormat: 'mm/yy' }, actually it always returns the current day instead of the chosen date.

Example: http://plnkr.co/edit/WHjbydkmOia45vbz9G3m?p=preview


scottgonzalez:
This is a date picker, not a month picker, so the provided date format is not valid.

PR #1784, opened 2017-01-12
Fix Bug 15094 in datepicker.js

claudi:
Removed addition of showCurrentAtPos to offset  calling adjustInstDate
Bug 15094


claudi:
Sorry, I was wrong with the issue number, this is about **15102**.

PR #1786, opened 2017-01-21
Calendar: callback return values, add refresh callback

fnagel:


scottgonzalez:
> To clarify: Why to we include this.options.value on change and select events when it could be retrieved through the options method?

Providing the value just makes it a little nicer to work with and creates consistency with other widgets where the selected values may not be easily retrievable, such as autocomplete and selectmenu.

jzaefferer:
> Providing the value just makes it a little nicer to work with and creates consistency with other widgets where the selected values may not be easily retrievable, such as autocomplete and selectmenu.

👍 

Good to land then.

PR #1790, opened 2017-01-25
Package: Add jQuery as explicit dependency, with min and max range

jzaefferer:
1.7.0 is our minimum supported version, 3.x.x our maximum.

Closes gh-1779

@Bajix this is supposed to replace your PR, making the dependency a version range, and without the whitespace changes. Would that work for you?

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,9 @@
 	"scripts": {
 		"test": "grunt"
 	},
-	"dependencies": {},
+	"dependencies": {
+		"jquery": ">= 1.7.0 < 4.0.0"
+	},
 	"devDependencies": {
 		"commitplease": "2.3.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -50,7 +50,9 @@
 	"scripts": {
 		"test": "grunt"
 	},
-	"dependencies": {},
+	"dependencies": {
+		"jquery": ">= 1.7.0 < 4.0.0"
```
scottgonzalez: I haven't tested locally, but all of the docs for npm and semver show no spaces between the operator and the version, so just to be safe, let's drop the space before `4.0.0`.

Bajix:
@jzaefferer 👍 

All that really matters is that it's represented

PR #1791, opened 2017-01-25
[WIP] Dialog: Add `icons` option to support titlebar icons

scottgonzalez:
Still needs tests, but I wanted to make sure the team is good with adding this option.

See [ticket 15121](https://bugs.jqueryui.com/ticket/15121).


jzaefferer:
Seems good!

scottgonzalez:
Added tests.

scottgonzalez:
Note that I also changed the implementation because the original implementation didn't handle updating the icon after init.

scottgonzalez:
This PR now also removes the back-compat for `dialogClass`. The deprecated tests were failing because I hadn't updated the options in `common-deprecated.js`. Since this can't land in 1.12 anyway, I figured I might as well just remove the back compat now.

scottgonzalez:
Yeah, this can't be released until 1.13.

nagyimre1980:
for me this would be an important function.  Could 1.13.3 be released?  Thanks

PR #1795, opened 2017-02-24
Resizable: Keep user defined handles on _setOption

kdinev:
Now handles that have been placed in the DOM by the user will persist through _setOption. However, I noticed they will be removed on destroy. I'm not certain if this is expected. If not, I can fix that as well.

Fixes #15084

Related to #1666

### DIFF
--- a/tests/unit/resizable/options.js
+++ b/tests/unit/resizable/options.js
@@ -434,11 +434,20 @@ QUnit.test( "zIndex, applied to all handles", function( assert ) {
 } );
 
 QUnit.test( "setOption handles", function( assert ) {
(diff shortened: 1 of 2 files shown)


scottgonzalez:
Thanks, I'll try to review this in a few hours. But I wanted to respond about the destroy behavior. Any elements that were created by the user should remain after destroy.

kdinev:
Do you want me to include the destroy logic to this PR, or should I create a new one?

scottgonzalez:
I guess since this destroy behavior has been in place for a while, it can be a separate PR.

scottgonzalez:
Thanks!

PR #1796, opened 2017-02-27
Support CommonJS and Node. Add dist directory for themes in Bower.

jhorowitz-firedrum:


scottgonzalez:
We will not commit the `dist` directory. The built themes definitely don't belong in the source repo, and we don't manage the Bower component.

jwalton:
I don't mind not having the dist directory in source, but it should be there in the published package.

jwalton:
Super weird that jquery package has a dist, and jquery-ui does not.

PR #1799, opened 2017-03-01
Corrections for Swedish localization

ajuliano:
Months and weekdays should be in small caps in Swedish.
https://sv.wikipedia.org/wiki/Versalisering


scottgonzalez:
Confirmed that everything should be lowercase, but according to CLDR, the abbreviated months should all have periods at the end.

scottgonzalez:
Also, juni and juli don't get abbreviated because the full name is the same length as the abbreviated version (same for maj).

scottgonzalez:
Merged with the mentioned changes.

PR #1801, opened 2017-03-24
Replace unique by uniqueSort

alaineman:
According to the new jquery 3+


alaineman:
Yes, I have signed the CLA.

scottgonzalez:
We can't use new methods like this because we support older versions of jQuery.

PR #1802, opened 2017-03-31
Calendar: improve date getter/setter, add ARIA tests, clean up

fnagel:


scottgonzalez:
Looks good, but we need to handle the timezone issue. We can probably take a timestamp with an expected date/time and then compare the actual date/time and calculate an offset. Then add that offset to each timestamp we use in the tests.

fnagel:
@scottgonzalez I guess we want some central point to generate and modify Date objects and use that all over our code base, right? Something like a replacement for all `new Date()` usages.

Anyway, I would like to fix this in another PR and merge this, ok?

Did you notice Jon commented in the wiki?

scottgonzalez:
> @scottgonzalez I guess we want some central point to generate and modify Date objects and use that all over our code base, right? Something like a replacement for all new Date() usages.

Yeah, we can put that in the helper file.

> Anyway, I would like to fix this in another PR and merge this, ok?

That's fine with me.

> Did you notice Jon commented in the wiki?

Yup, pretty much any time would work for me. Do you want to send out a [doodle](http://doodle.com/) or similar request to find a good time?

fnagel:
@scottgonzalez 

Ok, I've merged the PR and added a note in the wiki. 

Regarding a Skype call: I've had lots of time in the beginning of the year but I'm about to move in two week. Not sure if I will find time to prepare and join a call. Might be possible you guys need to do this without me -- at least if we don't want to wait ages again. Discussing and adjusting the specs should be impossible without me too...

PR #1805, opened 2017-04-18
Menu: Handle mouse movement mixed with keyboard navigation

scottgonzalez:
Fixes #9357

### DIFF
--- a/ui/widgets/menu.js
+++ b/ui/widgets/menu.js
@@ -78,6 +78,8 @@ return $.widget( "ui.menu", {
 			// them (focus should always stay on UL during navigation).
 			"mousedown .ui-menu-item": function( event ) {
 				event.preventDefault();
+
+				this._activateItem( event );
 			},
 			"click .ui-menu-item": function( event ) {
 				var target = $( event.target );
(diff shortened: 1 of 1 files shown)


ui/widgets/menu.js
```diff
@@ -78,6 +78,8 @@ return $.widget( "ui.menu", {
 			// them (focus should always stay on UL during navigation).
 			"mousedown .ui-menu-item": function( event ) {
 				event.preventDefault();
+
+				this._activateItem( event );
```
scottgonzalez: We activate on `mousedown` to handle the scenario of moving the mouse over an item, navigating away from that item with they keyboard, then clicking with the mouse. Since we won't get a `mouseenter` or `mousemove`, we need to activate on `mousedown`. This matches the OS X menu behavior.

ui/widgets/menu.js
```diff
@@ -107,29 +109,8 @@ return $.widget( "ui.menu", {
 					}
 				}
 			},
-			"mouseenter .ui-menu-item": function( event ) {
```
scottgonzalez: This method was moved to `_activateItem()`.

ui/widgets/menu.js
```diff
@@ -171,6 +152,35 @@ return $.widget( "ui.menu", {
...
+
+		// Ignore bubbled events on parent items, see #11641
+		if ( actualTarget[ 0 ] !== target[ 0 ] ) {
+			return;
+		}
+
+		// If the item is already active, there's nothing to do
+		if ( target.is( ".ui-state-active" ) ) {
+			return;
+		}
```
scottgonzalez: This conditional is the only thing that changed in this method. This is just the avoid the expensive work that follows.

PR #1806, opened 2017-04-18
Menu: Ignore mouse events triggered due to page scrolling

scottgonzalez:
Fixes #9356

### DIFF
--- a/tests/unit/selectmenu/core.js
+++ b/tests/unit/selectmenu/core.js
@@ -251,13 +251,13 @@ $.each( [
 			wrappers = menu.find( "li.ui-menu-item .ui-menu-item-wrapper" );
 
 			button.trigger( "click" );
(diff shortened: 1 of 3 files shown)


tests/unit/selectmenu/core.js
```diff
@@ -251,13 +251,13 @@ $.each( [
 			wrappers = menu.find( "li.ui-menu-item .ui-menu-item-wrapper" );
 
 			button.trigger( "click" );
-			wrappers.first().simulate( "mouseover" ).trigger( "click" );
+			wrappers.first().simulate( "mouseover", { clientX: 2, clientY: 2 } ).trigger( "click" );
```
scottgonzalez: This is hacky, but `simulate` doesn't fill in realistic values for anything other than drag. We should update simulate and remove this, but that seemed a little out of scope for this PR.

PR #1807, opened 2017-04-18
Menu: Close menus immediately on selection or click outside

scottgonzalez:
Fixes #15034

### DIFF
--- a/tests/unit/menu/events.js
+++ b/tests/unit/menu/events.js
@@ -670,7 +670,9 @@ QUnit.test( "handle keyboard navigation and mouse click on menu with dividers an
 		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
 		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
 		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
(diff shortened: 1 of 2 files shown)


tests/unit/menu/events.js
```diff
@@ -670,7 +670,9 @@ QUnit.test( "handle keyboard navigation and mouse click on menu with dividers an
 		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
 		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
 		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
-		assert.equal( logOutput(), "keydown,3,4,7", "Keydown focus skips divider and group label" );
+		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
+		element.simulate( "keydown", { keyCode: $.ui.keyCode.DOWN } );
+		assert.equal( logOutput(), "keydown,1,2,3,4,7", "Keydown focus skips divider and group label" );
```
scottgonzalez: This test changed because an item is selected in the step before this, causing the menu to close. This test only passed before because of the short delay; had it waited for the menu to fully react, the test would have failed.

PR #1810, opened 2017-04-23
make sure $.ui exists before writing to it

shmax:
Fix for https://bugs.jqueryui.com/ticket/15180

This change makes it possible to include ui/position.js directly in an application without the need to first include the core jqueryui file.


scottgonzalez:
This comes from the version component. Please see the ticket for more details.

PR #1816, opened 2017-05-11
Fixing Bug: Dialog with iframe while resizeing squeezes scrollbar

roopkt:
If dialog has an iframe , while resizing the dialog it is squeezing the scrollbar.  I could fix this issue while resizing sum is being calculated that sum calculation is not exactly working when iframe is there. 

Therefore, I deducted magical number 2.2 to apply correct sum.

### DIFF
--- a/ui/widgets/resizable.js
+++ b/ui/widgets/resizable.js
@@ -16,1192 +16,1202 @@
 //>>css.structure: ../../themes/base/resizable.css
 //>>css.theme: ../../themes/base/theme.css
 
-( function( factory ) {
-	if ( typeof define === "function" && define.amd ) {
-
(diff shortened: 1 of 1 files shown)


scottgonzalez:
+1,173 −1,163 is not an acceptable diff, please fix.

Also, magic numbers are unlikely to be approved.

roopkt:
Thanks for your review. Yeah, I know it is not a naive fix and even I hate magical hardcover numbers. But I just wanted to submit the issue with the quick fix that we did. 
I will continue looking at this at try to come up with solid fix.

scottgonzalez:
Closing due to inactivity. Please file a new PR if you continue working on this. Thanks.

PR #1817, opened 2017-05-11
Dialog: Fix shared event handler for modal dialogs

scottgonzalez:
The old logic worked when all widgets of the same type used the same
event namespace. However, now that each instance has its own namespace,
we cannot use `_on()` for shared event handlers.

Fixes #15182

### DIFF
--- a/ui/widgets/dialog.js
+++ b/ui/widgets/dialog.js
@@ -289,7 +289,7 @@ $.widget( "ui.dialog", {
 			that._trigger( "focus" );
 		} );
 
-		// Track the dialog immediately upon openening in case a focus event
+		// Track the dialog immediately upon opening in case a focus event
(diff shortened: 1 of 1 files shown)


jeroencornelissen:
@scottgonzalez what's the status on this? When is this (1.12.2) released?

sonyayu:
It should be an error message, because you are using Yahoo's email, please
delete the email! Thank you

Jeroen Cornelissen <notifications@github.com> 於 2018年10月1日 週一 19:49 寫道：

scottgonzalez:
I can’t speak on behalf of the project anymore as I have stepped down as project lead. I can say this project has not been maintained for quite a while, largely due to political fallout from the JS Foundation. I wouldn’t expect another release.

PR #1818, opened 2017-05-17
Calendar: Create helper methods for Date object creation

fnagel:
Extract all `new Date()` occurrences in order to simplify a possible fix for our unit test timezone issue.

### DIFF
--- a/tests/unit/calendar/helper.js
+++ b/tests/unit/calendar/helper.js
@@ -4,6 +4,17 @@ define( [
 ], function( $, helper ) {
 
 return $.extend( helper, {
+	createDate: function( year, month, day ) {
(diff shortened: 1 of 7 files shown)


tests/unit/calendar/helper.js
```diff
@@ -4,6 +4,17 @@ define( [
 ], function( $, helper ) {
 
 return $.extend( helper, {
+	createDate: function( year, month, day ) {
+		if ( arguments.length === 1 ) {
```
scottgonzalez: We shouldn't need to worry about this case for timezone correction, but having all dates go through the helper does make it easier to reason about in the tests.

PR #1819, opened 2017-05-26
handle null and undefined condition in the loop

dgofman:


dgofman:
Ticket: https://bugs.jqueryui.com/ticket/15192

scottgonzalez:
The existing code already handles `null` and `undefined`.

dgofman:
Your team should start using JSHint or JSLint I am getting many exceptions related comparison operators
 http://www.c-point.com/javascript_tutorial/jsgrpComparison.htm

arschmitz:
We do and it passes https://travis-ci.org/jquery/jquery-ui#L724. How are you linting these files are you not using our .jshintrc files? https://github.com/jquery/jquery-ui/blob/master/ui/.jshintrc, https://github.com/jquery/jquery-ui/blob/master/.jshintrc, https://github.com/jquery/jquery-ui/blob/master/tests/.jshintrc . We also use jscs asciilint csslint and htmllint.

dgofman:
I can see you are excluding validation of === what in my case always false.

// Whether `== null` comparisons should be allowed, even if `eqeqeq` is `true`.
"eqnull": false,

PR #1821, opened 2017-06-08
Datepicker: week number column title

dorinbujor93:
Adding title to datepicker week number column header.


scottgonzalez:
Thanks for doing the work to populate all that data. However, datepicker is undergoing a complete rewrite which uses CLDR for all locale data, which will be released in 1.13. Since this introduces a new API, it couldn't be released until 1.13 anyway, but at that point, all of this code would be gone.

If you're interested in the new datepicker, you can read about it in the [planning wiki](http://wiki.jqueryui.com/Datepicker) and see the current code in the [`datepicker` branch](https://github.com/jquery/jquery-ui/tree/datepicker/).

PR #1822, opened 2017-06-21
Datepicker: Fix erroneous update on setDate

georgeclaghorn:
Avoid updating the datepicker display when `setDate` is called for an instance that is not currently displaying.

Fixes [#6814](https://bugs.jqueryui.com/ticket/6814). This differs from the fix added in b4ef2f7e and reverted in 6e7bd4d89 in that it avoids modifying `_updateDatepicker`, which is used elsewhere.

### DIFF
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -578,7 +578,9 @@ $.extend( Datepicker.prototype, {
 		var inst = this._getInst( target );
 		if ( inst ) {
 			this._setDate( inst, date );
-			this._updateDatepicker( inst );
+			if ( inst === this._curInst ) {
+				this._updateDatepicker( inst );
+			}
 			this._updateAlternate( inst );
 		}
 	},


georgeclaghorn:
I'm not immediately sure why one test is failing. It looks like the datepicker widget is undergoing a rewrite, so I'm also unsure whether this is likely to be accepted even if I fix the failing test.

theosp:
Less elegant approach but, might pass tests, the issue is happening only when the datepicker is already showing, so no need to change the behaviour if it doesn't:

```
_setDateDatepicker = function(target, date) {
    var inst = this._getInst(target);

    if (inst) {
        this._setDate(inst, date);

        if ($.datepicker._datepickerShowing === true) {
            if (inst === $.datepicker._curInst) {
                // If a datepicker is already showing, set it to date only if it belongs
                // to the current target

                this._updateDatepicker(inst);
            }
        } else {
            this._updateDatepicker(inst);
        }

        this._updateAlternate(inst);
    }
};
```

PR #1823, opened 2017-06-27
bootstrap 3 options for styling

ICiscoStu:
this will allow a datepicker with option buttonGroup to be properly styled in bootstrap 3, it will add a span element that will contain the button as well as the proper classes to  style it as an input group element in bootstrap 3/4


scottgonzalez:
Thanks, but we're not adding anything to datepicker at this time. We're in the middle of a complete rewrite with a new API.

PR #1824, opened 2017-07-21
Xhosa datepicker translation.

stewartm83:
Please review


scottgonzalez:
Thanks, but datepicker is being rewritten to use the Unicode CLDR data for all localizations. As such, we're not accepting any new locale files.

PR #1826, opened 2017-08-04
check if parentNode exists before removeChild

babakhanov:
got rid of `Cannot read property 'removeChild' of null`


scottgonzalez:
When does this occur? We need a [ticket](https://bugs.jqueryui.com/newticket) with a reduced test case showing that a bug exists.

mannykary:
Any idea if this will be approved? I found this relevant ticket https://bugs.jqueryui.com/ticket/6054

md5:
I think the `if ( !this.cancelHelperRemoval ) {` should also be updated to `if ( !this.cancelHelperRemoval && this.helper ) {` to fix the scenario in the linked ticket.

dmethvin:
Closing as stale.

PR #1828, opened 2017-08-22
Fix autocomplete normalize, preserve empty value

Josh68:
In the normalize method, an item using a data object (label/value) with a value set to an empty string (useful, e.g., for an error response or an item in the list that should be disabled), the value is erroneously set to use the label string, because "item.value" is evaluated as false. Rewriting the normalize condition to check whether "item.value" is either truthy or an empty string.

### DIFF
--- a/tests/unit/autocomplete/options.js
+++ b/tests/unit/autocomplete/options.js
@@ -265,7 +265,7 @@ QUnit.test( "source, local object array, only labels", function( assert ) {
 		{ label: "java", value: null },
 		{ label: "php", value: null },
 		{ label: "coldfusion", value: "" },
-		{ label: "javascript", value: "" },
+		{ label: "javascript" },
 		{ label: "clojure" }
 	] );
 } );
--- a/tests/unit/autocomplete/remote_object_array_labels.txt
+++ b/tests/unit/autocomplete/remote_object_array_labels.txt
@@ -1,5 +1,5 @@
 [
 	{ "label": "java", "value": null },
-	{ "label": "javascript", "value": "" },
+	{ "label": "javascript" },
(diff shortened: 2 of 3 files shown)


scottgonzalez:
This is not behavior we want to change.

Josh68:
Do you think this can be explained better in the docs? I find it unexpected behavior. Thanks

scottgonzalez:
Feel free to send a PR: https://github.com/jquery/api.jqueryui.com/blob/master/entries/autocomplete.xml

Josh68:
I'd be happy to, but would like a brief rationale from your side before I submit something else that might get rejected. Thanks

scottgonzalez:
You cited two use cases that you're trying to solve: An error response and disabled items. Using an empty value for those cases is really just a hack. If you have an error that you'd like to report to the user, you should handle that inside your `source` option.

For disabled items, you can provide as many additional properties as you'd like on your item objects (This is likely what's missing from the docs). This allows you to provide a `disabled` property which would provide much more semantic value. You can then use the rendering extension points to take the `disabled` property into account.

Josh68:
Thanks for your explanations, Scott. I've cloned the docs repo, and will try to set myself up to add some commentary outlining what you've said.

It's true that handling service errors this way would be a hack, but the app I'm working on is already hack-central without good error handling. In any case, you're right about that.

I'm fine with the idea of using an additional property for indicating items that should be rendered as disabled, and I agree that should be added to the docs, so I'll do my best to contribute something to that effect. The current behavior should also be a little more fully documented, as you've laid out. 

Thanks for taking the time to discuss with me. 

Cheers,

Josh

PR #1832, opened 2017-09-06
Sortable: Add "fixed" options.

jhorowitz-firedrum:
The option "fixed.above" is a class whose presence
    prevents the sortable item from moving down when
    another sortable hovers over its top half.
The option "fixed.below" is a class whose presence
    prevents the sortable item from moving up when
    another sortable hovers over its bottom half.

### DIFF
--- a/ui/widgets/sortable.js
+++ b/ui/widgets/sortable.js
@@ -475,6 +475,17 @@ return $.widget( "ui.sortable", $.ui.mouse, {
 				) {
 
 					this.direction = intersection === 1 ? "down" : "up";
+					if ( typeof this.options.fixed !== "undefined" ) {
+						if ( this.direction === "down" ) {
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Thanks, but we don't accept new features without discussions. The proper way to request a feature is to [file a ticket](https://bugs.jqueryui.com/newticket) or post in the [developer forum](https://forum.jquery.com/developing-jquery-ui). The request must include the reason the feature is common enough that it belongs in the core of the plugin and not in an extension. Only after a solution has been agreed upon will a PR be accepted. At that point, the PR must include tests in addition to the implementation.

jhorowitz-firedrum:
Please see ticket 15236: https://bugs.jqueryui.com/ticket/15236#ticket

PR #1834, opened 2017-09-13
Pinpad: An interactive pinpad widget for point of sale (POS) or encrypted pinpad (EPP) simulation

yannickebongue:
The aim of this widget is to display a numeric pad with optionally a decimal point key and the three basic commands Cancel, Clear and Confirm.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -35,6 +35,7 @@ var
 		"dialog",
 		"draggable",
 		"menu",
+		"pinpad",
 		"progressbar",
 		"resizable",
 		"selectable",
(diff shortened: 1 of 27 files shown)


scottgonzalez:
This doesn't sound like something that belongs in jQuery UI.

yannickebongue:
I understand.
It's just a proposal for a new widget.
Is there a way to insert a new widget to the jQuery UI Framework?

scottgonzalez:
Generally you would start by discussing the widget on the [Developing jQuery UI forum](https://forum.jquery.com/developing-jquery-ui). Right now the focus is on rewrites of the interaction widgets with updates to support Pointer Events as well as updates for better globalization support. But something like pin pad would likely never be accepted because it's just not a common enough problem.

PR #1835, opened 2017-09-20
Draggable: Use relative coordinates for snapElements

uunicorn:
Since scrolling position may change while dragging, so are the absolute
coordinates of the snapElements.

Fixes #14694


dmethvin:
Is there a way to test this?

PR #1836, opened 2017-09-21
properly parse dates with 4 digit years when manually entered

haeflimi:
This solves a conversion problem that occured when manually entering 4 digit years as described here:
https://bugs.jqueryui.com/ticket/15239#comment:4


scottgonzalez:
We can't accept this because we will regress immediately after switching the parser to use Globalize.

haeflimi:
That's rather disappointing. - But i guess switching the whole date parsing to Globalize will fix the problem in the long run anyway.

Do you have any suggestions on how to work around this issue while we wait for this to become released reality?

In the Bugtracker Ticket according to this PR you have suggested to use maxLenght to work around the Problem. - As i've described here: https://bugs.jqueryui.com/ticket/15239#comment:4  this doesn't work however.

PR #1838, opened 2017-10-18
15129 previous button with multiple months and max date

chad56s:
Fix and tests for open bug trac 15129


chad56s:
I believe the CLA is now signed correctly

PR #1839, opened 2017-10-23
Add an .npmignore file to reduce weight of node_modules dependency

MrLeebo:
jquery-ui is the heaviest dependency in my node_modules folder at 14MB. It appears that this is coming from the demo and externals folders, which I don't think are intended to be part of the distribution.

I took a quick look at the folder structure and put together this .npmignore file, but I would appreciate it if someone more familiar with the project can double check that I didn't leave out anything that actually does need to be in the package.

https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package


dmethvin:
Agreed, but we need more discussion about what is truly needed. Let's consolidate under #1875.

PR #1841, opened 2017-10-25
Datepicker accessibility work.

NigelCunningham:
This branch improves the accessibility of the jQuery UI datepicker widget by replacing links with buttons and adding aria-pressed attributes to the date buttons.

The work was sponsored by the University of Southern Queensland and implemented by Salsa Digital.

### DIFF
--- a/tests/unit/datepicker/core.js
+++ b/tests/unit/datepicker/core.js
@@ -54,8 +54,8 @@ QUnit.test( "baseStructure", function( assert ) {
 			header = dp.children( ":first" );
 			assert.ok( header.is( "div.ui-datepicker-header" ), "Structure - header division" );
 			assert.equal( header.children().length, 3, "Structure - header child count" );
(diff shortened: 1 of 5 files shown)


fnagel:
Thanks for your contribution! However, datepicker is undergoing a complete rewrite which uses CLDR for all locale data, which will be released in 1.13. Since this introduces a new API, it couldn't be released until 1.13 anyway, but at that point, all of this code would be gone.

If you're interested in the new datepicker, you can read about it in the [planning wiki](http://wiki.jqueryui.com/Datepicker) and see the current code in the [`datepicker` branch](https://github.com/jquery/jquery-ui/tree/datepicker/).

ps: there is a pending PR for a11y improvements for the rewrite, see #1829 
Any feedback is very welcome!

NigelCunningham:
Closing, per @fnagel's comment above.

PR #1845, opened 2017-11-03
Droppable: Improve over/out events detection, when greedy option is used

dzmitry-paulenka:
Fixes #9389 and probably some others.
JSFiddle with the problem: http://jsfiddle.net/xhcofgvx/
JSFiddle with this fix: http://jsfiddle.net/xhcofgvx/3/

Obsoletes this PR: #1770

The logic changed to do two loops instead of one.
First one detects all droppables that should toggle over/out state, taking into consideration greedy option on inner droppables.
Second loop does the actual state change.

Also test is added, that fails without these commit.


Fantast:
Sorry, had to recreate PR to fix code style issues

PR #1847, opened 2017-11-15
Datepicker - selecting year improovement

PFight:
Replaced first and last year labels with '...' to make it clear, that user can select years earlier and later then yearRange.

image


dmethvin:
Closing as a stale branch. You can run tests and code checks locally to find issues before submitting a PR, see the README file.

PR #1848, opened 2017-12-17
Datepicker: Added Kabyle locale.

nabil509:
Added Kabyle locale for datepicker widget. Kabyle is a language of northern Algeria. Its ISO 639-2 code is "kab".


scottgonzalez:
Thanks, but the datepicker is being rewritten to use CLDR. As such, no new locales are being added.

PR #1849, opened 2018-01-13
Remove reference to .disableSelection()

gezquinndesign:
Apparently this is bad... http://api.jqueryui.com/disableSelection/


dmethvin:
See #1856

PR #1856, opened 2018-03-16
remove reference to .disableSelection() from demos

gezquinndesign:
.disableSelection() was deprecated in v1.9, I thought it should be removed from the demos default page.

http://api.jqueryui.com/disableSelection/


mgol:
Thanks!

PR #1862, opened 2018-06-14
Datepicker: Fix typo in currentText in da localization

OskarNS:
The term for "Today" in Danish is "I dag" not "Idag". This is a common misspelling in the Danish language. 
see https://dsn.dk/?retskriv=idag 
Dansk Sprognævn (DSN) is the authority over the Danish language.


mgol:
Thanks!

PR #1867, opened 2018-07-19
Rhayward msi patch 1

rhayward-msi:
This pull request fixes #15304 by adding an axisType option.  If it is "page", then pageX/Y will be used instead of clientX/Y.
and #15305 by not refreshing position on hidden elements.


scottgonzalez:
Thanks. However, adding new options to fix bugs is not acceptable. Code must work with any valid option configuration.

PR #1869, opened 2018-08-14
Slider: add support for crossing multiple handles

hsyl20:
With this patch, a slider which has both "range = true" and
"allowCrossingHandles = true" automatically determines the handlers
starting and stopping the range. Handlers can freely cross each other.


dmethvin:
Closing as a stale branch. You can run tests and code checks locally to find issues before submitting a PR, see the README file.

PR #1873, opened 2018-10-24
add locales for Afghanistan's Farsi (Dari) and Pashto

hamidsafdari:


scottgonzalez:
Thanks, but the datepicker is being rewritten to use CLDR. As such, no new locales are being added.

fnagel:
@hamidsafdari Things have changed. We'll adding localization files again. Please create a new PR and sorry...

PR #1875, opened 2018-10-26
add .npmignore

necccc:
Hi 👋

I’ve been doing a little research for a conference talk on how npm package size relates to their content, and your package was one in the several ones that were flagged by my scripts.  It has an outstanding weekly download count on npm and relatively large content that is not related directly to the package functionality.

I've added some files and folders to an .npmignore file, so they won't get packaged next time you release it on npm.

Thanks, and have a great day!

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,10 @@
+demos
+tests
+.editorconfig
+.jscsrc
+.jshintrc
+.mailmap
+.travis.yml
+Gruntfile.js
+.csslintrc
+.gitattributes


dmethvin:
I think we'll want to think through whether to exclude some of these. No argument at all about `demos` but some of the others might be useful to someone who isn't just taking the pre-built file.

fnagel:
The general idea makes sense to me, but I'm no npm package expert. 

@arschmitz @mgol Any opinions regarding this change?

mgol:
@fnagel This is a similar setup to what Core now uses. That said, I personally prefer declaring the `files` array in `package.json`; various tools may generate their own files inside of the repository main folder and it might be hard to always catch all of them.

fnagel:
Now that @arschmitz approved this: it would be great to include this in the next release!

@mgol are you fine with the content of the file? I personally would prefer an extra file for this, like git does.

@dmethvin You said "but some of the others might be useful to someone who isn't just taking the pre-built file". Which files do you have in mind here? `.jshintrc`, `.csslintrc` and `.gitattributes`?

mgol:
@fnagel In my experience, using a list of files/directories to _include_ as opposed of a list to exclude usually scales better. There is always a new IDE that will add its own directory that we wouldn't want to publish. On the other hand, we're using jquery-release here that always works with a fresh repo clone so it won't have this type of files/directories. So maybe there's no real risk here.

The list looks good to me. I haven't checked if we can exclude more but the current list looks good to exclude. Core also doesn't include Grunftfile or linter config files, or tests in its published bundle, compare the files/directories at https://github.com/jquery/jquery with https://unpkg.com/browse/jquery@3.5.1/.

fnagel:
@necccc One last thing before merging this: would you please change the commit message to "External: Add .npmignore file" or allow me to change your branch? Thanks!

fnagel:
@necccc I changed the commit message and merged the PR. Thanks for your contribution!

PR #1885, opened 2019-03-15
:focusable incorrectly requires tabindex on details and summary elements

bbenjamin:
I noticed this when opening a dialog with `<details>` elements that

contained no interactive elements other than `<summary>`.
Was not followed by any other interactive elements.
I was unable to access `<details> `with tab navigation as the focus was trapped at what it believed to be the last focusable element.
This is addressed by adding details and summary to the list of elements `$.ui.focusable` assumes tabbable.

### DIFF
--- a/ui/focusable.js
+++ b/ui/focusable.js
@@ -39,7 +39,7 @@ $.ui.focusable = function( element, hasTabindex ) {
 		return img.length > 0 && img.is( ":visible" );
 	}
 
-	if ( /^(input|select|textarea|button|object)$/.test( nodeName ) ) {
+	if ( /^(input|select|textarea|button|object|details|summary)$/.test( nodeName ) ) {
 		focusableIfVisible = !element.disabled;
 
 		if ( focusableIfVisible ) {


dmethvin:
@bbenjamin can you answer the question from @fuzzbomb ?

bbenjamin:
@dmethvin - @fuzzbomb  was correct, only `summary` is required. The PR was updated shortly after that based on his feedback. Looks like  I neglected to follow that up with a comment here.

In case it's of use to people with a similar problem - After realizing jQueryUI hadn't had a commit since 2017 I proposed an alternate solution for its use in Drupal https://www.drupal.org/project/drupal/issues/3038336 - the patch there provides the overrides necessary to correct this issue without having to alter jQueryUI.

dmethvin:
Thank you for the quick reply! We're trying to get a little love started here for jQuery UI, mainly bug fixes and compat with the latest jQuery version. This is a good one to add.

mgol:
This change requires a test as well.

Sorry for the late reply but I'm not really in the UI team, although I have commit access. If you're still interested in getting it landed, feel free to ping me when a test is added.

PR #1889, opened 2019-05-20
It is now possible to hide options

Eddcapone:
You can find an example here:
https://stackoverflow.com/a/56147978/4684797


Eddcapone:
It is unclear why the test failed.

PR #1891, opened 2019-06-26
Update .travis.yml

ghatwala:


dmethvin:
We could definitely use some help in updating the testing infrastructure, as you can tell it's not working quite right. If you'd like to help let us know!

PR #1895, opened 2019-08-09
Component: Add Droppable Tolerance Options

markliu2013:
Add droppable tolerance options, top-left-touch, top-right-touch, bottom-left-touch and bottom-right-touch.


markliu2013:
https://www.zfwhub.com/jquery-ui-draggable-to-smaller-droppable.html

PR #1896, opened 2019-08-11
add an alwaysShow option

hervegirod:
This ads an alwaysShow option to the jQueryUI autocomplete widget. What it does is that is always show the popup list if it is set to true (default is false). This can be useful to add for example things as a "full text" search in the popup, even if there are no results.

### DIFF
--- a/ui/widgets/autocomplete.js
+++ b/ui/widgets/autocomplete.js
@@ -42,6 +42,7 @@ $.widget( "ui.autocomplete", {
 	options: {
 		appendTo: null,
 		autoFocus: false,
+		alwaysShow: false,
 		delay: 300,
 		minLength: 1,
 		position: {
@@ -464,7 +465,7 @@ $.widget( "ui.autocomplete", {
 			content = this._normalize( content );
 		}
 		this._trigger( "response", null, { content: content } );
-		if ( !this.options.disabled && content && content.length && !this.cancelSearch ) {
+		if ( !this.options.disabled && (this.options.alwaysShow || (content && content.length && !this.cancelSearch)) ) {
 			this._suggest( content );
 			this._trigger( "open" );
 		} else {


hervegirod:
Diff:
diff --git a/ui/widgets/autocomplete.js b/ui/widgets/autocomplete.js
index 60d3265442..a94e0dc422 100644
--- a/ui/widgets/autocomplete.js
+++ b/ui/widgets/autocomplete.js
@@ -42,6 +42,7 @@ $.widget( "ui.autocomplete", {
 	options: {
 		appendTo: null,
 		autoFocus: false,
+		alwaysShow: false,
 		delay: 300,
 		minLength: 1,
 		position: {
@@ -464,7 +465,7 @@ $.widget( "ui.autocomplete", {
 			content = this._normalize( content );
 		}
 		this._trigger( "response", null, { content: content } );
-		if ( !this.options.disabled && content && content.length && !this.cancelSearch ) {
+		if ( !this.options.disabled && (this.options.alwaysShow || (content && content.length && !this.cancelSearch)) ) {
 			this._suggest( content );
 			this._trigger( "open" );
 		} else {

hervegirod:
That's strange, when I click on the licence agreement link, it is written that I have signed the CLA.

fnagel:
@arschmitz What do you think about this? Sounds like something we could add an example how to extend the widget (in this case the `__response` method) but not introduce a new option.

hervegirod:
> 
> 
> @arschmitz What do you think about this? Sounds like something we could add an example how to extend the widget (in this case the `__response` method) but not introduce a new option.
It would work I think!

arschmitz:
Yeah i don't think we should add an option for this there are many other ways to handle we could give examples of

fnagel:
Sorry @hervegirod we don't add another option here. Mainly as its a pretty minor change which could be done extending the widget.

PR #1893, opened 2019-08-01
Remove useless IE8 code, fixes bug.

joejoseph00:
Fixes a bug I found using views_load_more for Drupal with Chrome / Firefox latest versions.

PR #1897, opened 2019-08-20
Remove IE8 support that crashes in some cases (in my case using Drupa…

joejoseph00:
Remove useless IE8 support (causes problems in some use cases, particularly when using jQuery ui with the Drupal contrib module named "views_load_more" module.
History is here: https://github.com/jquery/jquery-ui/pull/1893
…l and the views_load_more module in combination with jQuery UI 12.1

### DIFF
--- a/ui/form.js
+++ b/ui/form.js
@@ -13,8 +13,9 @@
 // Support: IE8 Only
 // IE8 does not support the form attribute and when it is supplied. It overwrites the form prop
 // with a string, so we need to find the proper form.
+// However since we no longer need to support IE8 we can just return the form attribute.
 return $.fn._form = function() {
-	return typeof this[ 0 ].form === "string" ? this.closest( "form" ) : $( this[ 0 ].form );
+	return $( this[ 0 ].form );
 };
 
 } ) );


joejoseph00:
Yay, passing!

mgol:
I set the milestone to 1.14.0 it as we agreed not to remove IE 8 workarounds from 1.13 yet. If you'd like to land it sooner, we'd need a way to have the patch apply only when needed. However, the existing check already tries to achieve exactly that... So it'd be good to get more details about the crash: what exactly causes the crash, in which browsers, etc.

joejoseph00:
@mgol might as well wait for 1.14.0 .

with that said, IE8 is basically kaput for all intents and purposes.  The only people running it would be those not connected to the internet and they'd probably not be changing versions of jQuery.

joejoseph00:
I signed all the approvals now

mgol:
Sure, no problem. Take note, though, that we've released 1.13.0 not so long ago so I wouldn't expect 1.14.0 to arrive any time soon. We want to let it stay on 1.13.0 for a while, let folks update, discover & fix regressions from 1.12.1, etc.

mgol:
We'll be dropping all IE versions in 1.14.0 in a much larger PR. A part of it will be removing the `$.fn._form` API completely so this PR will not be needed anymore.

Thanks for your contribution anyway!

PR #1898, opened 2019-09-11
travis: test on node 12, ubuntu 18

nikolas:


mgol:
Our Travis config got significantly changed, it now uses more recent Node versions so this PR is obsolete. Closing, sorry!

PR #1900, opened 2019-10-03
pt-BR language file fixed

leofreitas:
The characters "greater than" and "less than" aren't required and can cause errors.

image

### DIFF
--- a/ui/i18n/datepicker-pt-BR.js
+++ b/ui/i18n/datepicker-pt-BR.js
@@ -14,8 +14,8 @@
 
 datepicker.regional[ "pt-BR" ] = {
 	closeText: "Fechar",
-	prevText: "&#x3C;Anterior",
-	nextText: "Próximo&#x3E;",
+	prevText: "Anterior",
+	nextText: "Próximo",
 	currentText: "Hoje",
 	monthNames: [ "Janeiro","Fevereiro","Março","Abril","Maio","Junho",
 	"Julho","Agosto","Setembro","Outubro","Novembro","Dezembro" ],


fnagel:
Hey @leofreitas, thanks for your PR and sorry for the delay. 
Can you elaborate on "can cause errors"? 


I noticed we have those sign (`&#x3C;`) in a few languages, but not in all:

No:
https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-hu.js#L16
https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-fr.js#L19
https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-en-NZ.js#L17

Yes
https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-de.js#L17
https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-es.js#L17
https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-ja.js#L17

We probably want to make this more consistent in general.

@arschmitz @jzaefferer  Do you have any idea whats the idea behind this?

leofreitas:
Hi @fnagel,
This sign (&#x3C;) is equal to <. I think this breaks the code because <is the beginning of an html tag. If you put this sign in Google Translator, it generates error.

fnagel:
@leofreitas The < and > are encoded, so this should never break any markup. Pasting HTML entities into Google Translate is not really a valid use case nor an error. Do you have a real world case where this breaks something?

Anyway, I would like to bring some consistency into the translations.

arschmitz:
i dont think we should change just this file and i'm not sure the reason for the inconsistency. However unless we can see an example of an actual error caused by this its probably not something to worry about just this second

mgol:
A larger version of this has landed for 1.13.2 in #2100 so this is obsolete now; sorry!

PR #1887, opened 2019-03-21
Use `expr.pseudos` instead of deprecated `expr.filters`

aTei:
https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryexprfilters-is-deprecated-use-jqueryexprpseudos

PR #1901, opened 2019-10-20
All: Migrate away from deprecated/removed Core APIs

mgol:
With this patch & the upcoming jQuery Color update the UI `master` branch passes more or less the same tests with jQuery `master` as with jQuery 3.4.1. Before it gets applied, it fails completely, hanging TestSwarm.


mgol:
I’d appreciate a review from @jquery/core. Commit by commit might be the easiest.

mgol:
Also, cc @arschmitz @fnagel

tests/unit/menu/helper.js
```diff
@@ -14,7 +14,7 @@ return $.extend( helper, {
 		if ( message === undefined ) {
 			message = lastItem;
 		}
-		log.push( $.trim( message ) );
+		log.push( String.prototype.trim.call( message ) );
```
gibson042: In my mind, this calls into question the replacement of `$.trim(…)` with `….trim()` in source files. Since the PR isn't intended to make any changes other than restoring functionality, perhaps we should be doing `String.prototype.trim.call(…)` everywhere (or defining a local `trim` function and using that, like you're doing with `isWindow`).

mattbrundage:
Pull request #1887 should be evaluated as well.

tests/unit/menu/helper.js
mgol: @gibson042 I changed the order of commits to make the `$.trim` changes the last one and modified all replacements to use `String.prototype.trim.call` instead, as you advised.

While an util would be easier to use, I don't see any infrastructure in UI to create internal utility modules that can be used in other files and this change is needed in various parts of the repository. Not wanting to expose anything new on `$.ui` (even if "internal" in theory), I went for `String.prototype.trim.call`.

mgol:
Landed in 98b539171b6e805fa79346a5e9896865e5213b9c.

melloware:
@mgol does this mean there is a Jquery UI release coming soon?  it has been almost 3 years since last release.

mgol:
@melloware While I have commit access to this repository, I'm a member of the jQuery Core team, not the jQuery UI team. My primary goal here is to make UI compatible with recent jQuery versions and to prepare it as much as possible for a future jQuery 4.0 release. In this way people using jQuery UI won't be blocked from jQuery Core updates. When those compatibility updates are done, we'll want to do a UI release.

I can't guarantee issues not related to Core compatibility will be addressed, though, that depends on whether people more familiar with UI internals review/merge PRs that touch them.

PR #1902, opened 2019-12-14
Fix: Adapt datepicker.js for some a11y standards

MIKOLAJW197:
Hi!
I am currently working on adapting application to work properlly with screen readers. 
My QA team pointed out that the datepicker.js component was not working correctly:
1. There wasn't information about currently selected day - fixed with `aria-current`
2. Selectpickers for choosing month/year wasn't labeled properly -  fixed with `aria-label`.

I fixed it locally in our project but I thought It will be nice to share my patch :)


fnagel (approved):
Looks good to me. Thanks for your contribution!

PR #1903, opened 2019-12-18
Datepicker: Fix Hungarian localization to use uppercase for first characters

zsoltii:
Apply Hungarian language rules on datepicker Hungarian translation.


fnagel:
@zsoltii Thanks for your contribution! 

So, this PR basically a `ucfirst()` for the control labels, right? Any idea why those were lower case until now?

fnagel:
For the record: asked a Hungarian friend and he approves this change.

PR #1905, opened 2019-12-30
Tests: Add the 3.x-git jQuery version to the dropdown

mgol:


ESRuth:
Does this mean that all 3.x migration errors should go away, soon?

mgol:
@ESRuth I'm working on it from time to time. That said, I have only so much time so I can't provide any ETA for when that's going to be finished.

PR #1883, opened 2019-03-04
Datepicker: Added Mongolian localisation

knutsenm:

PR #1907, opened 2020-01-10
Add de-at local

florianweg:


fnagel:
Hey @florianweg, thanks for your contribution! I'm not fully sure we're adding new localization files (see #1883) but I'm confident we will.

PR #1910, opened 2020-02-13
datepicker-bn.js for bengali language

Fathma:
Added Datepicker-bn.js for Bengali language support.


fnagel:
Thanks for your contribution! 

We'll need some review of the localization file, see https://jqueryui.pbworks.com/w/page/139157592/Adding%20translations

PR otherwise looks good. We'll need to squash the 4 commits into a single one with a proper commit message (something like "Datepicker: Add Bengali localization") but that's something I can do for you if needed.

Fathma:
Please change 4 commits into one proper commit. :) thank you @fnagel

PR #1916, opened 2020-05-13
Tests: Solve a frequent race condition in tests in Chrome/Safari

mgol:


mgol:
> I don't wanna know how you found this one.

Don't you? ;) I saw that the test was only failing every other run locally so I had a look at what it does. When it was failing, it was reporting `0` dimensions. Since this was an image and the test was synchronous, I thought the image may not be ready yet so I tried with cache on in various configurations and it was passing there. And, bingo!

PR #1919, opened 2020-05-13
All: Resolve most jQuery Migrate warnings

mgol:
This also adds a new internal `$.ui.__safeOffset__` method to provide a similar support to past jQuery versions while avoiding Migrate warnings.

### DIFF
--- a/ui/widgets/tooltip.js
+++ b/ui/widgets/tooltip.js
@@ -331,7 +331,7 @@ $.widget( "ui.tooltip", {
 					position( positionOption.of );
 					clearInterval( delayedShow );
 				}
-			}, $.fx.interval );
+			}, 13 );
 		}
 
 		this._trigger( "open", event, { tooltip: tooltip } );
--- a/demos/tabs/sortable.html
+++ b/demos/tabs/sortable.html
@@ -22,7 +22,7 @@
 			stop: function(event, ui) {
 				tabs.tabs( "refresh" );
 				if (previouslyFocused) {
-					ui.item.focus();
+					ui.item.trigger( "focus" );
 				}
 			}
 		});
--- a/tests/unit/button/core.js
+++ b/tests/unit/button/core.js
@@ -12,7 +12,7 @@ QUnit.test( "Disabled button loses focus", function( assert ) {
 	assert.expect( 2 );
 	var element = $( "#button" ).button();
 
-	element.focus();
+	element.trigger( "focus" );
 	setTimeout( function() {
 
 		assert.equal( element[ 0 ], $.ui.safeActiveElement( document ), "Button is focused" );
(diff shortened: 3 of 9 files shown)


ui/widgets/tooltip.js
```diff
@@ -331,7 +331,7 @@ $.widget( "ui.tooltip", {
 					position( positionOption.of );
 					clearInterval( delayedShow );
 				}
-			}, $.fx.interval );
+			}, 13 );
```
fnagel, line 334: Do we want a comment here why its 13?

ui/safe-offset.js
```diff
@@ -0,0 +1,25 @@
...
+
+		// Browser globals
+		factory( jQuery );
+	}
+} ( function( $ ) {
+return $.ui.__safeOffset__ = function( element ) {
+
+	// Simulate a jQuery short-circuiting when there are no client rects reported
+	// which usually means a disconnected node. This check in jQuery is meant just
+	// for IE but UI depends on it.
```
fnagel, line 16: Just for my understanding: this check is implemented in jQuery for IE only, bit UI needs it in every browser, right?

ui/widgets/tooltip.js
mgol: Good point. This is copied from the `$.fx.interval` value & follows the advice from https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryfxinterval-is-deprecated

ui/safe-offset.js
mgol: Not exactly. jQuery has a check meant for IE but it's run by all browsers so by pure coincidence it makes UI not hit errors here. This is about these lines:
https://github.com/jquery/jquery/blob/3.5.1/src/offset.js#L98-L104

mgol: How about:
```suggestion
	// for IE but is applied in all browsers & UI depends on it.
```
?

fnagel: I still don't understand why we need this in jQueryUI when its already in jQuery... sorry, my mind is a little slow today ;)

mgol: See https://jquery.com/upgrade-guide/3.0/#breaking-change-invalid-input-to-the-offset-method

jQuery 3.0.0 & newer require DOM elements with the `getBoundingClientRect` method. It's just that jQuery needs to run a check for `getClientRects()` due to the comment above the lines I linked, to workaround an IE issue. We shouldn't rely on this check for other purposes.

Also, jQuery Migrate warns if you use `offset()` in such a way.

That said, now I see the discussion in https://github.com/jquery/jquery/issues/2310 so maybe I need to re-read it & make sure Migrate is not warning too much.

mgol:
@fnagel There are some things to check for me in comments about offset changes. I didn't want to delay other changes so I extracted the second commit to #1922 & only landed the first one here.

ui/safe-offset.js
fnagel: Ahhh, got it. Thanks for the explanation and your patience :-)

mgol: @fnagel Thank you for raising these concerns! This change will most likely not be needed as we'll update Migrate instead.

PR #1921, opened 2020-05-15
Build: Update the version from 1.12.2-pre to 1.13.0-pre

mgol:


nagyimre1980:
is a new version expected?
Mainly because of jquery 4.0.0?
But due to many deprecated (e.g. trim, etc.) it would also be timely

PR #1924, opened 2020-06-11
Fixes for large data sets

fanfavorite:
- Don't bind the element if it already exists
- Use closest instead of parents and remove each for selectable mouseStart event


fnagel:
Any idea why the tests fail?

mgol:
Closing & re-opening to trigger GitHub Actions tests and the new CLA check...

PR #1925, opened 2020-06-19
Fix: CSP violation (style unsafe-inline) in resizable widget

Kwaadpepper:
This prevents triggering CSP violation. If we do not want to use 'unsafe-inine'.

### DIFF
--- a/ui/widgets/resizable.js
+++ b/ui/widgets/resizable.js
@@ -116,7 +116,8 @@ $.widget( "ui.resizable", $.ui.mouse, {
 		if ( this.element[ 0 ].nodeName.match( /^(canvas|textarea|input|select|button|img)$/i ) ) {
 
 			this.element.wrap(
-				$( "<div class='ui-wrapper' style='overflow: hidden;'></div>" ).css( {
+				$( "<div class='ui-wrapper'></div>" ).css( {
+					overflow: "hidden",
 					position: this.element.css( "position" ),
 					width: this.element.outerWidth(),
 					height: this.element.outerHeight(),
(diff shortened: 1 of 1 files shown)


melloware:
That is interesting this triggers CSP?   Do you know why it does?

dmethvin:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

Kwaadpepper:
I am using style-src: 'none'.
This small line forces me to use 'unsafe-inline'.
This is why. Also. It seems to me that there is no reason to mix inline-css and $.css

ui/widgets/resizable.js
```diff
@@ -116,7 +116,8 @@ $.widget( "ui.resizable", $.ui.mouse, {
 		if ( this.element[ 0 ].nodeName.match( /^(canvas|textarea|input|select|button|img)$/i ) ) {
 
 			this.element.wrap(
-				$( "<div class='ui-wrapper' style='overflow: hidden;'></div>" ).css( {
+				$( "<div class='ui-wrapper'></div>" ).css( {
```
dmethvin, line 119: This should be `overflow`.

Kwaadpepper: Nice, my bad. I made the change.

fnagel:
Moving it from inline to `$.css()` fixes this? Any idea why this makes a different for CSP?

> It seems to me that there is no reason to mix inline-css and $.css (css OM)

I agree

Kwaadpepper:
It's because js files are trusted (because csp policy allows it), therefore using css object model is allowed (manipulate css through the DOM). But online css via style attribute is unsafe.

This is the way I understand it.

fnagel:
@dmethvin Please review this so it can be merged.

PR #1930, opened 2020-06-24
Tests: Account for extra focus/blur listeners in jQuery >=3.4

mgol:
jQuery >=3.4.0 uses a special focus/blur handler pair
needed to fix various issues with checkboxes/radio buttons
as well as being able to pass data in focus triggers.
However, this leaves dummy focus & blur events if any of these
events were ever listened to at a particular element. There's not
a lot UI can do to fix this so we now just skip these handlers for
data comparisons in tests.

Ref jquery/jquery#4496

### DIFF
--- a/tests/lib/qunit-assert-domequal.js
+++ b/tests/lib/qunit-assert-domequal.js
@@ -151,10 +151,12 @@ function extract( selector, message ) {
 				return;
 			}
 
-			// We're going to do a shallow copy of the event list to
-			// avoid mutating internal jQuery structures when removing
-			// extraneous focus/blur handlers.
-			result.events[ eventType ] = result.events[ eventType ].slice();
+			// Only the special internal handlers
+			// have the namespace field set to boolean `false`;
+			// filter them out.
+			result.events[ eventType ] = result.events[ eventType ].filter( function( eventData ) {
+				return eventData.namespace !== false;
+			} );
 
 			eventDataList = result.events[ eventType ];
 			for ( i = eventDataList.length - 1; i > -1; i-- ) {
--- a/tests/lib/qunit-assert-domequal.js
+++ b/tests/lib/qunit-assert-domequal.js
@@ -93,6 +93,28 @@ function getElementStyles( elem ) {
 	return styles;
 }
 
+// Returns 0 if v1 == v2, -1 if v1 < v2, 1 if v1 > v2
+function compareVersions( v1, v2 ) {
+	var i,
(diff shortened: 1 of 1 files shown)


tests/lib/qunit-assert-domequal.js
```diff
@@ -113,6 +135,52 @@ function extract( selector, message ) {
...
+
+	// jQuery >=3.4.0 uses a special focus/blur handler pair
+	// needed to fix various issues with checkboxes/radio buttons
+	// as well as being able to pass data in focus triggers.
+	// However, this leaves dummy focus & blur events if any of these
+	// events were ever listened to at a particular element. There's not
+	// a lot UI can do to fix this so just skip these handlers for
+	// data comparisons in tests.
+	// See https://github.com/jquery/jquery/issues/4496
+	if ( result.events && jQueryVersionSince( "3.4.0" ) ) {
```
gibson042, line 147: Do we actually need to restrict this code to jQuery >= 3.4.0? What would happen without the check?

tests/lib/qunit-assert-domequal.js
```diff
@@ -113,6 +135,52 @@ function extract( selector, message ) {
...
+		var i, eventDataList, eventData;
+		$.each( [ "focus", "blur" ], function( index, eventType ) {
+			if ( !result.events[ eventType ] ) {
+				return;
+			}
+
+			// We're going to do a shallow copy of the event list to
+			// avoid mutating internal jQuery structures when removing
+			// extraneous focus/blur handlers.
+			result.events[ eventType ] = result.events[ eventType ].slice();
```
gibson042, line 157: Could we eliminate the following loop by doing the work here?
```suggestion
			// Only the special internal handlers
			// have the namespace field set to boolean `false`;
			// filter them out.
			result.events[ eventType ] = result.events[ eventType ].filter( function( eventData ) {
				return eventData.namespace !== false;
			} );
```

tests/lib/qunit-assert-domequal.js
mgol: I thought since this excludes some things from the compared objects, in older jQueries where there are no additional listeners we could perform a full check. There's always a risk that the workaround will not account for some real differences and by restricting it to jQuery >=3.4.0 we'd see a test error in older jQuery versions in such a case.

UI is currently tested with Core >=1.8.

tests/lib/qunit-assert-domequal.js
mgol: Ha, you're right! So much simpler.

PR #1931, opened 2020-07-15
Widget: fix handling of options === true under use strict in _show/_hide

Interghost:
When using jQuery UI under use strict, options === true (default for ie tooltip show/hide) will throw 
Uncaught TypeError: Cannot create property 'complete' on boolean 'true'
on line
options.complete = callback;

### DIFF
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -717,6 +717,8 @@ $.each( { show: "fadeIn", hide: "fadeOut" }, function( method, defaultEffect ) {
 		options = options || {};
 		if ( typeof options === "number" ) {
 			options = { duration: options };
+		} else if ( options === true ) {
+			options = {};
 		}
 
 		hasOptions = !$.isEmptyObject( options );


ui/widget.js
```diff
@@ -717,6 +717,8 @@ $.each( { show: "fadeIn", hide: "fadeOut" }, function( method, defaultEffect ) {
 		options = options || {};
 		if ( typeof options === "number" ) {
 			options = { duration: options };
+		} else if ( options === true ) {
```
dmethvin, line 720: When is `true` allowed for `options`? Is there a current test that fails without this? If not, could you add one?

Interghost: For me it was happening in tooltip with default show/hide option which is true
https://api.jqueryui.com/1.12/tooltip/#option-show
https://api.jqueryui.com/1.12/tooltip/#option-hide
this gets passed to the function.

Note that this is under "use strict" only, without use strict it just eats it up... I didn't run the tests, and I think adding one would be too much for me atm

dmethvin: Do you have a code example that shows the problem? It's still not clear to me how this would occur.

Interghost: https://jsfiddle.net/fgLoq65n/
Note I had to paste the entire UI in the JS editor so it will take some time to load, just be patient... after load open console and mouseover the input box

fnagel (approved):
+1 by reading

PR #1935, opened 2020-08-05
Datepicker: Fixed incorrect position on position:fixed elements

staaky:
Datepicker offsets are not calculated the same across methods. This causes incorrect positioning when a datepicker is attached to a bordered element within a position:fixed container after the page is scrolled.

I've changed `_showDatepicker` to use `outerHeight()` when measuring the input, so it matches the calculation in `_checkOffset`. This way `_checkOffset` no longer tries to compare two differently calculated offsets.

### DIFF
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -776,7 +776,7 @@ $.extend( Datepicker.prototype, {
 		}
 		if ( !$.datepicker._pos ) { // position below input
 			$.datepicker._pos = $.datepicker._findPos( input );
-			$.datepicker._pos[ 1 ] += input.offsetHeight; // add the height
+			$.datepicker._pos[ 1 ] += $( input ).outerHeight(); // add the height
 		}
 
 		isFixed = false;


chadrschroeder:
Thank you for making this change.  We're also seeing this issue when the browser zoom isn't at 100%, the page is scrolled down, and then a fixed position modal form is displayed which contains a datepicker.

For example, when the zoom is at 100% we'll have something like:

```javascript
// In _showDatepicker, input.offsetHeight is "36"
$.datepicker._pos[ 1 ] += input.offsetHeight; // add the height

// In _checkOffset, inst.input.outerHeight() is "36"
inputHeight = inst.input ? inst.input.outerHeight() : 0,

// In _checkOffset, this calculation works correctly and $(document).scrollTop() is subtracted
offset.top -= ( isFixed && offset.top === ( inst.input.offset().top + inputHeight ) ) ? $( document ).scrollTop() : 0;
```

When the zoom is at 80% we'll have something like:

```javascript
// In _showDatepicker, input.offsetHeight is "38"
$.datepicker._pos[ 1 ] += input.offsetHeight; // add the height

// In _checkOffset, inst.input.outerHeight() is "37.7344"
inputHeight = inst.input ? inst.input.outerHeight() : 0,

// In _checkOffset, this calculation doesn't work correctly and the datepicker is positioned too far down on the page
offset.top -= ( isFixed && offset.top === ( inst.input.offset().top + inputHeight ) ) ? $( document ).scrollTop() : 0;
```

PR #1937, opened 2020-09-25
Create datepicker-jp.js

khanamtechasia:
Japanese language for jquery datepicker.

### DIFF
--- a/ui/i18n/datepicker-jp.js
+++ b/ui/i18n/datepicker-jp.js
@@ -1,0 +1,37 @@
+/* Japanese initialisation for the jQuery UI date picker plugin. */
+/* Translated by Nguyen Kha Nam (khanamdev@gmail.com). */
+( function( factory ) {
+	if ( typeof define === "function" && define.amd ) {
+
+		// AMD. Register as an anonymous module.
+		define( [ "../widgets/datepicker" ], factory );
+	} else {
+
+		// Browser globals
+		factory( jQuery.datepicker );
+	}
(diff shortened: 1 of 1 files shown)


fnagel:
Thanks for your contribution!

We need some more reviews of the localization before merging this. Please see here for more info: https://jqueryui.pbworks.com/w/page/139157592/Adding%20translations

ui/i18n/datepicker-jp.js
```diff
@@ -0,0 +1,37 @@
...
+datepicker.regional.jp = {
+	closeText: "閉じる",
+	prevText: "&#x3C;前",
+	nextText: "次&#x3E;",
+	currentText: "今日",
+	monthNames: [ " 一月", "二月", "三月", "四月", "五月", "六月",
+	"七月", "八月", "九月", "十月", "十一月", "十二月" ],
+	monthNamesShort: [ "1月", "2月", "3月", "4月", "5月", "6月",
+	"7月", "8月", "9月", "10月", "11月", "12月" ],
+	dayNames: [ "日曜日", "月曜", "火曜日", "水曜日", "木曜日", "金曜日", "土曜日", "日曜日" ],
```
markvantilburg, line 24: I cannot read japanese. There are 8 dayNames here (according to google translate Sunday is there twice).

fnagel: Thanks for reviewing. You're right, that looks wrong. Even without speaking Japanese.

PR #1939, opened 2020-10-15
Core: Resolve csslint issues with the IE filter property

mgol:
IE filter property violates standard CSS parsing rules. Because of that, IE 8
added `-ms-filter` which accepts the same values but quoted:
https://web.archive.org/web/20101112132308/http://blogs.msdn.com/b/ie/archive/2008/09/08/microsoft-css-vendor-extensions.aspx
Both `filter` & `-ms-filter` serialize to the same `filter` property in
`currentStyle`.

This commit switches to that new property to resolve csslint issues.


fnagel (approved):

dharitdesai:
@mgol @fnagel Hi Guys, do we have any release eta on this 1.13 milestone ? Only reason for asking this question is, that we are having issues with jQuery Migrate and 1.12.1 doesn't have the fixes which master has.
Thanks!

mgol:
@dharitdesai There's no concrete release date, we need to fix all the tests & make a few other changes, then prepare a blog post & do a release. I cannot predict when exactly I'll have time to work on the remaining pieces or how long they'll take. That said, I feel we're closer to the end than to the beginning when I started submitting these patches (which was a bit more than a year ago).

PR #1940, opened 2020-12-14
#15262: Round calculated max before compare with max from options.

tkiseleva:
Fix for https://bugs.jqueryui.com/ticket/15262


tkiseleva:
I'm not sure that the failed test is the correct one.
The initial issue happens when in the setting I've set 8.003, but in js, it received as 8.0029999...

Or maybe provide some additional option if the max value should be covered or not.

tkiseleva:
I've returned main part as it was before, but added new options `includeMax`, if it's true, it will increase the max if it's not covered by range.
But tests are failed, I couldn't find how to fix, it something related to default options.

PR #1942, opened 2020-12-18
replaces jscs with eslint

PseudoNinja:
Component: Core Dependency Library
 

The JSCS project has been archived and now contains security flaws.  Migrated to ESLint to continue to provide linting service while alleviating the security concerns.
 
Fixes #15393


melloware:
Nice concise PR!

melloware (approved):

PseudoNinja:
Resolves bug [15393](https://bugs.jqueryui.com/ticket/15393)

eslint.json
```diff
@@ -0,0 +1,24 @@
...
+        "no-constant-condition": 1,
+        "no-control-regex" : 0,
+        "no-empty": 1,
+        "no-mixed-spaces-and-tabs": 1,
+        "no-redeclare": 0,
+        "no-shadow-restricted-names": 1,
+        "no-unexpected-multiline": 1,
+        "no-useless-escape": 1
+    }
+}
```
mgol, line 24: Please add an empty new line at the end of the file.

eslint.json
```diff
@@ -0,0 +1,24 @@
+{
+    "extends": "eslint:recommended",
```
mgol, line 2: UI should use the jQuery config, similarly as jQuery itself does: https://github.com/jquery/jquery-ui/pull/1946

mgol (commented):
Thanks for the PR. I think this needs some changes, though, and that may generate source changes that will be non-trivial. It's worth checking, though.

I'd expect this to remove JSHint as well.

eslint.json
PseudoNinja: added.

eslint.json
PseudoNinja: Im not sure we can get away from using the standardized eslint configuration file

PseudoNinja:
Went ahead and removed jshint from the package. Let me know if there are any other changes you would like to see

Gruntfile.js
```diff
@@ -176,25 +176,11 @@ grunt.initConfig( {
...
-		tests: {
-			options: {
-				config: true,
-				maximumLineLength: null
-			},
-			files: {
-				src: [ "tests/**/*.js" ]
-			}
-		}
+		target:["demos/**/*.js", "build/**/*.js", "ui/**/*.js"]
```
mgol, line 183: This doesn't follow the jQuery code style with regards to spacing; doesn't ESLint verify it?

eslint.json
```diff
@@ -0,0 +1,24 @@
+{
+	"extends": "eslint:recommended",
```
mgol, line 2: jQuery projects are not supposed to use this preset but the jQuery one. This is to me a prerequisite to the migration, we can't migrate to a preset that doesn't verify rules jQuery cares about.

mgol (changes requested):
The current preset is a blocker issue to me, unfortunately.

eslint.json
PseudoNinja: which preset would you recommend then?

mgol: The official jQuery one: https://www.npmjs.com/package/eslint-config-jquery

mgol:
I did the ESLint work in #1958, so I'm closing this PR.

PR #1945, opened 2021-02-19
Tests: Account for an extra noop focus/blur listener in jQuery >=3.4

mgol:
jQuery >=3.4.0 uses a special focus/blur handler pair needed to fix various
issues with checkboxes/radio buttons as well as being able to pass data in focus
triggers. This leaves extra focus & blur events if any of these events were ever
listened to at a particular element.

We've started skipping these handlers in the `domEqual` assertion in gh-1930 but
we missed a case where an event is triggered before any handler is attached -
jQuery >=3.4.0 attaches then an extra noop listener just to force the code path
to go through the setup code before the trigger happens. We now skip this extra
handler as well.

This fixes a test failure in "dialog: methods" destroy tests.

Ref jquery/jquery#4496
Ref gh-1930


fnagel (approved):
+1 by reading

PR #1946, opened 2021-02-21
Dialog: Workaround broken focus re-triggering in jQuery 3.4/3.5

mgol:
Focus re-triggering in jQuery 3.4/3.5 makes the original element
have its focus event propagated last, breaking the re-targeting.
Trigger focus in a delay in addition if needed to avoid the issue.

This fixes the "interaction between overlay and other dialogs" core dialog
test when tested against jQuery 3.4/3.5.

Ref jquery/jquery#4382


fnagel (approved):
+1 by reading

PR #1947, opened 2021-03-07
Tests: Make some number comparisons less strict

mgol:
Some of the APIs return fractional values in newer jQueries, making comparisons
sometimes not being 100% accurate. Allow some delta.

This is similar to what was already done in
98b539171b6e805fa79346a5e9896865e5213b9c but a few cases affecting IE and/or
Edge Legacy were missed.


fnagel (approved):
+1 by reading

PR #1949, opened 2021-03-07
Build: Add a new htmllint error regex

mgol:
Some bad pages are being detected as Catalan randomly which makes the build
fail. This is reproducible both locally & on Travis. This commit adds a new
regex to account for this error.


mgol:
You can see an example of this failure on Travis here: https://travis-ci.com/github/mgol/jquery-ui/jobs/488832293. I don't understand why it only happens on some builds...

fnagel (approved):
+1 by reading

PR #1950, opened 2021-03-09
Build: Fix the new htmllint error regex

mgol:
Some bad pages are being detected as Catalan randomly which makes the build
fail. This is reproducible both locally & on Travis. PR gh-1949 added a new
regex to account for this error but it didn't escape parens properly so it's
not matching the problematic error message.

Ref gh-1949

I tested the new regex against the error reported in https://travis-ci.com/github/jquery/jquery-ui/jobs/489208911 and it's matching.


fnagel (approved):
+1 by reading

PR #1951, opened 2021-03-14
Tests: Fix the "dialog: core: focus tabbable" test in IE

mgol:
In IE in jQuery 3.4+ a sequence:
```js
$( inputNode ).trigger( "focus" ).trigger( "blur" ).trigger( "focus" );
```
doesn't end up with a focused input. However, in this test we only want to check
that the last focused input receives the focus back when `_focusTabbable()` is
called which in reality doesn't happen so quickly so let's avoid the issue by
waiting a bit.

Ref jquery/jquery#4856


fnagel (approved):
+1 by reading

PR #1952, opened 2021-04-09
Tests: Accept delayed focusout in IE with jQuery 1.8

mgol:
In IE with jQuery 1.8 focusout may not happen immediately so some checks
need to be done later.


fnagel (approved):
+1 by reading

PR #1955, opened 2021-04-28
Position: Make sure `of` is treated as a CSS selector

mgol:


tests/unit/position/core.js
```diff
@@ -334,6 +351,7 @@ QUnit.test( "using", function( assert ) {
 			assert.deepEqual( position, expectedPosition, "correct position for call #" + count );
 			assert.deepEqual( feedback.element.element[ 0 ], elems[ count ] );
 			delete feedback.element.element;
+			delete feedback.target.element.prevObject;
```
mgol, line 354: `prevObject` is used for `.addBack()` so my change makes it different but it shouldn't matter.

fnagel (approved):

PR #1956, opened 2021-05-01
Build: create script to generate ESM-compatible code

eirslett:
The code, written in AMD module syntax, lives in the "ui"
folder.
A new Grunt task is added ("esmify") that will process every
JavaScript source file in the "ui" folder and convert the AMD
imports into ESM imports.
The transpiled code will be stored in the "esm" folder.

This means that consumers of jquery-ui can choose whether they
want to use the ESM version or the AMD version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 dist
+esm
 bower_components
 node_modules
 .sizecache.json
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -510,7 +510,7 @@ grunt.registerTask( "update-authors", function() {
 	} );
 } );
 
-grunt.registerTask( "default", [ "lint", "requirejs", "test" ] );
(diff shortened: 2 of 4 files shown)


mgol:
@fnagel what do you think?

eirslett:
Thanks for the reply! I think I will go ahead and just publish it as something like `@eirslett/jquery-ui-esm`.

> If we were to migrate to ESM and generate AMD, that would be something to think about but even then I'm not sure if we'd do it due to the above.

I tried doing that as well, what you could do is run the script I put in this PR and commit the results - it's jQuery-formatted and only affects the first couple of lines. Converting ESM to AMD is much simpler, so the gode generation script would be more manageable.

fnagel:
@mgol I agree we have very limited resources and should rather not add a new module system.

PR #1957, opened 2021-05-12
All: Drop $.ui.escapeSelector in favor of $.escapeSelector

mgol:
Fixes [#14991](https://bugs.jqueryui.com/ticket/14991)


fnagel (approved):
+1 by reading

PR #1958, opened 2021-06-02
Build: Migrate from JSHint & JSCS to ESLint

mgol:
The work is divided into three commits with (mostly, there are exceptions) the following work:
1. The actual switch to ESLint + fixes done by ESLint autofix.
2. Fixing the `strict` rule.
3. Manually fixing remaining violations.
4. Fixing a test issue caused by introducing strict mode.

Previously, while JSCS was only running in the build process in the command line, JSHint was also running in TestSwarm. There's no official ESLint version running in the browser so I opted to drop this feature; Core doesn't use it and I don't think it's important enough to hold off the switch to ESLint; you will get nice feedback when building in the terminal anyway. Please let me know if that's not acceptable for some reason.


fnagel:
>Previously, while JSCS was only running in the build process in the command line, JSHint was also running in TestSwarm. There's no official ESLint version running in the browser so I opted to drop this feature; Core doesn't use it and I don't think it's important enough to hold off the switch to ESLint

As long as the new ESLint runs with each build (and here in the GitHub / CI setup) I'm fine with the change.

tests/unit/tooltip/helper.js
```diff
@@ -14,10 +14,8 @@ define( [
 						timers = jQuery.timers;
 
 					jQuery.fx.stop();
-					var x = false;
```
fnagel, line 16: This looks like a breaking change. Is variable `x` not in use?

fnagel (approved):
+1 by reading

Looks good expect for one thing (the removed x variable). Amazing how many formatting issues have been fixed, even though there have been CGL tests in place...

tests/unit/tooltip/helper.js
mgol: It’s not used, that’s why ESLint complained. It’s only written to but never read.

fnagel: Ahh ok. Strange enough nobody noticed before...

PR #1959, opened 2021-06-22
UPDATE locale id on monthNames and monthNameShort

achmiral:
- "November" is the standard form based on Indonesian Dictionary
- Change "Agus" to "Agu" cause the other use 3 letters instead of 4 letters


fnagel:
Thanks for your contribution!

We need some more reviews of the localization before merging this. Please see here for more info: https://jqueryui.pbworks.com/w/page/139157592/Adding%20translations

mgol:
Also, signing our CLA is required.

PR #1960, opened 2021-06-23
jQuery 4 warning: Number-typed values are deprecated for jQuery.fn.css ("z-index", value)

nagyimre1980:
jquery-migrate-3.3.2.js: 100 JQMIGRATE: Number-typed values are deprecated for jQuery.fn.css ("z-index", value)

### DIFF
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -844,7 +844,7 @@ $.extend( Datepicker.prototype, {
 		if ( !inst.inline ) {
 			showAnim = $.datepicker._get( inst, "showAnim" );
 			duration = $.datepicker._get( inst, "duration" );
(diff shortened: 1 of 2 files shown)


mgol (commented):
Thanks for the PR! I made all Migrate warnings disappear when tested against the latest jQuery & jQuery Migrate versions but I haven't checked with the development jQuery version which is when this warning appears.

That said, I think it's a bug in Migrate. The intention was to not warn against CSS properties on which auto-suffixing was already not happening. We'll need another check in Migrate for this, I think.

mgol:
Actually, let me migrate that to a Migrate issue.

mgol:
Migrated to https://github.com/jquery/jquery-migrate/issues/438/

PR #1963, opened 2021-07-30
Build: Update dependencies passed to jquery-release

mgol:
The older versions were not compatible with Node.js 10 & newer.


melloware (approved):

mgol:
Note to self: wait with landing this until https://github.com/jquery/download.jqueryui.com/pull/580 lands so that we update `download.jqueryui.com` directly to a proper working release.

PR #1973, opened 2021-09-03
Effect: Define the jQuery variable before jQuery Color gets imported

mgol:
We need to create a local jQuery because jQuery Color relies on it and the
global may not exist with AMD and a custom build (trac-10199). This worked
in UI 1.12 but stopped in 1.13 as jQuery Color is now sourced as an AMD module
and the variable started being defined after jQuery Color code. To restore the
proper order, move the variable declaration to a separate small module loaded
before jQuery Color.

### DIFF
--- a/ui/effect.js
+++ b/ui/effect.js
@@ -23,6 +23,7 @@
 		// AMD. Register as an anonymous module.
 		define( [
 			"jquery",
+			"./jquery-var-for-color",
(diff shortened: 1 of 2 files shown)


melloware (approved):

fnagel (commented):
+1 by reading

PR #1984, opened 2021-09-24
Docs:Tests: Update the default jQuery version from 1.12.4 to 3.6.0

mgol:
1. Update the default jQuery version from 1.12.4 to 3.6.0 in tests
2. Update jQuery used for demos to 3.6.0. jQuery 1.12.4 misses some Core APIs that jQuery UI depends on and that are not polyfilled in demos.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -286,8 +286,8 @@ grunt.initConfig( {
 				"jquery-simulate/jquery.simulate.js": "jquery-simulate/jquery.simulate.js",
 				"jquery-simulate/LICENSE.txt": "jquery-simulate/LICENSE.txt",
 
-				"jquery/jquery.js": "jquery-1.x/dist/jquery.js",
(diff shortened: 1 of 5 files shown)


fnagel (commented):
Just seen this and I guess my #1986 PR is obsolete then. Much better solution. 

But is it safe to just override the jquery/jquery file? Any tests that rely on this version in this specific file? There is a jquery-1.12.4 folder though.

mgol:
> But is it safe to just override the jquery/jquery file? Any tests that rely on this version in this specific file? There is a jquery-1.12.4 folder though.

From what I've seen, this unversioned file is just used for demos and as a default jQuery version, e.g. for testing when you run `grunt`. Individual TestSwarm tests at https://swarm.jquery.org/project/jqueryui reach to specific Core subfolders like `jquery-1.12.4` or `jquery-3.6.0`, etc.

fnagel (approved):
Tested this locally, works as expected!

Issue #1998, opened 2021-10-22
Sortable Display Grid not working in 1.3


jwgda:
https://jqueryui.com/sortable/#display-grid

No longer drops into place.

mgol:
PR: #2008.

@borgboyone I'd appreaciate a quick review from you, at least of the source changes (I also added a test) if you have time. I also listed you as a co-author of the PR since source changes have been suggested by you. I hope you're fine with it.

mgol:
jQuery UI 1.13.1 including a fix for this issue has been released: https://blog.jqueryui.com/2022/01/jquery-ui-1-13-1-released/.

PR #2008, opened 2021-11-07
Sortable: Allow 0-height containers to be sortable as in 1.12.1

mgol:
Note that container specific events will not fire when the dragged element
is interacting with zero height containers.

Fixes gh-1998


mgol:
Ouch; I've just discovered the fix only allows the first drag; a second & subsequent ones are still broken just like before. I pushed a test modification that fails due to this. More changes will be needed...

mgol:
@borgboyone In addition to the changes you described, I also had to remove the `if` checking for `this.innermostContainer !== null` in https://github.com/jquery/jquery-ui/blob/1.13.0/ui/widgets/sortable.js#L891-L893.

This removed all occurrences of `this.innermostContainer` so I also removed setting the property.

tests/unit/sortable/core.js
```diff
@@ -42,4 +42,57 @@ QUnit.test( "ui-sortable-handle applied to appropriate element", function( asser
...
+		"	<li class='list-item-gh-1998'>Item 2</li>\n" +
+		"	<li class='list-item-gh-1998'>Item 3</li>\n" +
+		"</ul>" )
+		.sortable()
+		.appendTo( "#qunit-fixture" );
+
+	function step1() {
+		el.find( "li" ).eq( 0 ).simulate( "drag", {
+			dx: 100,
+			dy: 3,
```
mgol, line 73: @fnagel I'm not sure why `dy` matters here since we're dragging horizontally but it had a visible effect on the test... I'm not sure if it's an actual issue or just a jQuery Simulate quirk but in manual testing it seemed to work fine so I gave up on debugging this.

fnagel: Sorry, I'm not really helpful here, but I guess you are right: could be a jQuery Simulate quirk too.
Having some tests should be fine...

fnagel (approved):
+1 by reading

borgboyone:
@mgol - Smart move on removing the other checks for innermostContainer as this point.  FYI: As I see it, you'll want to leave L892...just remove L891 and L893.
> 
> @borgboyone In addition to the changes you described, I also had to remove the `if` checking for `this.innermostContainer !== null` in https://github.com/jquery/jquery-ui/blob/1.13.0/ui/widgets/sortable.js#L891-L893.
> 
> This removed all occurrences of `this.innermostContainer` so I also removed setting the property.

mgol:
> As I see it, you'll want to leave L892...just remove L891 and L893.

Yes, sorry for not being clear, I edited my comment.

mgol:

Issue #2001, opened 2021-10-27
draggable + sortable offset is wrong


barisusakli:
After upgrading from 1.12.1 to 1.13.0 dragging an item to a sortable causes the helper to go out of page when it is over the sortable. 

![](https://i.imgur.com/4Mzz0Gw.gif)

The code to create the draggable is below.

```
	$('#widgets .available-widgets .widget-panel').draggable({
			helper: 'clone',
			distance: 10,
			connectToSortable: '.widget-area',
		});
```
Same code works fine on 1.12.1. Feel free to close if this is already reported since I saw some other issues with draggable on 1.13.0

mgol:
I bisected it to c866e455373028a62a0956455a229fef63e91fac. PR that introduced the regression: #1793.

mgol:
Fix at #2009: restoring one removal from #1793 fixes the issue.

mgol:
jQuery UI 1.13.1 including a fix for this issue has been released: https://blog.jqueryui.com/2022/01/jquery-ui-1-13-1-released/.

gaojianzhuang:
I used the 1.13.1, I have another issue, when drag the element into the iframe and sortable, the cloned element offset change, but end sortable, the offset works again.
![rl6uh-c8xbw](https://user-images.githubusercontent.com/1866009/174605482-88d1e155-a137-44d8-98ba-569e3304414a.gif)

mgol:
@gaojianzhuang This issue is closed. If you have another similar one, please open a new issue and provide a runnable test case without third-party libraries.

knockout-componentsprost:
Found the answer—success!



https://stackdev.space/questions/jquery-draggable-library

PR #1793, opened 2017-02-16
Sortable Fix-Up

borgboyone:
Component: ui/widgets/sortable.js

* Created _scroll extension point and migrated scroll code from _mouseDrag (350-399, 411)
* Cleaned up logic for scrolled (412-417)
* Fixed appendTo functionality to match documentation (198-202, 333-338, 1108)
* Remove unnecessary function calls
* Move set-up position functions to appropriate place (340-344)
* Base scrollParent on placeholder and not helper (263-264, 1090-1091)
* Update scrollParent when switching containers (1090-1091)

PR #2009, opened 2021-11-07
Sortable: Fix positioning when moving a Draggable item into a Sortable

mgol:
PR gh-1793 removed setting `this.offset.parent` in the Draggable
`refreshPositions` method which broke position calculations when moving
a Draggable item into a connected Sortable. restore that assignment.

Ref gh-1793
Fixes gh-2001

### DIFF
--- a/tests/unit/sortable/options.js
+++ b/tests/unit/sortable/options.js
@@ -488,6 +488,63 @@ QUnit.test( "{ placholder: String } tbody", function( assert ) {
 	} );
 } );
 
+// gh-2001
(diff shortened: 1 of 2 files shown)


mgol:
@borgboyone I hope restoring this is safe but I can only guess as I can't know why you removed it in #1793. If you could say if it seems safe to restore to you, that'd be great!

fnagel (approved):
+1 by reading

Looks good to me and with having tests now, this should be good. Having feedback on why the code snippet was removed would be good though...

mgol:

PR #2021, opened 2021-11-22
Build: Switch from Travis to GitHub actions

fnagel:
Replaces Travis functionality.

**Follow up tasks:**

- [ ] Remove Travis / Github webhook


.github/workflows/test.yml
```diff
@@ -0,0 +1,58 @@
...
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
```
mgol, line 12: Should we try to rely more on defaults here, like in https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-a-list-of-events:
```suggestion
on: [push, pull_request]
```
In particular, the `edited` event seems to run when a review is edited if I understand it correctly. Docs are pretty scarce on what each event means, unfortunately...

.github/workflows/test.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: npm install
+
+      - name: Run lint
+        run: node_modules/.bin/grunt lint
+
+      - name: Run RequireJS
+        run: node_modules/.bin/grunt requirejs
+
+      - name: Run Qunit
+        run: node_modules/.bin/grunt test
```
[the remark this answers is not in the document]
fnagel: Ohh boy. Where is my mind... Added!

.github/workflows/test.yml
fnagel: I copied this from one of my projects and afair I did this so tags don't trigger a run as my project had much more jobs and was "expensive". Let's go with the default and change it if needed. I've changed this to your suggestion.

.github/workflows/test.yml
```diff
@@ -0,0 +1,50 @@
...
+          restore-keys: |
+            ${{ runner.os }}-node-${{ matrix.node-version }}-npm-
+            ${{ runner.os }}-node-${{ matrix.node-version }}-
+            ${{ runner.os }}-node-
+            ${{ runner.os }}-
+
+      - name: Install npm dependencies
+        run: npm install
+
+      #  Keep these steps in sync with the default command tasks in our Gruntfile!
```
fnagel, line 40: Added comment here

mgol (approved):
Looks good! Too bad we can't try it on the PR for some reason but let's get it merged and we'll see if that works.

Issue #2002, opened 2021-10-28
Mousemove on Autocomplete widget generates tons of cloned elements in hidden accessible


vitality82:
I thought I had a bug in my project after upgrading to jQuery 3.6.0 and jQuery UI - v1.13.0 but this is reproducible with downloading the full jQuery UI as of today:

Not just hovering, but any mouse move on any of the items in the autocomplete result list clones every item inside .ui-helper-hidden-accessible DOM element. Keep going and the debug console scrollbar starts to become very small.

Is this the intention? I have some performance concerns here over cloning so many elements in a hidden DOM, therefore polluting the DOM significantly.

<img width="1568" alt="CleanShot 2021-10-28 at 15 44 49@2x" src="https://user-images.githubusercontent.com/3256598/139269284-2b94e1a1-54a4-4887-9e0a-25aee7eca7a7.png">

mgol:
@vitality82 Thanks for the report. Do you see the same behavior with jQuery UI 1.12.1 or not?

vitality82:
No, I don't see dupes in the hidden accessible element, just one per each of the result items. Maybe they are triggered on a less sensitive event or perhaps additionally removed (this would be redundant), in any case I tested with: jQuery v1.12.4 on a live production app, and the class `ui-state-focus` is applied on hover of the list items in the original `ui-autocomplete` unordered list. I don't think this class is applied anymore. 

Hope this will help move the ball forward.

mgol:
> jQuery v1.12.4

The jQuery version shouldn't matter that much; I was asking whether the same issue appears when jQuery _UI_ 1.12.1 is used.

vitality82:
Sorry about the confusion. Prior to the upgrade I was using jQuery UI - v1.11.4 - 2015-05-14 and the same issue didn't show on that version.

mgol:
Thanks. Can you check if the issue is still there if you switch to jQuery UI 1.12.1?

vitality82:
You can reproduce it just by downloading jQuery-UI both v1.11.4 and v1.12.1 - there is the source of truth - and it looks like the issue also exists in v1.11.4 only a little bit different:

v1.11.4: one item inserted per hover (every time)
v.1.12.1: one item inserted per mouse move (often it inserts a lot more then on item due to the mouse move)

[...]

PR #2031, opened 2021-12-08
Autocomplete: Rewrite with a delay instead of appending the live region

bbenjamin:
Potential solution for https://github.com/jquery/jquery-ui/issues/2002. The bug is due to changes https://bugs.jqueryui.com/ticket/9357. This addresses the bug without undoing the fix reported there. Instead, it empties the live region instead of appending, and does so after a brief timeout so the live region isn't updated on every mousemove event or when quickly traversing through options

### DIFF
--- a/tests/unit/autocomplete/core.js
+++ b/tests/unit/autocomplete/core.js
@@ -345,11 +345,11 @@ QUnit.test( "ARIA", function( assert ) {
 						assert.equal( liveRegion.parent().length, 0,
 							"The liveRegion should be detached after destroy" );
 						ready();
-					}, 500 );
-				}, 500 );
-			}, 500 );
-		}, 500 );
-	}, 500 );
+					}, 110 );
+				}, 110 );
+			}, 110 );
+		}, 110 );
+	}, 110 );
 } );
 
 QUnit.test( "ARIA, aria-label announcement", function( assert ) {
@@ -374,7 +374,7 @@ QUnit.test( "ARIA, aria-label announcement", function( assert ) {
 		assert.equal( liveRegion.children().filter( ":visible" ).text(), "People : anders andersson",
 			"Live region changed on keydown to announce the highlighted value's aria-label attribute" );
 		ready();
-	}, 500 );
+	}, 110 );
 } );
 
 QUnit.test( "ARIA, init on detached input", function( assert ) {
--- a/tests/unit/autocomplete/core.js
+++ b/tests/unit/autocomplete/core.js
@@ -293,6 +293,7 @@ QUnit.test( "simultaneous searches (#9334)", function( assert ) {
 } );
 
 QUnit.test( "ARIA", function( assert ) {
+	var ready = assert.async();
 	assert.expect( 13 );
 	var element = $( "#autocomplete" ).autocomplete( {
 			source: [ "java", "javascript" ]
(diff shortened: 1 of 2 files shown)


ui/widgets/autocomplete.js
```diff
@@ -267,8 +268,11 @@ $.widget( "ui.autocomplete", {
 				// Announce the value in the liveRegion
 				label = ui.item.attr( "aria-label" ) || item.value;
 				if ( label && String.prototype.trim.call( label ).length ) {
-					this.liveRegion.children().hide();
-					$( "<div>" ).text( label ).appendTo( this.liveRegion );
+					clearTimeout( this.liveRegionTimer );
+					var that = this;
+					this.liveRegionTimer = setTimeout( function() {
+						that.liveRegion.html( $( "<div>" ).text( label ) );
+					}, 400 );
```
mgol, line 275: `400` is quite a big delay, can we use 100? See https://web.dev/rail/. Anything higher than 100 doesn't feel as immediate to the user.

tests/unit/autocomplete/core.js
```diff
@@ -308,43 +309,51 @@ QUnit.test( "ARIA", function( assert ) {
...
+							"Should be one child in the live region after the above" );
+						assert.equal( liveRegion.children().filter( ":visible" ).length, 1,
+							"Only one should be still visible" );
+						assert.ok( liveRegion.children().filter( ":visible" )[ 0 ] === liveRegion.children().last()[ 0 ],
+							"The last one should be the visible one" );
+						element.autocomplete( "destroy" );
+						assert.equal( liveRegion.parent().length, 0,
+							"The liveRegion should be detached after destroy" );
+						ready();
+					}, 500 );
```
mgol, line 348: This test will become very long with all these nested delays... However, if we can change the source delays to 100, maybe this can be reduced.

ui/widgets/autocomplete.js
bbenjamin: Changed to 100

tests/unit/autocomplete/core.js
bbenjamin: Changed these to 110, tests seem to be fine with 10ms of breathing room.

ui/widgets/autocomplete.js
```diff
@@ -267,8 +268,11 @@ $.widget( "ui.autocomplete", {
 				// Announce the value in the liveRegion
 				label = ui.item.attr( "aria-label" ) || item.value;
 				if ( label && String.prototype.trim.call( label ).length ) {
-					this.liveRegion.children().hide();
-					$( "<div>" ).text( label ).appendTo( this.liveRegion );
+					clearTimeout( this.liveRegionTimer );
+					var that = this;
+					this.liveRegionTimer = setTimeout( function() {
```
fnagel, line 273: Please try using `this._delay` here, see https://github.com/jquery/jquery-ui/blob/e90096e9dd25392118c2c578f490445870ced686/ui/widgets/autocomplete.js#L410-L411 for an example!

ui/widgets/autocomplete.js
```diff
@@ -663,8 +667,11 @@ $.widget( "ui.autocomplete", $.ui.autocomplete, {
 		} else {
 			message = this.options.messages.noResults;
 		}
-		this.liveRegion.children().hide();
-		$( "<div>" ).text( message ).appendTo( this.liveRegion );
+		clearTimeout( this.liveRegionTimer );
+		var that = this;
+		this.liveRegionTimer = setTimeout( function() {
```
fnagel, line 672: Please try using `this._delay`, too

fnagel (changes requested):

ui/widgets/autocomplete.js
bbenjamin: Looks like it works well, thanks for pointing that out!

ui/widgets/autocomplete.js
bbenjamin: Made the change here, too.

mgol (approved):

ui/widgets/autocomplete.js
```diff
@@ -267,8 +268,11 @@ $.widget( "ui.autocomplete", {
 				// Announce the value in the liveRegion
 				label = ui.item.attr( "aria-label" ) || item.value;
 				if ( label && String.prototype.trim.call( label ).length ) {
-					this.liveRegion.children().hide();
-					$( "<div>" ).text( label ).appendTo( this.liveRegion );
+					clearTimeout( this.liveRegionTimer );
+					var that = this;
```
fnagel, line 272: No need to use an extra variable as `this` is mapped when using `this._delay`. 

Sorry, I should have been more explicit about this in my previous review.

ui/widgets/autocomplete.js
```diff
@@ -663,8 +667,11 @@ $.widget( "ui.autocomplete", $.ui.autocomplete, {
 		} else {
 			message = this.options.messages.noResults;
 		}
-		this.liveRegion.children().hide();
-		$( "<div>" ).text( message ).appendTo( this.liveRegion );
+		clearTimeout( this.liveRegionTimer );
+		var that = this;
```
fnagel, line 671: Same here (see above)

fnagel (changes requested):

mgol:
I tested this locally and it works fine for me. I'll wait a bit for Felix to say whether he had any other concerns and I'll land it, making the changes requested by Felix by myself if @bbenjamin isn't available to do that.

mgol (approved):
Thank you for updating the PR!

fnagel (approved):

mgol:
Landed, thanks!

mgol:

PR #2036, opened 2021-12-23
Autocomplete: fix issue #2002 by rewriting live region after delay

theodoreb:
I know @bbenjamin is out until next year so adding the small change asked in #2031 in a new PR for the bug #2002 in case someone has time to look at it before the end of the year :)


mgol:
Thanks for the PR. However, we'll land gh-2031 to give the credit to the original PR author. If Felix's remarks aren't addressed soon, I'll do that by myself on the PR before merging.

theodoreb:
sounds good :) thanks!

Issue #2014, opened 2021-11-12
jquery autocomplete custom combobox- ui hangs


akashdeepjaiswal:
dropdown is too slow to render when data is close 5000. How to limit options like show first 100 data then add option like more data avaible then when clicked on this it bring another 100 and so on

mgol:
Thanks for the report. Does the issue you describe exist when jQuery UI 1.12.1 is used or only with jQuery UI 1.13.0 or newer?

noblesigma:
Even if it's less than 5000, it slows to crawl in the new one. Been using jQuery UI 1.12.1 for years, with a custom thing I wrote; albeit hacky, it does work in 1.12.1 and lags like crazy in 1.13.0 thus we cannot upgrade.

https://github.com/noblesigma/jQuery.UI.Combobox.Lazyload

dmethvin:
Can you provide a link to a simple test case? I'm curious what might have changed.

mgol:
Yes, a test case would definitely help.

fnagel:
This is about the https://api.jqueryui.com/autocomplete/ widget, right?

I don't think this widget is able to work with thousands of options. Not even hundreds. I would strongly recommend to use its AJAX filter functionality.

https://api.jqueryui.com/autocomplete/#option-source
https://jqueryui.com/autocomplete/#remote

noblesigma:
I will try to provide a test case soon. This is definitely a before I had it working now it doesn't work with the new version thing, sorry guys, I know it sucks I just haven't had much time. The github example admittedly is a bit old and it doesn't self host work again, so it will take me some time to make you a new demo.

noblesigma:
https://github.com/noblesigma/jQuery.UI.Combobox.Lazyload/tree/master/example/breakingExample

Sorry it took so long. Good luck. Please let me know if you need help with something.

mgol:
Thanks for the test case! That helps a lot.

I did a performance profile on both 1.12.1 & 1.13.0 test cases and it looks like the performance regression is caused by ef2e9bab92ae898311baa295590cd487d9071319. The added `bindRemoveEvent` function triggers `this._on` calls for all of those elements which takes a lot of time.

fnagel:
@noblesigma Thanks for the test case, I now understand what's the issue here!

@mgol Looks like an ugly bug here as Scott's commit fixed a few bugs. Seems like a complicated one. Not sure what to do :-/

mgol:
[...]

PR #2037, opened 2021-12-31
Widget: Optimize attachment of the _untrackClassesElement listener

mgol:
jQuery UI 1.13.0 changed the logic attaching the `_untrackClassesElement`
listener in the `_classes` widget method; one of the side effects was calling
`this._on` for each node that needed the listener. That caused a severe
performance degradation for large comboboxes as each `_on` jQuery UI call
causes a jQuery `add` call that calls Sizzle's `uniqueSort` underneath.

Instead, collect the nodes that need the listener and then, outside of the loop,
create a jQuery object out of them and attach the listener once. That's still
slower than the jQuery 1.12 version but only slightly: 936 ms to 1.03s on a very
large list on a recent MacBook Pro, compared to ~30 seconds before this patch.

Fixes gh-2014

### DIFF
--- a/tests/unit/autocomplete/core.js
+++ b/tests/unit/autocomplete/core.js
@@ -443,4 +443,27 @@ QUnit.test( "Close on click outside when focus remains", function( assert ) {
 	} );
 } );
 
+QUnit.test( "extra listeners created during typing (trac-15082, trac-15095)", function( assert ) {
+	assert.expect( 2 );
+
+	var origRemoveListenersCount;
+	var element = $( "#autocomplete" ).autocomplete( {
+		source: [ "java", "javascript" ],
+		delay: 0
+	} );
+
+	element.val( "j" ).autocomplete( "search", "j" );
+	origRemoveListenersCount = jQuery._data( element[ 0 ], "events" ).remove.length;
+
+	element.val( "ja" ).autocomplete( "search", "ja" );
+	assert.equal( jQuery._data( element[ 0 ], "events" ).remove.length,
+		origRemoveListenersCount,
+		"No extra listeners after typing multiple letters" );
+
+	element.val( "jav" ).autocomplete( "search", "jav" );
+	assert.equal( jQuery._data( element[ 0 ], "events" ).remove.length,
+		origRemoveListenersCount,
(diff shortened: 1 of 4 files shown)


mgol:
I'm not sure how to verify whether this doesn't break Scott's fixes from ef2e9bab92ae898311baa295590cd487d9071319 as that patch didn't add any test. @fnagel would you be able to test it a bit?

fnagel:
Some more references for the commit (ef2e9bab92ae898311baa295590cd487d9071319) that introduced the bug this PR fixes:

- https://github.com/jquery/jquery-ui/pull/1769
- https://bugs.jqueryui.com/ticket/15082
- https://bugs.jqueryui.com/ticket/15095
- https://bugs.jqueryui.com/ticket/15078
- https://bugs.jqueryui.com/ticket/15136
- https://bugs.jqueryui.com/ticket/15152

fnagel (approved):
Code and its logic loos good to me and makes sense.

fnagel:
I can do some manual testing. It is too bad there are no tests for this, but what about adding some now? Or is it tricky to search for a memory leak / performance issue?

mgol:
@fnagel I guess we could add a test for some of these, e.g. https://bugs.jqueryui.com/ticket/15082 as we can access our event handlers list and see if it doesn't grow too much. I'll think about possible tests next week.

fnagel:
Wonderful. I will do some more testing when you finished.

mgol:
@fnagel I added tests for all of the issues except for https://bugs.jqueryui.com/ticket/15136 but that one will be almost identical to the existing tests so it should be good for you to test now.

I verified those new tests fail when Scott's change is reverted and - as you can see - they pass on my branch.

mgol:
@fnagel I added a test case for the `tabs` leak as well. As for other tests, it fails if I revert Scott's patch and succeeds on my branch.

This should be ready for your final testing.

fnagel (approved):
Tested locally (reviewed in IDE, running tests and manual testing incl. the demo from [issue](https://github.com/jquery/jquery-ui/issues/2014#issuecomment-992522694)). Good to merge!

mgol:

PR #2050, opened 2022-02-11
Datepicker: removed arrow symbols from next and previous month hovertext in swedish

### LINKED ISSUE #2048: Datepicker, Next month button tooltip
In some languages there is some text after the word "Next" in whatever language. In my case it is swedish and the tooltip is this: Nästa&#xBB;

Added a picture of the same problem in chinese and for replication you can check here:
https://jqueryui.com/datepicker/#localization

d3dfef2e-09ff-4d3d-af6a-4cc56dee3ed3

adalid-1:
Fixes #2048


adalid-1:
Commit is made from work git account and PR is made with private git account.

mgol:
@fnagel any idea why these symbols were included in some language versions, leading to issues like #2048, and not in others, like en-GB?

adalid-1:
@mgol I rolled back the local branch and recomitted the change with correct author and did a forced push. I am not great with git so please be wary if you see something wierd. If so I can remake the pullrequest from scratch after the weekend.

mgol:
Yeah, I'd also opt into removing all of those prefixes/suffixes. @adalid-1 would you like to do it for all the locales?

adalid-1:
@mgol Sure thing!

ui/i18n/datepicker-bs.js
```diff
@@ -17,8 +17,8 @@
 
 datepicker.regional.bs = {
 	closeText: "Zatvori",
-	prevText: "&#x3C;",
-	nextText: "&#x3E;",
+	prevText: "",
```
adalid-1, line 20: hmm

ui/i18n/datepicker-hr.js
```diff
@@ -17,8 +17,8 @@
 
 datepicker.regional.hr = {
 	closeText: "Zatvori",
-	prevText: "&#x3C;",
-	nextText: "&#x3E;",
+	prevText: "",
```
adalid-1, line 20: empty

ui/i18n/datepicker-mk.js
```diff
@@ -17,8 +17,8 @@
 
 datepicker.regional.mk = {
 	closeText: "Затвори",
-	prevText: "&#x3C;",
-	nextText: "&#x3E;",
+	prevText: "",
```
adalid-1, line 20: empty

adalid-1:
A few languages turned up empty when removing arrow. Maybe that is ok?

adalid-1:
Do I need to do anything more here or can I just wait for someone to review and merge this?

mgol:
Hmm, I wonder if this is really behaving in the same way as 1.12.1 did. As `prevText` & similar were documented to accept text, not HTML but were in fact accepting HTML, we switched that to only accept text as a security fix. It looks like this is something that broke those arrows... @adalid-1 are you sure the behavior in 1.12.1 was the same?

Making the texts empty may be bad for accessibility, potentially worse than at least having an arrow (as long as it works, of course). Maybe we should leave the arrows - but fixed, inserted as Unicode, not HTML entities - in those languages that would otherwise have empty `prevText`, etc.? What do you think, @fnagel?

adalid-1:
You are right it works in the older version. I was misstaken, I must have made some changes that did not come through when testing earlier. Now that i tested downloading the older version from here:
https://jqueryui.com/download/
And setting nextText to contain the arrow it came out as intended.

fnagel:
> As prevText & similar were documented to accept text, not HTML but were in fact accepting HTML, we switched that to only accept text as a security fix.

Sounds reasonable. Anyway, a fix would be good.


> A few languages turned up empty when removing arrow. Maybe that is ok?
> Making the texts empty may be bad for accessibility, potentially worse than at least having an arrow (as long as it works, of course).
 
We should definitely not have those texts empty. I'm ok with using the arrows but having a real text would be better. I could add a commit introducing those missing texts translated by Google Translate (and verified by another). @mgol What do you think?

fnagel:
We might want to merge #2067 before working on this. This one seems related: #1900

mgol:
This landed in 3126e128691fd9fa78f33930ba9cb37609d0a64f, thanks!

PR #2053, opened 2022-02-17
Selectmenu: Remove a call to the deprecated .focus() method

bmode:
updated call to deprecated jQuery .focus() method in selectmenu.js which was triggering a warning from the migrate plugin


melloware (approved):

fnagel:
@mgol Just for my curiosity: why is the deprecation not mentioned here? https://api.jquery.com/focus/

fnagel (approved):

mgol (approved):
Thank you!

mgol:
@fnagel we haven't documented event shorthand deprecation yet... It's a bit non-standard as those pages also document the details about specific events which we'd like to keep.

PR #2057, opened 2022-03-04
fix offset top calculation

SailorMax:
in _showDatepicker() we use native input.offsetHeight
but in _checkOffset() we use jquery's input.outerHeight()

jquery's method returns more accurate height for zoomed out pages. In result _checkOffset()'s `offset.top === ( inst.input.offset().top + inputHeight )` never returns true

### DIFF
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -815,7 +815,7 @@ $.extend( Datepicker.prototype, {
 		}
 		if ( !$.datepicker._pos ) { // position below input
 			$.datepicker._pos = $.datepicker._findPos( input );
-			$.datepicker._pos[ 1 ] += input.offsetHeight; // add the height
+			$.datepicker._pos[ 1 ] += inst.input.outerHeight(); // add the height
 		}
 
 		isFixed = false;


fnagel:
Thanks for the report. Does the issue you describe exist when jQuery UI 1.12.1 is used or only with jQuery UI 1.13.0 or newer?

SailorMax:
the issue exist when jQuery UI 1.12.1 is used also

mgol (changes requested):
Thanks for the PR. We'll need tests for this change, though. Have a look at existing test in https://github.com/jquery/jquery-ui/tree/main/tests/unit/datepicker to see how to write one.

mgol:
Also, we'll need you to sign the CLA to proceed.

SailorMax:
> Also, we'll need you to sign the CLA to proceed.

I'll look later about test. But not ready to sign anything because of 1 tiny patch. You can merge this fix from your name.

mgol:
I understand your position but we're obliged to require a CLA signature to accept contributions. If you're not willing to sign the CLA then it's better for someone else to look at it so that ownership is clear. Together with tests this would be much larger than just a single line changed.

Given the above, I'm going to close this, sorry!

Issue #2011, opened 2021-11-09
Pre-built version of jquery-ui


MattIPv4:
Hey folks,

We're trying to update jquery-ui on cdnjs, but it looks like this repo as well as the NPM package only distribute jquery-ui in its assorted parts, rather than shipping a pre-built JS bundle like you get if you use the code.jquery.com CDN or the download option on the site.

Is there a distributed version that is pre-built available in an NPM package or tagged git repository anywhere that we can use as the update target for cdnjs? Or, should we just be hosting the content of UI & themes from the NPM package / git repo on the CDN?

Cheers, Matt.

(cc https://github.com/cdnjs/packages/issues/889)

mgol:
Hey! We distribute the built files to CDNs and share zipped copies of the bundles on the jQuery UI sites. We're not planning any changes here as the project is mostly in maintenance mode.

You can download the full source & theme packages by using the quick download links from [the jQuery UI Download Builder page](https://jqueryui.com/download/) or the [1.13 blog post](https://blog.jqueryui.com/2021/10/jquery-ui-1-13-0-released/).

I hope that's enough for you. Thanks!

MattIPv4:
Unfortunately, downloading from the site won't work for cdnjs -- we either need the files in an NPM package, or in a tagged git repo.

Many users rely on jQuery UI through cdnjs, so it'd be great if y'all could help out here so that we can get the updated version made available to them :)

mgol:
How do you handle older jQuery UI versions? I think none of them includes built files, don't they?

MattIPv4:
Looking at our config, we've never updated this library under the new system -- under the old system folks would manually commit files but we no longer permit that for an assortment of reasons (mainly security + infra).

glensc:
you could perhaps create a branch or tag (jquery-ui-1.13.0.cdn) and put built files there for cdnjs...

umer936:
Any chance this is going to happen?

mgol:
Alright, it might be that it's just a question of copying two files from the `dist/cdn` folder before [this line in jquery-release](https://github.com/jquery/jquery-ui/blob/54074fcb1d228ecb06d95362f8db57293d3ba8f4/build/release.js#L134). Since cdnjs is not able to consume what we have right now, let me try this.

If this works (I still need to try it), we could release it as `1.13.2` when it arrives.

mgol:
(please don't assume it will happen just yet, I am not 100% sure yet if this works)

[...]

PR #2059, opened 2022-03-04
Build: Include all the files published to the CDN in npm/Bower packages

mgol:
Both unminified & minified versions of JS & CSS files are included, for all
themes, as well as images.

Fixes gh-2011

### DIFF
--- a/build/release.js
+++ b/build/release.js
@@ -131,6 +131,20 @@ Release.define( {
 			shell.mkdir( "-p", "dist/cdn" );
 			shell.cp( tmpFolder + "/jquery-ui*.js", "dist/cdn" );
 			shell.cp( "-r", tmpFolder + "/themes", "dist/cdn" );
+
+			// Copy all the files to be published on the CDN to the dist directory
+			// as well.
+			shell.cp( "dist/cdn/jquery-ui.js", "dist" );
+			shell.cp( "dist/cdn/jquery-ui.min.js", "dist" );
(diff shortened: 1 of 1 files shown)


mgol:
> I'm not aware how the release script works so not a really reliable review.

The main thing here is that in the `jquery-ui` package there will be a new `dist` folder that will contain everything we upload to the CDN; if this was included in the `1.13.1` release then dist in the npm package would contain everything from [https://github.com/jquery/codeorigin.jquery.com/tree/main/cdn/ui/1.13.1](https://github.com/jquery/codeorigin.jquery.com/tree/main/cdn/ui/1.13.1?rgh-link-date=2022-03-04T23%3A38%3A28Z), maintaining the directory structure.

Are you OK with that? The current `jquery-ui` package has no `dist` folder.

mgol:
@fnagel can you say if you give a green light to the above?

fnagel:
@mgol Just for my understanding, we then have those packages:

* Github tag (containing everything, incl. the new `dist` folder) https://github.com/jquery/jquery-ui/releases/tag/1.13.1
* A minimal CDN version https://github.com/jquery/codeorigin.jquery.com/tree/main/cdn/ui/1.13.1
* A minimal download builder version https://jqueryui.com/resources/download/jquery-ui-1.13.1.zip

The only new thing is that the Github tag / release download then includes the processed files, right?

fnagel (approved):

mgol:
@fnagel yes, my understanding is the same.

PR #2061, opened 2022-03-23
fixing "height" everytime "auto"

ricko99m:
now the height is set by options
just a simple fix


mgol:
Thanks for the contribution. However, we need:
* A detailed description of the issue this PR is supposed to fix. Ideally, submit a separate issue and link to it from here.
* Information about jQuery UI versions affected.
* A test case added to this PR.

Issue #2062, opened 2022-03-28
Performance issue in sortable widget with long lists


naitsirch:
There is a huge performance issue in the `sortable` widget which started in version (`1.12.1`) and is still existing in *current* version `1.13.1`, when used with large lists.

There's already a ticket for this issue in the old tracker: https://bugs.jqueryui.com/ticket/15097 and a corresponding [stackoverflow question](https://stackoverflow.com/q/20398400/1119601).

Today I have created some new fiddles to show the issue. Please compare the console output in the fiddle:

| jquery-ui version | fiddle url | output | time of initialization in seconds |
| ----------------- | ---------- | ------ | --------------------------------- |
| 1.11.4            | https://jsfiddle.net/m0j61Lnw/0/ | "#1", "Mon Mar 28 2022 12:24:40 GMT+0200 (Mitteleuropäische Sommerzeit)" | 0 |
|  |  | "#2", "Mon Mar 28 2022 12:24:40 GMT+0200 (Mitteleuropäische Sommerzeit)" |  |
| 1.12.0            | https://jsfiddle.net/m0j61Lnw/2/ | "#1", "Mon Mar 28 2022 12:15:08 GMT+0200 (Mitteleuropäische Sommerzeit)" | 2 |
|  |  | "#2", "Mon Mar 28 2022 12:15:10 GMT+0200 (Mitteleuropäische Sommerzeit)" |  |
| 1.12.1            | https://jsfiddle.net/m0j61Lnw/3/ | "#1", "Mon Mar 28 2022 12:15:46 GMT+0200 (Mitteleuropäische Sommerzeit)" | 16 |
|  |  | "#2", "Mon Mar 28 2022 12:16:02 GMT+0200 (Mitteleuropäische Sommerzeit)" |  |
| 1.13.1            | https://jsfiddle.net/m0j61Lnw/1/ | "#1", "Mon Mar 28 2022 12:23:11 GMT+0200 (Mitteleuropäische Sommerzeit)" | 16 |
|  |  | "#2", "Mon Mar 28 2022 12:23:27 GMT+0200 (Mitteleuropäische Sommerzeit)" |  |

There are some solutions suggested in the stackoverflow answeres and in the old ticket.

It would be great if you could fix the issue. Otherwise we are stuck on version 1.11.4 :-(

naitsirch:
**Update**:
I can aprove that this workaround significantly reduces the performance issue:

```javascript
[...]

PR #2063, opened 2022-03-31
sortable: Use addClass in _setHandleClassName to improve performance

naitsirch:
This addresses #2062 and is relevant for large lists with more than 1000 items.

The solution to the issue was suggested at [stackoverflow.com](https://stackoverflow.com/a/46925710/1119601) (credits to @Ernst).
At the end this commit restores the function's behaviour of previous 1.11.x version.

### DIFF
--- a/ui/widgets/sortable.js
+++ b/ui/widgets/sortable.js
@@ -118,19 +118,19 @@ return $.widget( "ui.sortable", $.ui.mouse, {
 	},
 
 	_setHandleClassName: function() {
-		var that = this;
 		this._removeClass( this.element.find( ".ui-sortable-handle" ), "ui-sortable-handle" );
 		$.each( this.items, function() {
-			that._addClass(
-				this.instance.options.handle ?
-					this.item.find( this.instance.options.handle ) :
-					this.item,
-				"ui-sortable-handle"
-			);
+
+			// We use addClass() here, instead of _addClass(), because it is much faster.
+			( this.instance.options.handle ?
+				this.item.find( this.instance.options.handle ) :
+				this.item
+			).addClass( "ui-sortable-handle" );
 		} );
 	},
 
 	_destroy: function() {
+		this.element.find( ".ui-sortable-handle" ).removeClass( "ui-sortable-handle" );
 		this._mouseDestroy();
 
 		for ( var i = this.items.length - 1; i >= 0; i-- ) {


ui/widgets/sortable.js
```diff
@@ -118,19 +118,19 @@ return $.widget( "ui.sortable", $.ui.mouse, {
...
+			// We use addClass() here, instead of _addClass(), because it is much faster.
+			( this.instance.options.handle ?
+				this.item.find( this.instance.options.handle ) :
+				this.item
+			).addClass( "ui-sortable-handle" );
 		} );
 	},
 
 	_destroy: function() {
+		this.element.find( ".ui-sortable-handle" ).removeClass( "ui-sortable-handle" );
```
mgol, line 133: Looking at it, I'm not sure about this, it seems to be diverging from how classes are generally managed in jQuery UI.

@fnagel what do you think?

melloware (approved):

ui/widgets/sortable.js
melloware: I commented in the ticket that we use this same current workaround because users reported the same thing 4 years ago. The performance difference with 1000 items is 2 seconds instead of 2+ minutes in the browser

mgol: @melloware Thanks, that's an important data point. I'd still want to hear from @fnagel who was involved with UI when the team was larger, I may be lacking some context from those times.

fnagel: Sorry for the late response. Still in the process of moving my office and flat.

The thing is that `_addClass` does keep track of the elements and does some stuff when the element is removed (afair). It's part of the `this.options.classes` functionality and I think its related to https://github.com/jquery/jquery-ui/pull/2037

One objection against this change is that we might reintroduce old issues we once fixed by adding the classes option. Seems we paid for bug fixes with performance penalty.

naitsirch: Thanks for your reply. If you aren't comfortable with the patch, you can close it.

I have monkey patched that method in my project, so that I can still use v13 ;-)

mgol:
The risk here is that the proposed mechanism skips the logic used to handle classes which can bite in future changes to the library. I'm not that happy with hacks like that as they can bite in the future, edge cases can be discovered, etc.

It'd be better to understand what exactly causes the slowdown and see if there's a fix in the logic possible that would speed things up. See #2037 for an example of such a PR. It solved issue #2014 without skipping any core logic.

I don't have time to dive into this too much at the moment but if anyone finds a proper fix, I'd be happy to review such a PR, feel free to ping me in a PR then.

PR #2086, opened 2022-06-10
Update cancel check to be shadowRoot compatible

malarahfelipe:
This proposal is made to ensure if the event should be canceled if the first element in the [composedPath](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath) (or the event.target as fallback) matches with the options.cancel selector

This behaviour happens when we're using [shadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot) and we're trying to cancel the mousedown event when it starts from an element within the shadowRoot.

### DIFF
--- a/ui/widgets/mouse.js
+++ b/ui/widgets/mouse.js
@@ -73,6 +73,11 @@ return $.widget( "ui.mouse", {
 		}
 	},
 
+	/**
+	 *
+	 * @param {MouseEvent & { originalEvent: MouseEvent }} event
+	 * @returns {boolean}
+	 */
 	_mouseDown: function( event ) {
 
 		// don't let more than one widget handle mouseStart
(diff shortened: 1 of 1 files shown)


malarahfelipe:
@fnagel sure thing 👍

malarahfelipe:
Fixed @fnagel.

Surely this is a specific case scenario, where I'm using Sortable with Web Components, but this may come in hand when Web Components get more and more popular

As answered in the question below, the purpose of using event.composedPath() is to get specifically which element triggered the event within the Web-Component (container).
 
image

fnagel:
@malarahfelipe Tests still fail :-(

malarahfelipe:
Sorry, missed the grunt test workflow
I believe is fine now 👍

fnagel:
Still failing.

malarahfelipe:
Surely, the lint got me again sorry, now I was able to run them locally and seems fine
image

Thanks

mgol:
It'd be good to see what case currently doesn't work on a live test case. This PR needs tests anyway, preparing an isolated test case may be a good starting point in creating one.

As it stands, I don't have a specific opinion as I don't understand what exactly this PR is trying to fix.

fnagel:
@mgol Thanks for the feedback! I must confess I'm not really familiar with Web Components so I hoped for your input :-)

Maybe @dmethvin can help us out here?

@malarahfelipe Thanks for fixing the tests! Can you give some some more insights?

malarahfelipe:
Thanks for the feedback on this guys

Sure @fnagel 
I'll prepare two codepens one using web-components and other not to better explain the differences on this

malarahfelipe:
Okay, after long time 😅
I've set up three simple codepens to better explain the proposed solution

1. JqueryUI.Sortable without ShadowRoot https://codepen.io/malarahfelipe/pen/QWmEoVz
2. JqueryUI.Sortable with ShadowRoot ❌https://codepen.io/malarahfelipe/pen/XWEjpWo
3. JqueryUI.Sortable with ShadowRoot ✅ https://codepen.io/malarahfelipe/pen/eYMBYGb

Also I've record a <1min loom video just dragging things around so you can see the output in the html / console

https://www.loom.com/share/a2c4f95d5eae4589967a00041bcb7307

Thank you

melloware (approved):

PR #2093, opened 2022-06-24
Build: Add dependabot.yml config (GitHub Actions)

coliff:
As requested here https://github.com/jquery/jquery/pull/5057#issuecomment-1164341634

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: monthly


mgol (commented):
UI is not as maintained as jQuery; is it possible to make it happen, say, bi-yearly?

coliff:
> UI is not as maintained as jQuery; is it possible to make it happen, say, bi-yearly?

Bi-yearly isn't an option, but using monthly should be fine. You don't need to action (merge) them if you don't want/need to.

mgol (approved):
OK, fair enough. Thank you!

PR #2100, opened 2022-07-08
Remove special chars from localization

fnagel:
- [x] Rebase and merge #2050 branch
- [x] Remove all missing prefixes/suffixes
- [x] Remove all left symbols
- [x] Make sure there are no empty texts, create localization by using Google Translate (and verified by another)
- [ ] Update related issues
   - [ ] https://github.com/jquery/jquery-ui/issues/2048
   - [ ] https://github.com/jquery/jquery-ui/issues/1900
   - [ ] https://github.com/jquery/jquery-ui/pull/2050


mgol (approved):
LGTM. Just one question: do we want to merge #2050 separately before this one to preserve the original author of that change?

fnagel:
@mgol We can do this but another (maybe more easier option) would be to just go ahead and merge all three commits of this PR without squashing them to one. I would prefer the latter one.

Issue #2101, opened 2022-07-08
Checkboxradio's label text being evaluated as HTML on refresh


Elkano:
If you generate a Checkboxradio from a checkbox/radio with a label that contains encoded HTML, e.g. `&lt;em&gt;test&lt;/em&gt;` this will work fine at first.  
If however a refresh is triggered on that instance (explicitly or e.g. by turning it into a `Controlgroup`), the previously escaped HTML will now be evaluated.

If the label was created based on some user input, this could lead to unexpected code execution even though the initial output was escaped.

Example:
https://jsfiddle.net/69krwj75/

This is caused by the initial label being read as text (and thus decoded) for text nodes.
https://github.com/jquery/jquery-ui/blob/d66fdd5c9a1afac13138c7f48b068c36236b9358/ui/widgets/checkboxradio.js#L78

When the label is later updated, it is however done via `append`.
https://github.com/jquery/jquery-ui/blob/d66fdd5c9a1afac13138c7f48b068c36236b9358/ui/widgets/checkboxradio.js#L263

A fix would likely be to read the initial label text as HTML as well to preserve the encoding of the entities, however the likely was some reason to do things the way they currently are.

Judging from the blame, this bug has existed since the initial code from 8 years ago.

---

I'm aware that the teams resources are limited but I wanted to at least report the bug.  
As a workaround, one can wrap their text into a `<span>` and thus causing it to be read via `this.outerHTML` instead.

mgol:
Thanks for the report! PR: #2102. We've just released it as part of 1.13.2: https://blog.jqueryui.com/2022/07/jquery-ui-1-13-2-released/

I'll request a CVE shortly.

In the future, please direct security reports at security@jquery.com instead of GitHub issues so that they are not immediately visible to the whole world. I'll add the `SECURITY.md` file to this repo to make this policy clearer; we already have one for jQuery but we missed one for UI.

mgol:
@Elkano I submitted a security advisory at:
https://github.com/jquery/jquery-ui/security/advisories/GHSA-h6gj-6jjq-h8g9
and added you in the Credits section; if you want to show up there, you’ll need to approve.

mgol:
The issue now has its CVE identifier: `CVE-2022-31160`.

PR #2102, opened 2022-07-14
Checkboxradio: Don't re-evaluate text labels as HTML

mgol:
If you generate a Checkboxradio from a checkbox/radio with a label that
contains encoded HTML, e.g. `&lt;em&gt;test&lt;/em&gt;` this will work fine
at first. If, however a refresh is triggered on that instance (explicitly or
e.g. by turning it into a `Controlgroup`), the previously escaped HTML will
now be evaluated.

If the label was created based on some user input, this could lead to
unexpected code execution even though the initial output was escaped.

Fixes gh-2101

### DIFF
--- a/tests/unit/checkboxradio/checkboxradio.html
+++ b/tests/unit/checkboxradio/checkboxradio.html
@@ -64,6 +64,18 @@
 <label>
 	<input type="checkbox" id="label-with-no-for"/>
 </label>
+<label>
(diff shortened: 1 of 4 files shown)


mgol:
This was approved by @fnagel outside of the context of this PR, hence I'll merge it without waiting for an official review here.

PR #2103, opened 2022-07-18
Docs: Add a security policy in SECURITY.md

mgol:
This is needed so that people know how to report security issues.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,11 +2,10 @@
 
 ## Supported Versions
 
-The [latest released version](https://github.com/jquery/jquery-ui/releases) of jQuery is supported.
+The [latest released version](https://github.com/jquery/jquery-ui/releases) of jQuery UI is supported.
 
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Supported Versions
+
+The [latest released version](https://github.com/jquery/jquery-ui/releases) of jQuery is supported.
+
+## Reporting a Vulnerability
+
+Please email security@jquery.com, and we will respond as quickly as possible.
+
+If the vulnerability is considered valid and accepted, a patch will be made for the latest jQuery version.
```
fnagel, line 11: Same here

fnagel (changes requested):
Very good idea to add this file!

I think we should change from jQuery to jQuery UI and then we are good to go.

mgol:
@fnagel I was too quick. 🤦🏻 All fixed now, PTAL.

Issue #2107, opened 2022-08-03
Script error for "jquery-patch.js"


mihaelams1:
I know core.js is deprecated, but until then if one is using individual files as opposed to the bundled version they might encounter an issue in which requirejs can't find jquery-patch:
![image](https://user-images.githubusercontent.com/1153973/182703838-893ab86a-3c05-4b83-8a06-d74b80acb869.png)
This is due to the fact that in [core.js](https://github.com/jquery/jquery-ui/blob/main/ui/core.js) it is referenced as "jquery-patch.js" as opposed to just "jquery-patch":

![image](https://user-images.githubusercontent.com/1153973/182704452-279b5eff-ac4a-4ad9-9b6d-593a5bf08fec.png)

mgol:
Thanks for the report! Fix at #2109.

PR #2109, opened 2022-08-10
Core: Fix dependencies in ui/core.js

mgol:
The `jquery-patch` reference was included with the `.js` extension which is
not allowed in AMD.

Fixes gh-2107


fnagel (approved):
+1 by reading

PR #2115, opened 2022-09-01
Fixes removing the namespace from an event name

TimWerdin:
Bootstrap uses events which need to include a namespace. E.g. the event name to bind to showing a modal is `show.bs.modal`. Just setting the event on `show` will not work.

Example:
```js
$(document).on('show', '.modal', function() { console.log('this will never get fired')});
$(document).on('show.bs.modal', '.modal', function() { console.log('this will be fired')})
```

The issue when using jquery-UI's `_on` method is that the namespace gets cut off. So this will not work:
```js
$.widget('mywidget', {
    _create: function() {
        this._on({
            'show.bs.modal .modal': function() {
                alert('this will not be fired');
            }
        });
    }
});
```

The issue is the regex in the `_on` method. It does not allow a dot in the event name and also it searches for zero to unlimited amount of spaces to split of the selector.
This results into the following:
```js
var match = event.match( /^([\w:-]*)\s*(.*)$/ );
var eventName = match[ 1 ] + instance.eventNamespace; // "show.mywidget"
var selector = match[ 2 ]; // ".bs.modal .modal"
```

To prevent this I changed the regex to also allow dots so namespaces will not be broken apart and also to require at least one whitespace.
```js
var match = event.match( /^([\w:.-]*)\s+(.*)$/ );
var eventName = match[ 1 ] + instance.eventNamespace; // "show.bs.modal.mywidget"
var selector = match[ 2 ]; // ".modal"
```

### DIFF
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -621,7 +621,7 @@ $.Widget.prototype = {
 					handler.guid || handlerProxy.guid || $.guid++;
 			}
 
-			var match = event.match( /^([\w:-]*)\s*(.*)$/ );
+			var match = event.match( /^([\w:.-]*)\s+(.*)$/ );
 			var eventName = match[ 1 ] + instance.eventNamespace;
 			var selector = match[ 2 ];
 


mgol:
Thanks for the PR. All changes require tests, though. I’m not sure why existing tests didn’t run for this PR but maybe that would happen on the next update to the PR.

fnagel:
Closing and reopening in order to run the tests.

melloware (approved):

mgol:
@TimWerdin in addition to writing new tests, you need to ensure existing ones pass. You can now see that's not the case right now.

TimWerdin:
I was on vacation until yesterday. I will have a look this week

PR #2139, opened 2022-12-03
Demos: Better wording for the purposefully broken tab

DeerBear:
The wording was extremely jarring and could put potential users off. Made it better by fixing both grammar and punctuation.


mgol (approved):
Thanks!

Issue #1971, opened 2021-08-31
Datepicker: beforeShowDay is passed wrong date when dayNamesShort are used 


pramodjodhani:
Please see the demo of the issue here: https://jsfiddle.net/ea8utdyx/

Here are the steps to reproduce the problem:

1. Use a nonstandard date format, like 'D M d yy'
2. Pass dayNamesShort, example:  dayNamesShort: ["Su", "Mo", "Tu", "We", "Th", "Fr", "Sa"],
3. Pass a callback function to beforeShowDay argument

Notice that the date which is to the beforeShowDay callback is incorrect, it's about 5 years in the future, as a result we cannot determine if a date should be visible or not. 

I notice the problem doesn't happen when we comment the `dayNamesShort` argument.

kendebacker:
The error is occurring because the strings you are passing to defaultDate, minDate, and maxDate are not in the specified dateFormat. Because you set dayNamesShort to  "Su", "Mo", "Tu", "We", "Th", "Fr", or "Sa", D must be one of those, but in the strings you pass to defaultDate, minDate, and maxDate, D is "Fri" and "Tue", and this causing the problems.  It is a little confusing that an error is not thrown, I am trying to work on that.

pramodjodhani:
Looks like jQuery UI team have fixed the issue we had raised. YAY! 🕺

mgol:
Thanks for the report. Since the issue is already in 1.12, given limited team resources it's not likely to be fixed by the UI team; see the project status at https://blog.jqueryui.com/2021/10/jquery-maintainers-update-and-transition-jquery-ui-as-part-of-overall-modernization-efforts/. PRs are welcome if they're not too complex and contain tests.

PR #2143, opened 2022-12-25
Tests: Test only valid values for Datepicker defaultDate, min/maxDate

kendebacker:
I was working on an issue with Datepicker (#1971) and when running some tests found that the tests are run with periods other than "y", "m", "w", and "d", which are the only valid periods listed in the docs https://api.jqueryui.com/datepicker/.

I think this is a particular problem for the tests with "m" and "d", as "M" and "D" refer to something very different in the Datepicker dateFormat. The latter refer to day/month names while the former to dates/month numbers.

Happy Holidays!

### DIFF
--- a/tests/unit/datepicker/options.js
+++ b/tests/unit/datepicker/options.js
@@ -303,49 +303,49 @@ QUnit.test( "defaultDate", function( assert ) {
 		simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
 	date.setDate( date.getDate() - 1 );
 	testHelper.equalsDate( assert, inp.datepicker( "getDate" ), date, "Default date -1d" );
-	inp.datepicker( "option", { defaultDate: "+3D" } ).
+	inp.datepicker( "option", { defaultDate: "+3d" } ).
 		datepicker( "hide" ).val( "" ).datepicker( "show" ).
 		simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
 	date.setDate( date.getDate() + 4 );
-	testHelper.equalsDate( assert, inp.datepicker( "getDate" ), date, "Default date +3D" );
+	testHelper.equalsDate( assert, inp.datepicker( "getDate" ), date, "Default date +3d" );
 	inp.datepicker( "option", { defaultDate: " -2 w " } ).
 		datepicker( "hide" ).val( "" ).datepicker( "show" ).
 		simulate( "keydown", { keyCode: $.ui.keyCode.ENTER } );
 	date = new Date();
 	date.setDate( date.getDate() - 14 );
 	testHelper.equalsDate( assert, inp.datepicker( "getDate" ), date, "Default date -2 w" );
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the PR.

> I think this is a particular problem for the tests with "m" and "d", as "M" and "D" refer to something very different in the Datepicker dateFormat. The latter refer to day/month names while the former to dates/month numbers.

In that case, how come all tests pass on `main`?

kendebacker:
Hi,

All tests pass at the moment because currently both capital and lowercase time periods for relative dates are being accepted

Screen Shot 2022-12-27 at 1 09 38 PM

I am opting to change this to just lowercase time periods because that is the allowed input in the docs for relative dates

Screen Shot 2022-12-27 at 12 28 07 PM

which I believe was set that way because for date formats there is significant difference in meaning between upper and lower case 

Screen Shot 2022-12-27 at 1 02 20 PM

I was planning to modify the code in 1588-1602 in another PR but thought I should wait to see if my thinking was correct.

mgol:
Thanks for the info. At this stage of the project we're trying to avoid unnecessary breaking changes so I'm not sure we'd want to change the source here but I'll defer to @fnagel here who has way more experience with datepicker code.

kendebacker:
Ah ok, thank you for explaining. 

Apologies for straying outside the scope of this PR, but would a fix to #1971 be something worth pursuing? The issue is caused when the defaultDate, minDate, or maxDate setting option is given as a date string that is not in the specified dateFormat, the year is set 2028 and no error is thrown. 

I was thinking that some code could be added to ensure that if the defaultDate, minDate, or maxDate is given as a date string that it is in the specified format, or else alert the user, as the user is not being alerted to the inconsistency in formatting.

Thanks again for taking the time to respond. I've used jQuery a lot in the past and am looking to get experience contributing to open-source, but if if the project stage is not ideal for this, no worries

mgol:
@kendebacker The issue sounds worth fixing, I think, but I'd still prefer to hear Felix's opinion first. 🙂 This is holiday time so let's wait a bit for his answer, perhaps in 2023.

kendebacker:
Ok sounds good, thanks @mgol , happy holidays!

fnagel:
Hi @kendebacker, hi @mgol! Happy new year!

I'm not that familiar with the Datepicker code as I only worked on the never published rewrite of it (Calendar and Datepicker 2.0).

Anyway, I don't see why we should test with other values than documented. So the current code of the PR seems valid to me. We might want to think about testing both (upper and lowercase) as both are supported. @mgol What do you think?


> At this stage of the project we're trying to avoid unnecessary breaking changes so I'm not sure we'd want to change the source here

Yes. As long as changing the actual code doesn't fix a bug, we should not touch it IMO.

Regarding a fix for #1971 - sure, why not :-)

mgol (approved):

mgol:
@kendebacker sorry for the delay; landed now!

kendebacker:
Thanks @mgol !

Issue #2082, opened 2022-06-02
Hidden select and visible ui-select have different values


janhaertel:
When I hide **options** in my select by setting the **hidden** attribute and also preselect an option, then I get an error when opening the selectmenu.

Here is the code the jquery executes until the error occures:

`
...
this.button.one( "focusin"); -> that._refreshMenu(); -> this._getSelectedItem().data( "ui-selectmenu-item" ); -> this.menuItems.eq( this.element[ 0 ].selectedIndex ).parent( "li" );
 `

This happens because the options' DOM-element attribute **selectedIndex** does not match the index from the rendered ui-select-menu.
When I start selecting other entries now, the native select has a different value than what is displayed in the ui-selectmenu-button.

I also created a fiddle, in which the selectmenu won't open at all.

https://jsfiddle.net/isaraystanz/750p69qu/19/

On my local system the selectmenu will show and I can select entries, but the select's value gets messed up so I can't use it safely.

mgol:
It looks like I messed the 1.12.1 version of the test case above. So, let me share new ones with the latest version of jQuery jQuery UI for each line:
* (working) with jQuery 3.6.3 & jQuery UI 1.12.1: https://jsfiddle.net/m_gol/h4vrs6wn/
* (broken) with jQuery 3.6.3 & jQuery UI 1.13.2: https://jsfiddle.net/m_gol/cfkxa1pL/

mgol:
The first bad commit is a2b25ef6caae3e1a272214839b815a6387618124.

mgol:
The first bad commit is a2b25ef6caae3e1a272214839b815a6387618124. There's even a comment from 4 years ago suggesting a similar issue: https://github.com/jquery/jquery-ui/commit/a2b25ef6caae3e1a272214839b815a6387618124#commitcomment-26579991

mgol:
PR with a fix: #2144

PR #2144, opened 2023-01-03
Selectmenu: Fix selecting options following hidden ones

mgol:
Change a2b25ef6caae3e1a272214839b815a6387618124 made options with the `hidden` attribute skipped when rendering. However, that makes indexes misaligned with native options as hidden ones maintain their index values. Instead, don't skip hidden options but add the `hidden` attribute to the respective jQuery UI elements as well.

Fixes gh-2082
Ref a2b25ef6caae3e1a272214839b815a6387618124

### DIFF
--- a/tests/unit/selectmenu/core.js
+++ b/tests/unit/selectmenu/core.js
@@ -394,11 +394,70 @@ QUnit.test( "Options with hidden attribute should not be rendered", function( as
 	setTimeout( function() {
 		button.trigger( "click" );
 		options = menu.children()
-			.map( function() {
-				return $( this ).text();
+			.get()
+			.filter( function( item ) {
+				return $( item ).is( ":visible" );
 			} )
-			.get();
+			.map( function( item ) {
+				return $( item ).text();
+			} );
 		assert.deepEqual( options, [ "Slower", "Medium", "Fast", "Faster" ], "correct elements" );
+
+		ready();
+	} );
+} );
+
+QUnit.test( "Options with hidden attribute should not break the widget (gh-2082)",
+		function( assert ) {
+	var ready = assert.async();
+	assert.expect( 1 );
+
+	var button;
+	var element = $( "#speed" );
+
+	element.find( "option" ).slice( 0, 2 ).prop( "hidden", true );
+	element.val( "Faster" );
+	element.selectmenu();
+
+	button = element.selectmenu( "widget" );
+	button.simulate( "focus" );
+	setTimeout( function() {
+		try {
+			button.trigger( "click" );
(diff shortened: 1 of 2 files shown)


tests/unit/selectmenu/core.js
```diff
@@ -394,16 +394,75 @@ QUnit.test( "Options with hidden attribute should not be rendered", function( as
 	setTimeout( function() {
 		button.trigger( "click" );
 		options = menu.children()
-			.map( function() {
-				return $( this ).text();
+			.get()
+			.filter( function( item ) {
+				return $( item ).is( ":visible" );
```
mgol, line 399: I had to change the test as the options are now back in the DOM, just hidden.

ui/widgets/selectmenu.js
```diff
@@ -354,7 +354,12 @@ return $.widget( "ui.selectmenu", [ $.ui.formResetMixin, {
 		if ( item.disabled ) {
 			this._addClass( li, null, "ui-state-disabled" );
 		}
-		this._setText( wrapper, item.label );
+
+		if ( item.hidden ) {
+			li.prop( "hidden", true );
```
fnagel, line 359: I'm not sure about adding hidden prop here. We don't do that anywhere else. But it seems there is no other usage of the hidden attribute, so there is nothing to compare it to directly. 

What about using "aria-hidden" attribute and "ui-helper-hidden" CSS class?

fnagel (commented):
Please see my comment, otherwise looks good.

ui/widgets/selectmenu.js
mgol: @fnagel Sure, I can change the code. The behavior in IE 9 & 10 will be different but perhaps if the `hidden` attribute is there, it's expected that we should hide it even if the browser doesn't support it natively.

One thing - I see the `ui-helper-hidden` class definition in CSS but other than that, I don't see it anywhere in code; only the `ui-helper-hidden-accessible` one seems to be used. Am I missing anything?

fnagel: I'm not sure about my proposed change either :-D Just wanted to point out our options and get your opinion on it.

You are right, `ui-helper-hidden` is not used in JS. And `ui-helper-hidden-accessible` seems not correct here. 

Maybe its best to use it as is it is (with the hidden property) and let the client (or its assistive technology) decide what to do.

mgol: In that case, I think I'd leave it at `hidden`. @fnagel do you have other concerns? If not, could you approve?

fnagel (commented):

PR #2157, opened 2023-03-30
Build:Tests: Fix various issues with running tests on Apple Silicon Macs

mgol:
I wasn't able to run UI tests on my new Apple Silicon Mac so I updated all dependencies and fixed all the issues I found. Tests now work fine for me and we have the latest QUnit which is also nice.


fnagel (commented):
Uhhh, that is a big one! Nice work here!! Sooner or later we'll gotten in trouble with all those old dependencies.

For a proper review, this one is a little too big. I took a brief look at everything and I think its fine. Merged anyway :-D

And again: sorry for the delay!

PR #2159, opened 2023-03-30
Tests: Fix selectmenu width tests in Firefox with jQuery 3.0 & 3.1

mgol:
jQuery 3.0 & 3.1 used `getBoundingClientRect()` in its `width`/`height` calculations and that causes minor differences in fractional width computations. Allow a tiny delta in tests to fix those tests breaking in Firefox.

Ref jquery/jquery#3561


fnagel (commented):

PR #2160, opened 2023-03-30
Dialog: Fix IE tests about interactions with other dialogs

mgol:
The workaround triggering a second delayed focus to a tabbable element was applied in jQueries 3.4 & 3.5 to fix IE issues. It also needs to be applied in jQuery 3.6 which this commit adds.

This will no longer be needed in jQuery 3.7 and jQuery 3.x-git tests are already passing because of that.


fnagel (commented):

PR #2180, opened 2023-07-29
Slider: Improve behaviour on overlapping ranges

zalito12:
Improve the handle detection when they overlap in slider ranges.

With this change when you click on the slider it will check if you click over the current overlapping value it will move the max handle, if you click under the value it will move the min handler. Also if you are on the ends it will move the only one you can move.

### DIFF
--- a/ui/widgets/slider.js
+++ b/ui/widgets/slider.js
@@ -198,17 +198,31 @@ return $.widget( "ui.slider", $.ui.mouse, {
 
 		position = { x: event.pageX, y: event.pageY };
 		normValue = this._normValueFromMouse( position );
-		distance = this._valueMax() - this._valueMin() + 1;
-		this.handles.each( function( i ) {
-			var thisDistance = Math.abs( normValue - that.values( i ) );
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the PR. Can you report an issue with a test case showing what exactly it is that you'd like to fix? Also, any change like that will need unit tests.

zalito12:
> Thanks for the PR. Can you report an issue with a test case showing what exactly it is that you'd like to fix? Also, any change like that will need unit tests.

Sure! I will, but I'm not sure about how to test this one, I should test when handles are overlapping and you click on rigth/left side, instead of trying to move the last one touched it should move the left/rigth handle correspondingly. Will ``click`` simulate over ``element`` slider work?

mgol:
jQuery UI uses jQuery Simulate to simulate events and you can find many examples in unit tests. This folder is for slider: https://github.com/jquery/jquery-ui/tree/main/tests/unit/slider, have a look especially at `event.js` for examples of using events in tests.

PR #2196, opened 2023-11-06
Build: Update grunt-contrib-qunit

mgol:
This also switches us to the new Chrome Headless mode.

Ref https://developer.chrome.com/articles/new-headless/
Ref https://pptr.dev/#default-runtime-settings


mgol:
Something's broken here, moving this to the draft status.

mgol:
It will be deleted instead

PR #2205, opened 2023-12-29
Draggable: Set containment size after helper size change

velezh:
When helper size was changed in 'start' event, containment still store old values.
Call _setContainment() right after triggering 'start' event.

### DIFF
--- a/ui/widgets/draggable.js
+++ b/ui/widgets/draggable.js
@@ -208,14 +208,14 @@ $.widget( "ui.draggable", $.ui.mouse, {
 			this._adjustOffsetFromHelper( o.cursorAt );
 		}
 
-		//Set a containment if given in the options
-		this._setContainment();
-
 		//Trigger event + callbacks
 		if ( this._trigger( "start", event ) === false ) {
 			this._clear();
 			return false;
 		}
+
+		//Set a containment if given in the options
+		this._setContainment();
 
 		//Recache the helper size
 		this._cacheHelperProportions();


mgol:
Thanks for the PR. We need all contributors to sign the CLA via the provided link.

Also, could you explain the issue you want to tackle in detail, ideally with a test case on a platform like JS Bin?

All such changes need to be accompanied by unit tests as well.

velezh:
https://jsbin.com/fahugidoho/edit?html,output
Try to drag item with long text to the right side of page. Problem in calculating dragging helper size, when it created it has longer width which get cached, after that I change helper size in `start` callback to match original item size, and new size is not gets recached.

mgol:
@fnagel What do you think about this?

@velezh Since the issue is already in 1.12, given limited team resources it's not likely to be fixed by the UI team; see the project status at https://blog.jqueryui.com/2021/10/jquery-maintainers-update-and-transition-jquery-ui-as-part-of-overall-modernization-efforts/. PRs are welcome if they're not too complex.

fnagel:
I have no real knowledge on how draggable works and if this change might have some negative impact. A test which shows what issue exactly is targeted here would be helpful. I've approved the test workflow, so we can see how existing tests work with this change.

@velezh Are you willing to sign the CLA? 


> https://jsbin.com/fahugidoho/edit?html,output
Try to drag item with long text to the right side of page. Problem in calculating dragging helper size, when it created it has longer width which get cached, after that I change helper size in start callback to match original item size, and new size is not gets recached.

What exactly are you trying to achieve here? The size of the box should change while dragging? Not quite sure where the issue is exactly... Maybe a test would help to understand.

PR #2220, opened 2024-03-25
Tests: Fix tests with the Git Core build

mgol:
jQuery PR jquery/jquery#5452 removed special handling of boolean attributes. Thankfully, this only broke a single jQuery UI test.


mgol:
You can see the current state here: https://swarm.jquery.org/job/14697

timmywil (approved):

PR #2227, opened 2024-04-04
Tests: Workaround a jQuery 1.8 issue with one draggable test

mgol:
The test "#8958: dialog can be opened while opening" had special behavior in jQuery 1.8 and the patch from commit 7adb13ac7c30a7ba33f1db256952be77b9e7a310 broke those assumptions. Adjust the test so that it passes again.


mgol:
It's about dialog, not draggable; I'll open a new PR>

PR #2230, opened 2024-04-04
Tests: Increase robustness of a draggable `revert` test

mgol:
One of the changes is increasing `revertDuration` from 200 to 300. For some reason, when jQuery 3.1 is used in an iframe (which TestSwarm uses) the timings are lower than expected and `setTimeout` too slow and the check for animation happens too late.

### DIFF
--- a/tests/unit/draggable/options.js
+++ b/tests/unit/draggable/options.js
@@ -1018,7 +1018,7 @@ QUnit.test( "opacity, default, switching after initialization", function( assert
 } );
 
 QUnit.test( "revert and revertDuration", function( assert ) {
-	var ready = assert.async();
+	var ready = assert.async( 2 );
 	assert.expect( 7 );
 
 	var element = $( "#draggable2" ).draggable( {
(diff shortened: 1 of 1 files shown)


tests/unit/draggable/options.js
```diff
@@ -1045,6 +1045,7 @@ QUnit.test( "revert and revertDuration", function( assert ) {
 	testHelper.move( element, 50, 50 );
 	setTimeout( function() {
 		assert.ok( $( "#draggable2" ).is( ":animated" ), "revert: true with revertDuration should animate" );
+		ready();
```
fnagel, line 1048: Is this added on purpose?

fnagel (approved):

timmywil (approved):

tests/unit/draggable/options.js
mgol: Yes, I wanted the test to wait for this as well - that's why the call `assert.async()` now takes `2` as an argument - it means `ready()` needs to be called twice before the test is marked as finished.

PR #2249, opened 2024-05-11
Core: Drop support for old jQuery or old browsers

mgol:
Changes:
* Drop support for jQuerys different than the latest for each major (also, remove the ability to run the test suite in jQuery <1.12.4 & <2.2.4)
* Drop support for IE & some other browsers
* Sortable: Rename `storedStylesheet` to `_storedStylesheet`
* Drop jquery-patch.js. Avoid patching jQuery. Instead:
	* use `CSS.escape` instead of `jQuery.escapeSelector`
	* use `.filter()` with a proper handler instead of `.even()`
* Remove the deprecated `ui/core.js` file
* Remove unused build/themes


ui/jquery-patch.js
```diff
@@ -1,89 +0,0 @@
...
-		},
-		odd: function() {
-			return this.filter( function( i ) {
-				return i % 2 === 1;
-			} );
-		}
-	} );
-}
-
-} );
```
timmywil, line 89: Are we sure we want to remove this? This had the side effect of patching jQuery so that even and odd could be used even in older versions of jQuery, so while UI's internal usage compensates, it could be unexpected behavior for certain users. Though, admittedly rare.

mgol: Hmm, good point. I was mostly thinking that patching jQuery Core is not a great thing to do if we can avoid it but you're right this affects backwards compat. I'll revert this change to be on the safe side.

mgol: I guess I can keep source changes that e.g. use `CSS.escape` directly. And I can alias `jQuery.escapeSelector` here to `CSS.escape` instead of this longer definition. Since it was supposed to be a polyfill, it should be fine, I think.

demos/slider/multiple-vertical.html
```diff
@@ -5,11 +5,6 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<title>jQuery UI Slider - Multiple sliders</title>
 	<link rel="stylesheet" href="../../themes/base/all.css">
-	<script src="../../external/jquery/jquery.js"></script>
-	<script src="../../ui/core.js"></script>
-	<script src="../../ui/widget.js"></script>
-	<script src="../../ui/mouse.js"></script>
-	<script src="../../ui/slider.js"></script>
```
timmywil, line 12: Why doesn't this need to load JS anymore?

timmywil (approved):
LGTM, pending my question

demos/slider/multiple-vertical.html
mgol: It does still load JS. This was the old way of doing it that also included loading the deprecated core.js file that this PR removes.

This template actually used both ways of loading JS; it looks like a migration oversight. I checked that the demo still works fine after these removals.

mgol:
Landed in daa6fb55b35065c49c0ffc879c94627bbf85404c...5f615c275d2c88305dd0733a12de3915a7c501af.

demos/slider/multiple-vertical.html
fnagel: Good catch!

Issue #2246, opened 2024-04-28
With the modal option set to true, elements in the background are still available in the AOM


rpkoller:
If you take a look for example at https://jqueryui.com/dialog/#modal-message  and you test the page with voiceover on macos and you are using the rotor functionality you will notice that elements in the background of the dialog modal are still available in the accessibility object model (aom). the number of available elements is sort of limited due to the fact that the example is wrapped in an iframe: 

https://github.com/jquery/jquery-ui/assets/1665422/7b711bef-fe50-4868-abaa-eaf36950f430

but for example in the context of a drupal sites which are also using jquery ui for building dialog modals they are not wrapped into an iframe, and you have all the elements in the background available in the aom as well. and there is no way for a screenreader user to distinguish what is an element within a dialog modal and what is an element outside of the dialog modal in its background. 

https://github.com/jquery/jquery-ui/assets/1665422/6de82506-3346-401b-98f3-29dd459d5e20

An easy fix would be if the modal option is set to true adding the `aria-modal="true"` attribute to the div of the dialog modal. i've manually added the aria-modal attribute via devtools and rerecorded the same example from before to illustrate the difference

https://github.com/jquery/jquery-ui/assets/1665422/04029429-a0a8-4f5d-9e1f-f9099c7d195e

mgol:
Thanks for the report. Does the issue you describe exist when jQuery UI 1.12.1 is used or only with jQuery UI 1.13.0 or newer?

rpkoller:
I've just rechecked, in Drupal 10.2.5. and 11.x-dev jQuery UI Dialog 1.13.2 is used. Took me a while to go back in versions to finally fin,  but turns out drupal 9.0.0 was using jQuery UI Dialog 1.12.1. And for all three versions of Drupal 11.x-dev, 10.2.5, and 9.0.0 the problem described in the issue summary exists.

mgol:
Thanks for the info. Since the issue is already in 1.12, given limited team resources it's not likely to be fixed by the UI team; see the project status at https://blog.jqueryui.com/2021/10/jquery-maintainers-update-and-transition-jquery-ui-as-part-of-overall-modernization-efforts/. PRs are welcome if they're not too complex and contain tests.

[...]

PR #2257, opened 2024-05-24
Dialog: Add aria-modal support

rpkoller:
PR for the issue outlined in https://github.com/jquery/jquery-ui/issues/2246 . Hope I've done everything in the correct manner. Was trying to follow http://contribute.jquery.org/commits-and-pull-requests/ as good as I was able to. 

In regards of the PR itself, broader considerations are outlined in the corresponding issue. But I've went with adding `aria-modal="true` only, since it is the least invasive and breaking option in regards of backward compatibility  to fix the problem. In short, elements in the background of a dialog modal are added to the AOM without the `aria-modal` attribute or the `dialog` element in place. And for the test i went with the same pattern that was applied if the `role` attribute is in place with a value of `dialog`. Therefore I've checked if the attribute `aria-modal` is there with a value set to `true`

### DIFF
--- a/tests/unit/dialog/core.js
+++ b/tests/unit/dialog/core.js
@@ -85,7 +85,7 @@ QUnit.test( "ARIA", function( assert ) {
 } );
 
 QUnit.test( "aria-modal", function( assert ) {
-	assert.expect( 12 );
+	assert.expect( 9 );
 
 	var element = $( "<div>" ).dialog( { modal: true } ),
 		wrapper = element.dialog( "widget" );
--- a/tests/unit/dialog/core.js
+++ b/tests/unit/dialog/core.js
@@ -70,12 +70,11 @@ QUnit.test( "title id", function( assert ) {
 } );
 
 QUnit.test( "ARIA", function( assert ) {
-	assert.expect( 5 );
+	assert.expect( 4 );
 
 	var element = $( "<div>" ).dialog(),
 		wrapper = element.dialog( "widget" );
 	assert.equal( wrapper.attr( "role" ), "dialog", "dialog role" );
-	assert.equal( wrapper.attr( "aria-modal" ), "true", "aria-modal attribute" );
 	assert.equal( wrapper.attr( "aria-labelledby" ), wrapper.find( ".ui-dialog-title" ).attr( "id" ) );
 	assert.equal( wrapper.attr( "aria-describedby" ), element.attr( "id" ), "aria-describedby added" );
 	element.remove();
--- a/tests/unit/dialog/core.js
+++ b/tests/unit/dialog/core.js
@@ -84,6 +84,26 @@ QUnit.test( "ARIA", function( assert ) {
 	element.remove();
 } );
 
+QUnit.test( "aria-modal", function( assert ) {
+	assert.expect( 3 );
+
+	var element = $( "<div>" ).dialog(),
+		wrapper = element.dialog( "option", "modal", true  );
+		assert.equal( wrapper.attr( "aria-modal" ), "true", "aria-modal attribute set to true" );
+	element.remove();
+
+	var element = $( "<div>" ).dialog(),
+		wrapper = element.dialog( "option", "modal", false );
+		assert.equal( wrapper.attr( "aria-modal" ), "false", "aria-modal attribute set to false" );
+	element.remove();
+
+	var element = $( "<div>" ).dialog();
+		assert.equal( element.dialog( "widget" ).attr( "aria-describedby" ), null, "no aria-modal attribute added" );
+	element.remove();
+
+} );
+
+
 QUnit.test( "widget method", function( assert ) {
 	assert.expect( 1 );
 	var dialog = $( "<div>" ).appendTo( "#qunit-fixture" ).dialog();
--- a/tests/unit/dialog/core.js
+++ b/tests/unit/dialog/core.js
@@ -91,31 +91,31 @@ QUnit.test( "aria-modal", function( assert ) {
 		wrapper = element.dialog( "widget" );
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
 	element.dialog( "option", "modal", false );
-	assert.ok( !wrapper.attr( "aria-modal" ), "modal option set to false, aria-modal attribute not added" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
 	element.dialog( "option", "modal", null );
-	assert.ok( !wrapper.attr( "aria-modal" ), "modal option not set, aria-modal attribute not added" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
 	element.dialog( "option", "modal", true );
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
 	element.remove();
 
 	var element = $( "<div>" ).dialog( { modal: false } ),
 		wrapper = element.dialog( "widget" );
-	assert.ok( !wrapper.attr( "aria-modal" ), "modal option set to false, aria-modal attribute not added" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
 	element.dialog( "option", "modal", true );
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
 	element.dialog( "option", "modal", null );
-	assert.ok( !wrapper.attr( "aria-modal" ), "modal option not set, aria-modal attribute not added" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
 	element.dialog( "option", "modal", false );
-	assert.ok( !wrapper.attr( "aria-modal" ), "modal option set to false, aria-modal attribute not added" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
 	element.remove();
 
 	var element = $( "<div>" ).dialog(),
 		wrapper = element.dialog( "widget" );
-	assert.ok( !wrapper.attr( "aria-modal" ), "modal option not set, aria-modal attribute not added" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
 	element.dialog( "option", "modal", true );
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
 	element.dialog( "option", "modal", false );
-	assert.ok( !wrapper.attr( "aria-modal" ), "modal option set to false, aria-modal attribute not added" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
 	element.dialog( "option", "modal", null );
 	assert.equal( wrapper.attr( "aria-modal" ), null, "modal option not set, aria-modal attribute not added" );
 	element.remove();
--- a/tests/unit/dialog/core.js
+++ b/tests/unit/dialog/core.js
@@ -92,8 +92,6 @@ QUnit.test( "aria-modal", function( assert ) {
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
 	element.dialog( "option", "modal", false );
 	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
-	element.dialog( "option", "modal", null );
-	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
 	element.dialog( "option", "modal", true );
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
 	element.remove();
@@ -103,8 +101,6 @@ QUnit.test( "aria-modal", function( assert ) {
 	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
 	element.dialog( "option", "modal", true );
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
-	element.dialog( "option", "modal", null );
-	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
 	element.dialog( "option", "modal", false );
 	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
 	element.remove();
@@ -116,8 +112,6 @@ QUnit.test( "aria-modal", function( assert ) {
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
 	element.dialog( "option", "modal", false );
 	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
-	element.dialog( "option", "modal", null );
-	assert.equal( wrapper.attr( "aria-modal" ), null, "modal option not set, aria-modal attribute not added" );
 	element.remove();
 } );
 
--- a/tests/unit/dialog/core.js
+++ b/tests/unit/dialog/core.js
@@ -87,20 +87,20 @@ QUnit.test( "ARIA", function( assert ) {
 QUnit.test( "aria-modal", function( assert ) {
 	assert.expect( 3 );
 
-	var element = $( "<div>" ).dialog(),
-		wrapper = element.dialog( "option", "modal", true  );
-		assert.equal( wrapper.attr( "aria-modal" ), "true", "aria-modal attribute set to true" );
+	var element = $( "<div>" ).dialog( "option", "modal", true ),
+		wrapper = element.dialog( "widget" );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "aria-modal attribute set to true" );
 	element.remove();
 
-	var element = $( "<div>" ).dialog(),
-		wrapper = element.dialog( "option", "modal", false );
-		assert.equal( wrapper.attr( "aria-modal" ), "false", "aria-modal attribute set to false" );
+	var element = $( "<div>" ).dialog( "option", "modal", false ),
+		wrapper = element.dialog( "widget" );
+	assert.equal( wrapper.attr( "aria-modal" ), "false", "aria-modal attribute set to false" );
 	element.remove();
 
-	var element = $( "<div>" ).dialog();
-		assert.equal( element.dialog( "widget" ).attr( "aria-describedby" ), null, "no aria-modal attribute added" );
+	var element = $( "<div>" ).dialog( "option", "modal", null ),
+		wrapper = element.dialog( "widget" );
+	assert.equal( wrapper.attr( "aria-modal" ), null, "no aria-modal attribute added" );
 	element.remove();
-
 } );
 
 
--- a/tests/unit/dialog/core.js
+++ b/tests/unit/dialog/core.js
@@ -87,17 +87,17 @@ QUnit.test( "ARIA", function( assert ) {
 QUnit.test( "aria-modal", function( assert ) {
 	assert.expect( 3 );
 
-	var element = $( "<div>" ).dialog( "option", "modal", true ),
+	var element = $( "<div>" ).dialog( "options", "modal", true ),
 		wrapper = element.dialog( "widget" );
 	assert.equal( wrapper.attr( "aria-modal" ), "true", "aria-modal attribute set to true" );
 	element.remove();
 
-	var element = $( "<div>" ).dialog( "option", "modal", false ),
+	var element = $( "<div>" ).dialog( "options", "modal", false ),
 		wrapper = element.dialog( "widget" );
 	assert.equal( wrapper.attr( "aria-modal" ), "false", "aria-modal attribute set to false" );
 	element.remove();
 
-	var element = $( "<div>" ).dialog( "option", "modal", null ),
+	var element = $( "<div>" ).dialog( "options", "modal", null ),
 		wrapper = element.dialog( "widget" );
 	assert.equal( wrapper.attr( "aria-modal" ), null, "no aria-modal attribute added" );
 	element.remove();
--- a/tests/unit/dialog/core.js
+++ b/tests/unit/dialog/core.js
@@ -70,7 +70,7 @@ QUnit.test( "title id", function( assert ) {
 } );
 
 QUnit.test( "ARIA", function( assert ) {
-	assert.expect( 4 );
+	assert.expect( 5 );
 
 	var element = $( "<div>" ).dialog(),
 		wrapper = element.dialog( "widget" );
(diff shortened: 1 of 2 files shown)


tests/unit/dialog/core.js
```diff
@@ -75,6 +75,7 @@ QUnit.test( "ARIA", function( assert ) {
 	var element = $( "<div>" ).dialog(),
 		wrapper = element.dialog( "widget" );
 	assert.equal( wrapper.attr( "role" ), "dialog", "dialog role" );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "aria-modal attribute" );
```
fnagel, line 78: Please update line 73 from `assert.expect( 4 );` to `assert.expect( 5 );` as you added a test :-)

fnagel (changes requested):
Apart from the little change I've noted this looks looks good to me, thanks for your contribution!

tests/unit/dialog/core.js
rpkoller: ohhhhh that's what the number four in `assert.expect( 4 )` refers to, the number of tests. that makes sense, understood now. thanks! and just for the context, mentioned it on the issue, i am not a developer nor have i ever written a test. just tried to replicate the test pattern for one of the other attributes. pushed another commit updating to the suggested changes. setting the conversation to resolved

fnagel (approved):

mgol (changes requested):
Thanks for the PR!

One question - shouldn't setting `aria-modal` be tied to the `modal` option on the dialog widget? See https://api.jqueryui.com/dialog/#option-modal. When `modal` is `false`, the intention seems to be that you can still interact with what's under the dialog so we shouldn't set `aria-modal`.

@fnagel what do you think?

ui/widgets/dialog.js
```diff
@@ -347,7 +347,8 @@ $.widget( "ui.dialog", {
 
 				// Setting tabIndex makes the div focusable
 				tabIndex: -1,
-				role: "dialog"
+				role: "dialog",
+				"aria-modal": "true"
```
mgol, line 351: If @fnagel agrees with my suggestion then this would need to be changed to only be set when `this.options.modal` is truthy. Since setting `"false"` is apparently equivalent to omitting the attibute, we should be able to just check this condition here:
```suggestion
				"aria-modal": this.options.modal ? "true" : "false"
```
We'd also need a test for both cases.

rpkoller: does that mean using the dialog with the modal option set to false is for cases integrating the div within a page while with the modal option set to true the dialog gets the modal behavior as an overlay? i've just tested your suggested change on an instance of drupal 11 and it is still a drop in change with no further changes necessary. so i would agree. i was just simply unaware of that modal option. but let's wait what @fnagel says. if he agrees i'll commit your suggestion.

mgol: Dialog traps focus - if you press TAB while the last tabbable element has focus, it focuses the first element in the dialog. If you have SHIFT pressed (so tabbing is done in reverse) and the first element is focused, it focuses the last focusable one:
https://github.com/jquery/jquery-ui/blob/9180a8180b17c38f6c3f27ba46d4546d800d3508/ui/widgets/dialog.js#L364-L384

This happens regardless of the `modal` option, perhaps a bit confusingly - you can test on https://jqueryui.com/resources/demos/dialog/animated.html. But you can still click on elements outside, it's just the TAB key that's blocked. I'm not sure that's intended but it does seem to suggest we shouldn't add `aria-modal` to such dialogs.

The overlay only appears if the `modal` flag is truthy and it's done in the `_createOverlay` method here:
https://github.com/jquery/jquery-ui/blob/9180a8180b17c38f6c3f27ba46d4546d800d3508/ui/widgets/dialog.js#L875-L877

That method checks the `modal` flag at the beginning:
https://github.com/jquery/jquery-ui/blob/9180a8180b17c38f6c3f27ba46d4546d800d3508/ui/widgets/dialog.js#L845-L848

This leads me to thinking the `aria-modal` attribute should reflect the `modal` option.

mgol: BTW, I'd tweak my suggestion - no need to set `"false"` when we can pass `null`; that's equivalent to removing the attribute:
```suggestion
				"aria-modal": this.options.modal ? "true" : null
```

fnagel: Interesting question that Michał opened here. I'm not quite sure on how to use it as I'm not really familiar with the `aria-modal` attribute but when reading some resources, there are some hints that the attribute should be used when its not possible to interact with the content outside the dialog (sometimes even mentioning an overlay / dimmer in the background). In some it is formulated quite clearly: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/ (see the "note" box at the bottom.

In my experience, all ARIA attributes can behave different for specific browser / AT setups but to me it looks like we should use the the approach @mgol suggested. Just to be sure we do not break stuff for some users.

tests/unit/dialog/core.js
```diff
@@ -70,11 +70,12 @@ QUnit.test( "title id", function( assert ) {
 } );
 
 QUnit.test( "ARIA", function( assert ) {
-	assert.expect( 4 );
+	assert.expect( 5 );
 
 	var element = $( "<div>" ).dialog(),
 		wrapper = element.dialog( "widget" );
 	assert.equal( wrapper.attr( "role" ), "dialog", "dialog role" );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "aria-modal attribute" );
```
rocketeerbkw, line 78: Discussed with @rpkoller, this is a possible change to the tests given the  suggested changes in https://github.com/jquery/jquery-ui/pull/2257/files#r1619122875.

``` javascript
if (element.dialog( "option", "modal" )) {
  assert.equal( wrapper.attr( "aria-modal" ), "true", "aria-modal attribute" );
} else {
  assert.ok( !element.attr( "aria-modal" ), "aria-modal not set" );
}
```

rpkoller:
@mgol thanks for the explanation. but one word about the example of animated.html. i'Ve tested with safari, yes tabbing is limited within the dialog modal,  you the user are unable to tab outside, but clicking outside the dialog modal you cant as well, at least it has not effect. "technically" from my understanding the general expection of clicking outside a dialog modal should actually close / collapse the modal (not only by clicking the close button). at the moment you can click everywher outside the dialog modal  but  it as no effect at all, for example the cursor is also not changing to a pointer hovering  the open dialog button. but i disagree about the usage of aria-modal in this situation, it should be added to such dialogs, alone for the fact if the dialog modal is open and you start the rotor in voiceover you have two buttons available; you have open dialog button  AND the close button both available under the form controls section as well as the buttons section. that is highly confusing already on a simple example page like that. and on a sidenot i've noticed that the dialog is also draggable. this isnt announced by the screenreader nor is the positions controlable via the keyboard (but that is completely out of the scope for this issue, just noticed while playing with the animated example).  

and one additional note from the mdn docs: 

> If a dialog is not modal — there is no inert background and focus isn't confined to the dialog — either include aria-modal="false" or omit the attribute altogether.

that would also support the point you'Ve  made with the w3c link @fnagel. and i also agreee in regards of the different behavior across browsers and AT setups. a not up to date overview is provided on here: https://a11ysupport.io/tech/aria/aria-modal_attribute

i've committed @mgols last suggestion, now using `"aria-modal": this.options.modal ? "true" : null` . in regards of updating the test and the requirement to also check the other cases. i've got some help from @rocketeerbkw at a local meetup last night where he came up with the suggestion he posted in the comments here. question to you two, @mgol and @fnagel, would that test be enough with that general else and the way the if is called or would it be necessary to be more granular and specific?

tests/unit/dialog/core.js
mgol: `if`s like that are not a good practice in test - you'd still be testing only one of them and the other would be left untested. We need both tests to run. We need one dialog with the `modal` option set - and then verify the `aria-modal` attribute is set to `"true"`, and another one without the `modal` option that verifies the attribute has not been set.

This test is failing right now because the dialog was created without the `modal` option.

I'd recommend creating a separate test just after this one titled `ARIA, aria-modal` where you create three dialogs - one with the `modal` option set to `true`, one to `false` and one with no such option passed, and verify proper `aria-model` value or presence for each of them.

This `"ARIA"` test should be reverted to its initial form, all new assertions would be added to this new test.

mgol:
> @mgol thanks for the explanation. but one word about the example of animated.html. i'Ve tested with safari, yes tabbing is limited within the dialog modal, you the user are unable to tab outside, but clicking outside the dialog modal you cant as well, at least it has not effect.

@rpkoller This is because the only thing you can interact with in this demo is the Open Dialog button and the dialog is already open. If you add another button with a handler to alert something, the alert will show up. So you definitely can interact with what's outside of the dialog without closing the dialog if the `modal` option is not enabled.

mgol (changes requested):

rpkoller:
apologies just noticed your comment today. i've reverted the ARIA test back to its initial state and i will try to come up with a test called aria-modal as you suggested. will give it a try tonight and see how far i get.

rpkoller:
please bare with me, took me a while, and as i've mentioned i am not a developer (my focus is in content, ux and a11y). but i've tried to piece together a draft for an additional test. added three assertions based on the setter in the api docs: https://api.jqueryui.com/dialog/#option-modal . not sure if that is the most elegant way. but for the first assertion i set the option to true and then check if the aria-modal attribute is there and set to true. for the option set to false i check if the aria-modal attribute is there as well but set to false. for the last i am not entirely sure but i think instead of yes or no i went with null assuming that would not add the modal option. in consequence the "value" for the aria-modal attribute is also null aka not added.  that was my rational behind the draft.

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,26 @@ QUnit.test( "ARIA", function( assert ) {
 	element.remove();
 } );
 
+QUnit.test( "aria-modal", function( assert ) {
+	assert.expect( 3 );
+
+	var element = $( "<div>" ).dialog( "options", "modal", true ),
```
mgol, line 90: It's `option`, not `options`, but you can see in checks on this PR you're getting test failures like:
> cannot call methods on dialog prior to initialization; attempted to call method 'options'

The widget needs to be initialized first. It'd help if you read https://learn.jquery.com/jquery-ui/how-jquery-ui-works/ which explains these basics.

Here, you should just create the widget passing proper options to it as described in the "Initialization" section of that doc.

That also made me realize we also need to modify the `_setOption` method of the dialog. Find it in the file, you'll need to add:
```js
		if ( key === "modal" ) {
			uiDialog.attr( "aria-modal", value ? "true" : null );
		}
```
in there. Unfortunately, it seems the one in `_createWrapper` is needed as well.

Then, we need tests for both the setting the option initially and then modifying it later to cover all grounds.

To summarize, I'd create three sections in this test - one with `modal` set to `true`: `.dialog( { modal: true } )`; one when it's set to `false`, and one when no options are passed: `.dialog()` (you still need to initialize like that before calling `.dialog( "option", "modal", true )`!)

And then, within each of those sections, first do a proper assertion and then call `.dialog( "options", "modal", VALUE )` substituting `VALUE` for these three options and check again if the value is as expected. That should give you total 12 assertions in this test.

mgol (changes requested):

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
 } );
 
+QUnit.test( "aria-modal", function( assert ) {
+	assert.expect( 12 );
+
+	var element = $( "<div>" ).dialog( { modal: true } ),
+		wrapper = element.dialog( "widget" );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.dialog( "option", "modal", false );
+	assert.ok( !wrapper.attr( "aria-modal" ), "modal option set to false, aria-modal attribute not added" );
```
mgol, line 94:
```suggestion
	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+	element.dialog( "option", "modal", null );
+	assert.ok( !wrapper.attr( "aria-modal" ), "modal option not set, aria-modal attribute not added" );
```
mgol, line 96:
```suggestion
	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+	element.dialog( "option", "modal", true );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.remove();
+
+	var element = $( "<div>" ).dialog( { modal: false } ),
+		wrapper = element.dialog( "widget" );
+	assert.ok( !wrapper.attr( "aria-modal" ), "modal option set to false, aria-modal attribute not added" );
```
mgol, line 103:
```suggestion
	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+	element.dialog( "option", "modal", true );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.dialog( "option", "modal", null );
+	assert.ok( !wrapper.attr( "aria-modal" ), "modal option not set, aria-modal attribute not added" );
```
mgol, line 107:
```suggestion
	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+	element.dialog( "option", "modal", false );
+	assert.ok( !wrapper.attr( "aria-modal" ), "modal option set to false, aria-modal attribute not added" );
```
mgol, line 109:
```suggestion
	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+	element.remove();
+
+	var element = $( "<div>" ).dialog(),
+		wrapper = element.dialog( "widget" );
+	assert.ok( !wrapper.attr( "aria-modal" ), "modal option not set, aria-modal attribute not added" );
```
mgol, line 114:
```suggestion
	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+	element.dialog( "option", "modal", true );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.dialog( "option", "modal", false );
+	assert.ok( !wrapper.attr( "aria-modal" ), "modal option set to false, aria-modal attribute not added" );
```
mgol, line 118:
```suggestion
	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
```

mgol (approved):
Looks good! Just a few suggestions for tests.

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+QUnit.test( "aria-modal", function( assert ) {
+	assert.expect( 12 );
+
+	var element = $( "<div>" ).dialog( { modal: true } ),
+		wrapper = element.dialog( "widget" );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.dialog( "option", "modal", false );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
+	element.dialog( "option", "modal", null );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
```
mgol, line 96: I guess we shouldn't really test setting `modal` to `null` since this is an option, not an `attr` call where that has a special meaning...
```suggestion
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.remove();
+
+	var element = $( "<div>" ).dialog( { modal: false } ),
+		wrapper = element.dialog( "widget" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
+	element.dialog( "option", "modal", true );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.dialog( "option", "modal", null );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
```
mgol, line 107:
```suggestion
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,43 @@ QUnit.test( "ARIA", function( assert ) {
...
+
+	var element = $( "<div>" ).dialog(),
+		wrapper = element.dialog( "widget" );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option not set, aria-modal attribute not added" );
+	element.dialog( "option", "modal", true );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.dialog( "option", "modal", false );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
+	element.dialog( "option", "modal", null );
+	assert.equal( wrapper.attr( "aria-modal" ), null, "modal option not set, aria-modal attribute not added" );
```
mgol, line 120:
```suggestion
```

mgol (commented):
Looks good! Just a few suggestions for tests.

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,37 @@ QUnit.test( "ARIA", function( assert ) {
 	element.remove();
 } );
 
+QUnit.test( "aria-modal", function( assert ) {
+	assert.expect( 12 );
```
mgol, line 88:
```suggestion
	assert.expect( 9 );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,37 @@ QUnit.test( "ARIA", function( assert ) {
 	element.remove();
 } );
 
+QUnit.test( "aria-modal", function( assert ) {
+	assert.expect( 9 );
+
```
mgol, line 89: Let's declare the variables once, that's what really happens anyway:
```suggestion

var element, wrapper;

```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,37 @@ QUnit.test( "ARIA", function( assert ) {
...
+	var element = $( "<div>" ).dialog( { modal: true } ),
+		wrapper = element.dialog( "widget" );
```
mgol, line 91:
```suggestion
	element = $( "<div>" ).dialog( { modal: true } );
	wrapper = element.dialog( "widget" );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,37 @@ QUnit.test( "ARIA", function( assert ) {
...
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.dialog( "option", "modal", false );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
+	element.dialog( "option", "modal", true );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.remove();
+
+	var element = $( "<div>" ).dialog( { modal: false } ),
+		wrapper = element.dialog( "widget" );
```
mgol, line 100:
```suggestion
	element = $( "<div>" ).dialog( { modal: false } );
	wrapper = element.dialog( "widget" );
```

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,37 @@ QUnit.test( "ARIA", function( assert ) {
...
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
+	element.dialog( "option", "modal", true );
+	assert.equal( wrapper.attr( "aria-modal" ), "true", "modal option set to true, aria-modal attribute added" );
+	element.dialog( "option", "modal", false );
+	assert.equal( wrapper.attr( "aria-modal" ), undefined, "modal option set to false, aria-modal attribute not added" );
+	element.remove();
+
+	var element = $( "<div>" ).dialog(),
+		wrapper = element.dialog( "widget" );
```
mgol, line 109:
```suggestion
	element = $( "<div>" ).dialog();
	wrapper = element.dialog( "widget" );
```

mgol (commented):
Looks good! Just a few suggestions for tests.

tests/unit/dialog/core.js
```diff
@@ -84,6 +84,39 @@ QUnit.test( "ARIA", function( assert ) {
 	element.remove();
 } );
 
+QUnit.test( "aria-modal", function( assert ) {
+	assert.expect( 9 );
+
+var element, wrapper;
```
mgol, line 90: Oops, I messed up indentation:
```suggestion
	var element, wrapper;
```

mgol:
@rpkoller I added a few minor test tweaks by myself and landed the PR. Thanks for going through the whole review process!

mgol:
@rpkoller This patch has been released as part of jQuery UI `1.14.0-beta.2`: The second beta to jQuery UI 1.14.0 has been released: https://blog.jqueryui.com/2024/06/jquery-ui-1-14-0-beta-2-released/. I'd appreciate testing if it works as intended! 🙂

mgol:
@rpkoller This patch has been released as part of jQuery UI `1.14.0-beta.2`: https://blog.jqueryui.com/2024/06/jquery-ui-1-14-0-beta-2-released/. I'd appreciate testing if it works as intended! 🙂

tests/unit/dialog/core.js
rpkoller: thanks a lot for the detailed explanation, that helped figuring things out, even though it was way out of my comfort zone. but with the help of @rocketeerbkw i was able to finish the draft thursday night.

rpkoller:
yay, and thanks to you @mgol for your patience and all the explanations along the way! and in regards of testing, there was a new MR created in drupal for beta2. i've tested the MR now, it was applied correctly, but somehow the aria-modal wasn't added. As you know i am not a developer therefore i never really minded the technical details on the drupal end implementation wise. but I've asked on the issue how the dialog modal is initialized in drupal and also referenced this PR. I'll keep you posted as soon as i get any feedback.

rpkoller:
just a brief update. things look good. beta2 got merged into drupal 11.x three days ago. the only change necessary, `'aria-modal': this.options.modal ? 'true' : null,`had to be added since drupal has a custom wrapper for dialog.js. with that in place everything works like a charm. the aria-modal attribute is getting added if the modal option is set to true and all the existing functionality still works as expected.

mgol:
@rpkoller thanks for the update!

PR #2258, opened 2024-05-29
Build: Test on more jQuery Core versions

mgol:
The initial 1.14.0 announcement only promised support for the latest version of jQuery Core within each major. The main goal was to avoid testing on too many old versions. However, just testing on the latest one seems too drastic. For now, add `3.6.4` to the mix; we'll work on a more detailed policy soon.


fnagel (approved):

timmywil:
Was there an issue related to jQuery 3.6.4 that prompted this?

timmywil (approved):

PR #2260, opened 2024-06-04
Tests: conslidate GH actions runs; separate git & stable jQuery versions

timmywil:
- also add hard retries to runs

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -17,18 +17,17 @@ env:
 jobs:
   build-and-test:
     runs-on: ubuntu-latest
-    name: ${{ matrix.BROWSER }} - jQuery ${{ matrix.JQUERY }}
+    name: |
+      ${{ matrix.BROWSER }} | ${{ matrix.JQUERYS.name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -17,18 +17,17 @@ env:
...
-          - "git"
-          - "3.x-git"
-          - "3.7.1"
-          - "3.6.4"
-          - "2.2.4"
-          - "1.12.4"
+        JQUERYS:
+          - versions: --jquery git --jquery 3.x-git
+            name: jQuery git
+          - versions: --jquery 3.7.1 --jquery 3.6.4 --jquery 2.2.4 --jquery 1.12.4
```
mgol, line 29: Is it possible to make it a global variable to not repeat it 3 times or is the answer still no?

mgol (approved):
LGTM with one question

.github/workflows/node.js.yml
timmywil: Unfortunately, no. Not in a matrix.

Issue #2178, opened 2023-07-24
Datepicker: programmatically destroy don't work anymore


differens:
After upgrading from version 1.12.1 to 1.13.2 the method datepicker( "destroy" ) do not close and destroy the calendar ui. 

I have prepared an example at the link below. 

https://jsfiddle.net/7kdwvgp8/

After starting the execution just open the calendar and wait for the timeout which calls the destroy after 5 seconds. The result is that the calendar remains open and each iteration leads to an error

![es--anomalie](https://github.com/jquery/jquery-ui/assets/71696303/cef1c364-443b-41ab-81a0-c1fdf8df14b0)

Problem not occurs in 1.12.1.

Thank you in advance.

porterclev:
I'm working on this

DawnSovereign:
I will be helping to contribute to this.

tchan102:
I will be working on this issue!

porterclev:
@fnagel When comparing the two versions, we found that on version 1.12.1, the widget will only hide after being destroyed once the user clicks on another part of their screen. Would you prefer to have the widget disappear when destroyed or only hide the date picker once the user clicks away?

1.12.1 Functionality
![Sequence 01](https://github.com/jquery/jquery-ui/assets/116387727/1fa1ee9f-5384-4817-9a1f-628919b24028)

Proposed Fix
![fix ](https://github.com/jquery/jquery-ui/assets/116387727/6749e2ea-2604-47d8-b533-fc5e156598a2)

mgol:
Note that the description here is not fully correct - in 1.12.1, the datepicker UI was not hidden on destroy either! It was just hidden after the first mousedown. If you just hovered over the datepicker or clicked on any of its elements, you'd get lots of console errors even in 1.12.1.

But yes, the 1.13.2-1.13.3 behavior is worse since there's no way to get rid of the widget from the user perspective, irrespective of the underlying errors.

More info at https://github.com/jquery/jquery-ui/pull/2268#issuecomment-2262670431.

PR #1852, opened 2018-01-31
datepicker: memory leak

jigar140291:
Ticket: [15250](https://bugs.jqueryui.com/ticket/15270)

PR #2268, opened 2024-07-10
Datepicker: Hide the UI on destroy

porterclev:
Issue: https://github.com/jquery/jquery-ui/issues/2178 - date picker remains after being destroyed

We confirmed through git bisect that this commit, https://github.com/jquery/jquery-ui/commit/817ce38,  introduced a bug while fixing a memory leak with the date picker. This commit sets the current instance to null, https://github.com/jquery/jquery-ui/blob/5665215a8560832193735d9507a90d10e03b90c8/ui/widgets/datepicker.js#L440, to remove the date picker logic but fails to remove the UI widget. 

Our proposed fix is to hide the date picker when the destroy function using `$.datepicker._hideDatepicker();` so the user can no longer hover over a "ghost" date picker that can no longer be interacted with.

### DIFF
--- a/tests/unit/datepicker/methods.js
+++ b/tests/unit/datepicker/methods.js
@@ -11,7 +11,7 @@ var beforeAfterEach = testHelper.beforeAfterEach;
 QUnit.module( "datepicker: methods", beforeAfterEach()  );
 
 QUnit.test( "destroy", function( assert ) {
-	assert.expect( 35 );
+	assert.expect( 39 );
 	var inl,
 		inp = testHelper.init( "#inp" ),
 		dp = $( "#ui-datepicker-div" );
@@ -21,6 +21,15 @@ QUnit.test( "destroy", function( assert ) {
 	assert.equal( dp.css( "display" ), "block", "Datepicker - visible" );
 	inp.datepicker( "hide" ).datepicker( "destroy" );
 	assert.ok( $.datepicker._curInst == null, "Datepicker - destroyed and cleared reference" );
+	assert.equal( dp.css( "display" ), "none", "Datepicker - absent" );
+
+	// Destroy Without Hiding
+	inp = testHelper.init( "#inp" );
+	inp.datepicker( "show" );
+	assert.equal( dp.css( "display" ), "block", "Datepicker - visible" );
+	inp.datepicker( "destroy" );
+	assert.ok( $.datepicker._curInst == null, "Datepicker - destroyed and cleared reference" );
+	assert.equal( dp.css( "display" ), "none", "Datepicker - absent" );
 
 	inp = testHelper.init( "#inp" );
 	assert.ok( inp.is( ".hasDatepicker" ), "Default - marker class set" );
--- a/tests/unit/datepicker/methods.js
+++ b/tests/unit/datepicker/methods.js
@@ -23,7 +23,7 @@ QUnit.test( "destroy", function( assert ) {
 	assert.ok( $.datepicker._curInst == null, "Datepicker - destroyed and cleared reference" );
 	assert.equal( dp.css( "display" ), "none", "Datepicker - absent" );
 
-	// Destroy Without Hiding
+	// Destroy without manual hiding (ensure datepicker is hidden after calling destroy)
 	inp = testHelper.init( "#inp" );
 	inp.datepicker( "show" );
 	assert.equal( dp.css( "display" ), "block", "Datepicker - visible" );
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -423,7 +423,6 @@ $.extend( Datepicker.prototype, {
 
 		nodeName = target.nodeName.toLowerCase();
 		$.removeData( target, "datepicker" );
-		$.datepicker._hideDatepicker();
 		if ( nodeName === "input" ) {
 			inst.append.remove();
 			inst.trigger.remove();
@@ -436,9 +435,11 @@ $.extend( Datepicker.prototype, {
 			$target.removeClass( this.markerClassName ).empty();
 		}
 
+		inst.dpDiv.remove();
 		if ( datepicker_instActive === inst ) {
 			datepicker_instActive = null;
 			this._curInst = null;
+			this.dpDiv = null;
 		}
 	},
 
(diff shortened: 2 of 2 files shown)


porterclev:
@fnagel

mgol (changes requested):
Thanks for the PR.

If the DOM is not getting destroyed, we should make sure it is after the fix. Is just hiding enough to achieve that?

I'd also want to understand fully why just setting `this._curInst` to `null` caused this issue; without this, the fix may be incomplete. Can you analyze this & report back?

Also, we'll need tests.

DawnSovereign:
The following text is the understanding we have of what happened. The date picker was getting destroyed, but a lingering reference to it was kept alive (curInst). This meant it was a memory leak. The PR fixed the memory leak by setting curInst to null but the datepicker remained open which we assume because other parts of the code accessed curInst without checking if it's null(?)

To answer your question about setting this.curInst to null; curInst is being set to null, but the datepicker still thinks it is open, so other parts of the code still access _curInst which is why we set this.curInst to null.

In order to directly address how we could get rid of the daterpicker UI that's still open, we get rid of the div instance of it and set the instance to null. 

Here are the screenshots with our fix for before and after the div instance is destroyed. 
Screenshot 2024-07-15 100202
Screenshot 2024-07-15 100147

OmarShehata:
> Here is a screenshot of the stack trace:

@DawnSovereign are you able to get this stack trace with the local, unminified code? or click on it and see what function it's in, and find the corresponding function in the source code? 

from the screenshot, it looks like it's this function, right?

https://github.com/jquery/jquery-ui/blob/5e4f1b86296c5a977abc7efaff671ea9004d4d37/ui/widgets/datepicker.js#L2166-L2177

mgol (changes requested):
I'll review the code logic in more detail later but I'll reiterate this needs tests before we can merge anything.

mgol:
You can also see that existing tests are not passing, please make sure they do.

porterclev:
@mgol  Thanks for taking a look. We've pushed a unit test that reproduces the issue. It fails in `main` and passes in this branch with the fix.

We've reverted the fix back to hiding the datepicker instead of destroying the div. This is because several tests expect to re-use the element after it is destroyed. If we want to properly remove the element when the datepicker is destroyed, it looks like this may be a breaking change? I can open a ticket for that separately with more details.

The underlying issue is that destroying the datepicker by itself does not hide it. The current tests in `main` don't show this problem because they always hide the datepicker first before destroying it. So it is possible for a user to work around this problem at the moment by hiding before destroying.

The long term solution is to remove the element when the datepicker is destroyed, and update the tests, and mark this as a (potential?) breaking change.

mgol (approved):
LGTM. @fnagel can you have a look as well?

tests/unit/datepicker/methods.js
```diff
@@ -21,6 +21,15 @@ QUnit.test( "destroy", function( assert ) {
 	assert.equal( dp.css( "display" ), "block", "Datepicker - visible" );
 	inp.datepicker( "hide" ).datepicker( "destroy" );
 	assert.ok( $.datepicker._curInst == null, "Datepicker - destroyed and cleared reference" );
+	assert.equal( dp.css( "display" ), "none", "Datepicker - absent" );
+
+	// Destroy Without Hiding
```
fnagel, line 26: Can we use something more describing`? Maybe like:

`// Destroy without manual hiding (ensure datepicker is hidden after calling destroy)`

fnagel (changes requested):

fnagel:
At first I thought that just hiding the elements instead of destroying sounds not ideal, but then I've seen the comment about "This is because several tests expect to re-use the element after it is destroyed." so the change makes sense to me. 

It's not ideal that the destroy method does not really destroy everything (elements, events, ...) like other widgets do, but the Datepicker does not use the widget factory and always was a little special. 

So to me, this change improves the current state and fixes a bug introduces with 1.13.x. Only the test comment is a little confusing (see my review).

@mgol We might want to squash the commits when merging this.

tests/unit/datepicker/methods.js
mgol: Let's do it now:
```suggestion
	// Destroy without manual hiding (ensure datepicker is hidden after calling destroy)
```
(yes, I'll squash it when merging)

mgol:
@fnagel Was the comment your only concern to address before merging? I get your comment about destroying; I had a look at the code and it seemed not that easy to retrofit it into the existing design. `_hideDatepicker` may use an animation and we'd have to destroy only when it's done, but it doesn't provide any hook to do so. We'd have to add one, then add tests, also fix existing tests to not depend on it existing... Maybe let's leave it as it is for now.

Please approve if you have no further concerns to address.

tests/unit/datepicker/methods.js
fnagel: Perfect!

fnagel (approved):

fnagel:
@mgol I'm fine with merging it as it is. Datepicker was always special (as mentioned before, no widget factory in use) and SHOULD have been replaced with the new calendar and datepicker widget but never has.

Not sure what little surprises will come up when we start changing how the structure and design currently looks ;-)

mgol:
Just to make sure, I checked myself what was the cause of the original issue. Datepicker attaches a global `mousedown` on document:
https://github.com/jquery/jquery-ui/blob/cd41c45d917d18839ab800816070db09a359d5bf/ui/widgets/datepicker.js#L2201
It never removes it. This handler hides the datepicker in certain conditions - but the first thing it checks is whether `$.datepicker._curInst` is truthy:
https://github.com/jquery/jquery-ui/blob/cd41c45d917d18839ab800816070db09a359d5bf/ui/widgets/datepicker.js#L1005-L1022
Since #1852 made `_curInst` cleared, this method stopped doing anything.

Some parts of the description from #2178 are incorrect, though - in 1.12.1, the datepicker UI was _not hidden_ on destroy either! It was just hidden after the first `mousedown`. If you just hovered over the datepicker or clicked on any of its elements, you'd get lots of console errors even in 1.12.1.

Therefore, this PR actually changes the behavior compared to 1.12.1 which was already partially broken. I think it's a reasonable change - the datepicker UI should hide when destroying the widget - but let's call it out.

PR #2273, opened 2024-07-29
Tests: Update jQuery Migrate from 3.4.1 to 3.5.2, simplify Migrate setup

mgol:
Apart from updating jQuery Migrate, the setup is now changed so that exact Migrate versions are only references when fetching npm packages; other than that, they're represented as `jquery-migrate-1.x` & `jquery-migrate-3.x`. This will allow smaller changes when updating jQuery Migrate in the future.


Krinkle (approved):

fnagel (commented):
+1

PR #2274, opened 2024-07-29
Demos: Make autocomplete demos work on jqueryui.com

mgol:
The `jqueryui.com` demos build process deAMDifies HTML files, replacing required JS files with direct script tags. On the other hand, when running demos locally from the jQuery UI repository, RequireJS is used.

This used to work fine until we got a new `search.js` file introduced in gh-2187. The deAMDifying process doesn't touch non-HTML files which made loading autocomplete demos crash on "require is not a function"

To resolve the issues without a major rearchitecture of the build process, the `search.js` file now detects AMD and uses `require`, falling back to relying on the `jQuery` global in the other case.

Ref gh-2187
Ref jquery/jqueryui.com#216

### DIFF
--- a/demos/search.js
+++ b/demos/search.js
@@ -1,5 +1,18 @@
-require( [ "jquery", ], function( $ ) {
+( function( factory ) {
 	"use strict";
+
+	if ( typeof define === "function" && define.amd ) {
+
+		// Fetch jQuery as an AMD dependency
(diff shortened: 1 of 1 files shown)


Krinkle (approved):

fnagel (commented):
+1

Issue #2271, opened 2024-07-26
Make the uiDialogTitle configurable


rpkoller:
With https://github.com/jquery/jquery-ui/issues/2246 the element in the background of the dialog are getting removed from the AOM. The problem the title is just wrapped in a span which isn't semantic. The title is properly announced when dialog is opened (tested on drupal) but if the content of that modal doesnt contain any heading elements if you open up the rotor in for example voiceover the headings section wont be available at all. this might be potentially confusing since you are unable to get the context the dialog is about. there are no headings available and with VO-shift-i you only get announced the title and stats of the hidden page not the dialog modal. technically you would have to step through the modal with the voiceover cursor since the title isnt focusable or close the modal and reopen it again to get it reannounced. 

which element to use instead of the generic span is not specified in the wcag, see this comment for context
https://github.com/w3c/wcag/discussions/2722#discussioncomment-3842484 . according to the comment an h1 or any other heading level element is perfectly fine. the aria apg example uses an h2 for example: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/ while in a discussion on the a11y slack https://web-a11y.slack.com/archives/C042TSFGN/p1708471859401949  the recommendation by a few folks to go with an h1. with the modal dialog open only the content of the modal is accessible the overall heading structure of the page is irrelevant. and james scholes emphasized that the modal title should be h1. there is no reason to use an h2 since all it does is to take away one heading level from the modal content (james is a screen reader user). 

i "think" i've raised the aspect of the html element wrapping the title before in the context of the linked issue or the corresponding PR but unable to find the detail. but since it is still debatable which heading level to pick as well as the aspect of backwards compatability i wonder if it would be feasible to make the html element wrapping the title configurable? so adding somthing like titleElement to the options and set it to "span" per default that way it would be a none breaking change. the user would be able to to change the element proactively via the options?

mgol:
Thanks for the report. I'd rather avoid adding more options as then we need to maintain them indefinitely. Since h1 seems to be OK per your description, I'd try that.

It'd be good to get that into 1.14.0 just in case someone considers this a compatibility break.

Would you like to submit a PR? I think we'd just need updating the wrapping `span` to a `h1` in the `uiDialogTitle` definition and maybe add a simple test checking that the title is wrapped in a `h1`.

mgol:
[...]

PR #2275, opened 2024-07-31
Dialog: Add option to set the html element for the dialog title

rpkoller:
PR for https://github.com/jquery/jquery-ui/issues/2271. Initial draft adding an option based on the naming @mgol suggested on the issue (`uiDialogTitleTagName `) and using `span` as the default. Currently working on adding tests as well in the next step.

### DIFF
--- a/ui/widgets/dialog.js
+++ b/ui/widgets/dialog.js
@@ -81,6 +81,7 @@ $.widget( "ui.dialog", {
 		resizable: true,
 		show: null,
 		title: null,
+		uiDialogTitleTagName: "span",
 		width: 300,
 
 		// Callbacks
@@ -437,7 +438,7 @@ $.widget( "ui.dialog", {
 			}
 		} );
 
-		uiDialogTitle = $( "<span>" ).uniqueId().prependTo( this.uiDialogTitlebar );
+		uiDialogTitle = $( "<" + this.options.uiDialogTitleTagName + ">" ).uniqueId().prependTo( this.uiDialogTitlebar );
 		this._addClass( uiDialogTitle, "ui-dialog-title" );
 		this._title( uiDialogTitle );
 


rpkoller:
Ok i've finished an initial draft. I am not sure if the multiline syntax and formatting i've used is correct but i had to satisfy the linter and get to a line length < 100 characters. and i've also added a test. i've created a modal dialog and set the desired tag name to h2. i've picked an html tag != the default value `span`. I then used the queryselector to pick the title element which has the class name of `.ui-dialog-title` and with the `tagName` property i've checked if that element is the given `h2`. and i thought it is unnecessary to check several different scenarios like for the aria-modal test. i thought a single one should be enough here i hope?

rpkoller:
The good thing, the test works (the expected value is an h2 but the actual value is still a span). so my fix for the line length linting error isn't working (as a one liner it works locally).

rpkoller:
Got the test to green with the help of @rocketeerbkw. the only thing we were not sure if it was necessary to add the new option to the common-deprecated.js file as well.

ui/widgets/dialog.js
```diff
@@ -437,7 +438,8 @@ $.widget( "ui.dialog", {
 			}
 		} );
 
-		uiDialogTitle = $( "<span>" ).uniqueId().prependTo( this.uiDialogTitlebar );
+		uiDialogTitle = $( this.options.uiDialogTitleTagName )
```
mgol, line 441: Passing user input directly to the jQuery factory is dangerous - apart from obvious abuse when people pass selectors instead of tags here, they can e.g. pass `<script>`, etc. Protecting from that would require significantly more code, so I'm thinking we should actually change the approach.

How about introducing an option called `uiDialogTitleHeadingLevel` with the default of `0` and possible values all integers between `0` & `6`; you can use `Number.isInteger` for that. We can check the value and revert to `0` if it's invalid. Then, for `0` we use `<span>` but for other values we use a proper heading level.

That should cover the important use cases here and avoid possible security issues. Thoughts?

tests/unit/dialog/common-deprecated.js
```diff
@@ -34,6 +34,7 @@ common.testWidget( "dialog", {
 		resizable: true,
 		show: null,
 		title: null,
+		uiDialogTitleTagName: "<span>",
```
mgol, line 37: Yes, this needs to be added here as well. 👍🏻 

Most modules have regular & deprecated test modes and run tests in both. The deprecated one loads `common-deprecated.js` instead of `common.js` and it loads an additional `deprecated.js` file:
https://github.com/jquery/jquery-ui/blob/02a6e6bcb004696b21324ef14c07374a56d7cd02/tests/lib/bootstrap.js#L155-L164

mgol (changes requested):
Thinking about it more, I have doubts about this approach. See my comments.

ui/widgets/dialog.js
rpkoller: in my initial draft i kept the <> out of the option but thought it would over complicate things so i ended up with the current version. but i agree that your suggested approach would be the better choice. would something like that work? 

```function between ( headingLvl ) {
  return Number.isInteger( headingLvl ) && headingLvl > 0 && headingLvl <= 6 ? "h" + headingLvl : "span";
};
uiDialogTitle = $( "<" + between( this.options.uiDialogTitleHeadingLevel ) + ">" ).uniqueId().prependTo( this.uiDialogTitlebar );```

i am not sure if it would be a clean code wise and i also dont know what the preferred name for the function would be. but the shown snippet is at least already working: the only difference to your suggestion is that the condition is only true between 1 and 6 everything else 0 included becomes a span. 

and i would also add a few test cases.

mgol: That works, but since this logic is quite simple and only used in a single place, it's not necessary to create a function for it, I think. You can just define a local variable with the computed tag name & use it below.

Try to keep lines ideally below 80 characters (and definitely below 100), breaking them into multiple lines when necessary.

And yes, a few test cases would be needed, when the option was:
1. not passed (testing default behavior).
2. `0`
3. `1`
4. `6`
5. `7` (testing that it reverts to `span`)
6. `"foo"` (testing that it reverts to `span`)

Those should be enough, I think. Optionally, instead of testing `1` & `5` individually, you can create a loop through values 1-6, but I don't think it's absolutely necessary.

rpkoller: ok thanks! the suggested changes are in place and the tests are green.

mgol (approved):
LGTM. I'd like one addtional review but we're not in a hurry - 1.14.0 was released today so the next release won't happen immediately.

We also need docs for this new option before we can include it. The docs for dialog are at https://api.jqueryui.com/dialog/. Can you document it? It'd be a similar entry to `closeOnEscape` here:
https://github.com/jquery/api.jqueryui.com/blob/1c46b8ea6ca5b0220856649283c049c4946cb391/entries/dialog.xml#L134-L137

We also need a note that 1.13.1 is the first release with this API, like the one we have here: https://api.jqueryui.com/datepicker/#option-onUpdateDatepicker. This is triggered by adding the `<added>` tag; see https://github.com/jquery/api.jqueryui.com/blob/1c46b8ea6ca5b0220856649283c049c4946cb391/entries/datepicker.xml#L461-L465 for the source of it.

rpkoller:
so you mean creating a pull request for api.jqueryui..com including the note about the initial release in 1.14.1? i can create that initial draft over there.

mgol:
Just a PR will be fine. We can have it reference the corresponding UI issue + PR in the commit message.

rpkoller:
oki, and should i omit using an issue number in the branch name?

mgol:
The branch name doesn’t matter, you can use whatever is convenient for you.

fnagel (approved):
Looks good to me. Thanks for your contribution!

mgol:
Thanks, merged! The change will arrive in jQuery UI 1.14.1.

rpkoller:
Thank you! and one question do you already have a rough plan when you want to release 1.14.1? i ask if there is a slight chance it might be released before the first beta for the next minor release drupal (11.1) arrives , which is planned for november 11th?

mgol:
@rpkoller yeah, a release in October is likely if nothing critical comes out and the milestone (https://github.com/jquery/jquery-ui/milestone/8) is clean; currently, there's one issue to fis there, but it's simple.

rpkoller:
that would be perfect, thank you for the headsup!

mgol:
@rpkoller jQuery UI 1.14.1 with this change has been released.

Issue #2277, opened 2024-08-05
jQueryUI Dialog - content shrinks on resizing when box-sizing: content-box


Daniel-Garmig:
Dialog content shrinks when dialog have scrollbars and is resized.
![jQueryUI_DialogResizeError](https://github.com/user-attachments/assets/5fb131e5-e20e-4d23-90e2-e1e5c92225f5)

This is a similar problem as in #1979 and #2083. 
It was marked as solved in jQueryUI 1.13.3 but it still happen on some cases (when `box-sizing: content-box`).

I have done the following tests:
| jQueryUI Version | box-sizing | Work status |
|--------|--------|--------|
| jQueryUI 1.13.2 | border-box | ❌ Error |
| jQueryUI 1.13.2 | content-box | ❌ Error |
| jQueryUI 1.13.3 | border-box | ✔️ Working |
| jQueryUI 1.13.3 |  content-box | ❌ Error | 

Bug only seems to happen on Chromium-based browsers (tested on Chrome 127.0.6533.89 and Edge 127.0.2651.86). It works fine on Firefox (tested on Firefox 128.0.3)

Sample fiddle: https://jsfiddle.net/67j9nm43/1/
Fiddle uses jQueryUI 1.13.3. You may change box-sizing between `border-box`/`content-box` to test both cases.

I know this is kind of a edge case, as most people today uses `box-sizing: border-box`, but some old projects may still use `box-sizing: content-box` as it is the default option.

mgol:
Thanks for the report with a clear test case! Since the issue is already in 1.12, given limited team resources it's not likely to be fixed by the UI team; see the project status at https://blog.jqueryui.com/2021/10/jquery-maintainers-update-and-transition-jquery-ui-as-part-of-overall-modernization-efforts/. PRs are welcome if they're not too complex and contain tests.

Daniel-Garmig:
I'm working on a fix for this. I will make a PR soon.

---

Here I give some more details about what I found.
It's a weird case when scrollbars are visible on Chromium browsers. It's like size is reserved for the scrollbar so content size is reduced.
The problem is that this behavior is different between Chromium based browsers and Firefox.

I have make the following test case:

Chrome 127.0.6533.119:
![ScrollbarSizeChrome](https://github.com/user-attachments/assets/f85ebab4-8410-433d-907d-c1429dbf10ab)

Firefox 129.0:
[...]

PR #2281, opened 2024-08-20
Resizable: Fix content shrink on resize

Daniel-Garmig:
Fixes: #2277. 
This end up being a bug on "resizable", and as I have shown on the issue it's quite a weird behavior and it's not consistent across browsers.

The same bug applies to the resizable element and "alsoResize" elements, so I fixed it on both places.

Dialog widget uses "alsoResize" property to update dialog contents size (as seen on the initial issue). This PR fixes that as well.

I have added some tests for both "box-sizing: content-box" and "box-sizing: border-box".

### DIFF
--- a/tests/unit/resizable/core.js
+++ b/tests/unit/resizable/core.js
@@ -246,107 +246,28 @@ QUnit.test( "nested resizable", function( assert ) {
 
 QUnit.test( "Resizable with scrollbars and box-sizing: border-box", function( assert ) {
 	assert.expect( 4 );
-
-	$( "<style> * { box-sizing: border-box; } </style>" ).appendTo( "#qunit-fixture" );
-
-	// Both scrollbars
-	var elementContent = $( "<div>" )
-			.css( {
-				width: "200px",
-				height: "200px",
-				padding: "10px",
-				border: "5px",
-				borderStyle: "solid",
-				margin: "20px"
-			} )
-			.appendTo( "#resizable1" ),
-		element = $( "#resizable1" ).css( "overflow", "auto" ).resizable(),
-		handle = ".ui-resizable-se";
-
-	testHelper.drag( handle, 10, 10 );
-	assert.equal( element.width(), 110, "element width (both scrollbars)" );
-	assert.equal( element.height(), 110, "element height (both scrollbars)" );
-
-	// Single (vertical) scrollbar.
-	elementContent.css( "width", "50px" );
-	testHelper.drag( handle, 10, 10 );
-	assert.equal( element.width(), 120, "element width (only vertical scrollbar)" );
-	assert.equal( element.height(), 120, "element height (only vertical scrollbar)" );
+	testResizableWithBoxSizing( assert, true, false );
 } );
 
 QUnit.test( "Resizable with scrollbars and box-sizing: content-box", function( assert ) {
 	assert.expect( 4 );
-
-	$( "<style> * { box-sizing: content-box; } </style>" ).appendTo( "#qunit-fixture" );
-
-	// Both scrollbars
-	var widthBefore, heightBefore,
-		elementContent = $( "<div>" )
-			.css( {
-				width: "200px",
-				height: "200px",
-				padding: "10px",
-				border: "5px",
-				borderStyle: "solid",
-				margin: "20px"
-			} )
-			.appendTo( "#resizable1" ),
-		element = $( "#resizable1" ).css( "overflow", "auto" ).resizable(),
-		handle = ".ui-resizable-se";
-
-	// In some browsers scrollbar may change element size (when "box-sizing: content-box")
-	widthBefore = element.innerWidth();
-	heightBefore = element.innerHeight();
-
-	testHelper.drag( handle, 10, 10 );
-	assert.equal( parseFloat( element.innerWidth() ), widthBefore + 10, "element width (both scrollbars)" );
-	assert.equal( parseFloat( element.innerHeight() ), heightBefore + 10, "element height (both scrollbars)" );
-
-	// Single (vertical) scrollbar.
-	elementContent.css( "width", "50px" );
-
-	testHelper.drag( handle, 10, 10 );
-	assert.equal( parseFloat( element.innerWidth() ), widthBefore + 20, "element width (only vertical scrollbar)" );
-	assert.equal( parseFloat( element.innerHeight() ), heightBefore + 20, "element height (only vertical scrollbar)" );
+	testResizableWithBoxSizing( assert, false, false );
 } );
 
 QUnit.test( "Resizable with scrollbars, a transform and box-sizing: border-box", function( assert ) {
 	assert.expect( 4 );
-
-	$( "<style> * { box-sizing: border-box; } </style>" ).appendTo( "#qunit-fixture" );
-
-	// Both scrollbars
-	var elementContent = $( "<div>" )
-		.css( {
-			width: "200px",
-			height: "200px",
-			padding: "10px",
-			border: "5px",
-			borderStyle: "solid",
-			margin: "20px"
-		} )
-		.appendTo( "#resizable1" ),
-		element = $( "#resizable1" ).css( { overflow: "auto", transform: "scale(1.5)" } ).resizable(),
-		handle = ".ui-resizable-se";
-
-	testHelper.drag( handle, 10, 10 );
-	assert.equal( element.width(), 110, "element width (both scrollbars)" );
-	assert.equal( element.height(), 110, "element height (both scrollbars)" );
-
-	// Single (vertical) scrollbar.
-	elementContent.css( "width", "50px" );
-	testHelper.drag( handle, 10, 10 );
-	assert.equal( element.width(), 120, "element width (only vertical scrollbar)" );
-	assert.equal( element.height(), 120, "element height (only vertical scrollbar)" );
+	testResizableWithBoxSizing( assert, true, true );
 } );
 
 QUnit.test( "Resizable with scrollbars, a transform and box-sizing: content-box", function( assert ) {
 	assert.expect( 4 );
+	testResizableWithBoxSizing( assert, false, true );
+} );
 
-	$( "<style> * { box-sizing: content-box; } </style>" ).appendTo( "#qunit-fixture" );
-
-	// Both scrollbars
+function testResizableWithBoxSizing( assert, isBorderBox, applyScaleTransform ) {
 	var widthBefore, heightBefore,
+		cssBoxSizing = isBorderBox ? "border-box" : "content-box",
+		cssTrasform = applyScaleTransform ? "scale(1.5)" : "",
 		elementContent = $( "<div>" )
 			.css( {
 				width: "200px",
@@ -357,13 +278,16 @@ QUnit.test( "Resizable with scrollbars, a transform and box-sizing: content-box"
 				margin: "20px"
 			} )
 			.appendTo( "#resizable1" ),
-		element = $( "#resizable1" ).css( { overflow: "auto", transform: "scale(1.5)" } ).resizable(),
+		element = $( "#resizable1" ).css( { overflow: "auto", transform: cssTrasform } ).resizable(),
 		handle = ".ui-resizable-se";
+
+	$( "<style> * { box-sizing: " + cssBoxSizing + "; } </style>" ).appendTo( "#qunit-fixture" );
 
 	// In some browsers scrollbar may change element size (when "box-sizing: content-box")
 	widthBefore = element.innerWidth();
 	heightBefore = element.innerHeight();
 
+	// Both scrollbars
 	testHelper.drag( handle, 10, 10 );
 	assert.equal( parseFloat( element.innerWidth() ), widthBefore + 10, "element width (both scrollbars)" );
 	assert.equal( parseFloat( element.innerHeight() ), heightBefore + 10, "element height (both scrollbars)" );
@@ -374,6 +298,6 @@ QUnit.test( "Resizable with scrollbars, a transform and box-sizing: content-box"
 	testHelper.drag( handle, 10, 10 );
 	assert.equal( parseFloat( element.innerWidth() ), widthBefore + 20, "element width (only vertical scrollbar)" );
 	assert.equal( parseFloat( element.innerHeight() ), heightBefore + 20, "element height (only vertical scrollbar)" );
-} );
+}
 
 } );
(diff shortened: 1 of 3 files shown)


melloware (approved):
awesome!

melloware:
@Daniel-Garmig do you know a way I can "MonkeyPatch" your fix on top of existing jQuery UI?  We do stuff like this in our app to patch other jquery UI functions but the magic is never quite the same for all patches.

https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js

I would like to be able to apply this patch until its accepted by jQueryUI team or (gulp) rejected.

tests/unit/resizable/core.js
```diff
@@ -244,4 +244,57 @@ QUnit.test( "nested resizable", function( assert ) {
...
+				height: 200,
+				padding: 10,
+				border: 5,
+				borderStyle: "solid"
+			} )
+			.appendTo( "#resizable1" ),
+	element = $( "#resizable1" ).css( { overflow: "auto" } ).resizable(),
+	handle = ".ui-resizable-se";
+
+	$( "*" ).css( "box-sizing", "border-box" );
```
mgol, line 262: That's unnecessarily slow & not clean - every element is selected and has an inline style added to set `border-box`. If you want a test with global `border-box`, add it via a `<style>` tag.

The same applies to other places.

tests/unit/resizable/core.js
```diff
@@ -244,4 +244,57 @@ QUnit.test( "nested resizable", function( assert ) {
...
+	var elementContent = $( "<div>" )
+			.css( {
+				width: 50,
+				height: 200,
+				padding: 10,
+				border: 5,
+				borderStyle: "solid"
+			} )
+			.appendTo( "#resizable1" ),
+	element = $( "#resizable1" ).css( { overflow: "auto" } ).resizable(),
```
mgol, line 259:
```suggestion
	element = $( "#resizable1" ).css( "overflow", "auto" ).resizable(),
```
Similarly in other tests.

tests/unit/resizable/core.js
```diff
@@ -244,4 +244,57 @@ QUnit.test( "nested resizable", function( assert ) {
...
 } );
 
+QUnit.test( "Resizable with scrollbars and box-sizing: border-box", function( assert ) {
+	assert.expect( 2 );
+
+	var elementContent = $( "<div>" )
+			.css( {
+				width: 50,
+				height: 200,
+				padding: 10,
```
mgol, line 254: Can you also set some margin to make sure that doesn't interfere?

Also, provide explicit values with the `px` suffix, not numbers. Passing numbers & depending on them getting `px` auto-appended is deprecated and disabled for most CSS properties in jQuery 4.0.0.

The properties here work (see the regex at https://github.com/jquery/jquery/blob/4.0.0-beta.2/src/css/isAutoPx.js) but this is mostly for backwards compat reasons, let's not depend on that here.

ui/widgets/resizable.js
```diff
@@ -690,6 +684,21 @@ $.widget( "ui.resizable", $.ui.mouse, {
 		};
 	},
 
+	_calculateAdjustedElementDimensions: function( element ) {
+		if ( !( /content-box/ ).test( element.css( "box-sizing" ) ) ) {
```
mgol, line 688: No need for a regex test here; just compare the value directly.

ui/widgets/resizable.js
```diff
@@ -690,6 +684,21 @@ $.widget( "ui.resizable", $.ui.mouse, {
...
+			return {
+				height: parseFloat( element.css( "height" ) ),
+				width: parseFloat( element.css( "width" ) )
+			};
+		}
+
+		var outerDimensions = this._getPaddingPlusBorderDimensions( element );
+		return {
+			height: element[ 0 ].getBoundingClientRect().height - outerDimensions.height,
+			width: element[ 0 ].getBoundingClientRect().width - outerDimensions.width
```
mgol, line 698: `getBoundingClientRect` takes transforms into consideration while the CSS `width` & `height` - as well as jQuery APIs - do not. Using `getBoundingClientRect` is not valid. We should probably add a test case where the whole content is treated via a scale transform to make sure we didn't break it.

mgol (changes requested):

tests/unit/resizable/core.js
Daniel-Garmig: Ok, I'll change it!

Would it be better if I append a `<style>` tag using jQuery and remove it after the tests?
Something like this:
```javascript
var style = $( "<style> * { box-sizing: border-box; } </style>" ).appendTo( "head" );
//do test
style.remove();
``` 

One more thing. I noticed that other tests (not my new ones) do the same (see https://github.com/jquery/jquery-ui/blob/main/tests/unit/resizable/options.js#L558 ). Should I change those as well?

Thanks!

Daniel-Garmig:
@mgol I have changed the method to avoid using `getBoundingClientRect()`. 
Instead I get the value using `parseFloat( element.css( "width" ) );`, but without the scrollbars. So I disable scrollbars, get the correct computed size and set scrollbars back. (This seems kind of expensive and slow to me, but idk).

I tried to optimize it, so this calculation is only done when necessary. After a resize, width and height properties will be added as inline css to the element, so I will use those when available to avoid the expensive calculation.

I added some test for resizable elements with `transform` properties set. It doesn't matter after this changes but why not.

Let me know what do you think.

melloware (approved):

ui/widgets/resizable.js
```diff
@@ -690,6 +699,45 @@ $.widget( "ui.resizable", $.ui.mouse, {
...
+			width: elWidth
+		};
+	},
+
+	_getElementSizeWithoutOverflow: function( element, sizeProperty ) {
+		var overflowProperty = sizeProperty === "width" ? "overflow-y" : "overflow-x";
+
+		var origOverflow = element.css( overflowProperty );
+		element.css( overflowProperty, "hidden" );
+		var elSize = parseFloat( element.css( sizeProperty ) );
```
mgol, line 735: This will trigger an extra layout per function call which is undesirable. Since it gets called twice, it's two extra layouts, which is even worse.

There are other ways to get a scrollbar size, e.g. the hack that jQuery itself employs; see https://github.com/jquery/jquery/blob/dbc9dac7aecb106b66050342ff8daf1ecdd4239f/src/css.js#L99-L110

mgol (changes requested):

tests/unit/resizable/core.js
mgol: You can also change that test. If there are too many of them, though, you don't have to do this here.

tests/unit/resizable/core.js
```diff
@@ -244,4 +244,142 @@ QUnit.test( "nested resizable", function( assert ) {
 	outer.remove();
 } );
 
+QUnit.test( "Resizable with scrollbars and box-sizing: border-box", function( assert ) {
+	assert.expect( 4 );
+
+	var style = $( "<style> * { box-sizing: border-box; } </style>" ).appendTo( "head" );
```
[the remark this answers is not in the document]
Daniel-Garmig: Ok! I changed it.

There are some unit test cases on Resizable/options that appends elements to `body` and don't clear them after. See https://github.com/jquery/jquery-ui/blob/54f96eea31b21d9ecb00912261df3e5aaebf8cce/tests/unit/resizable/options.js#L15-L20
I can fix those so it appends elements to `#qunit-fixture`. Maybe on other PR?

ui/widgets/resizable.js
Daniel-Garmig: Thanks for the suggestion!

I have changed the calculation to use something similar to the "jQuery hack" you linked.
In this case, I used it to calculate the "theoretical" size of the element instead of scrollbar size delta.

tests/unit/resizable/core.js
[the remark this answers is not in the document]
mgol: Yes, feel free to submit another PR fixing this. No need to fix it here since those tests are unrelated.

tests/unit/resizable/core.js
```diff
@@ -244,4 +244,134 @@ QUnit.test( "nested resizable", function( assert ) {
 	outer.remove();
 } );
 
+QUnit.test( "Resizable with scrollbars and box-sizing: border-box", function( assert ) {
+	assert.expect( 4 );
+
+	$( "<style> * { box-sizing: border-box; } </style>" ).appendTo( "#qunit-fixture" );
+
+	//Both scrollbars
```
mgol, line 252: Please start comments with a space. Please fix other comments as well:
```suggestion
	// Both scrollbars
```

ui/widgets/resizable.js
```diff
@@ -381,20 +385,25 @@ $.widget( "ui.resizable", $.ui.mouse, {
 		this.offset = this.helper.offset();
 		this.position = { left: curleft, top: curtop };
 
+		var calculatedSize = undefined;
```
mgol, line 388: Please move the declaration to the single var we have above, adding it as a first entry in its own line, without the `undefined` initializer - that's our code style:
```js
		var calculatedSize,
			scroll = ( a && a === "left" ) ? "scrollLeft" : "scrollTop",
			has = false;
```

ui/widgets/resizable.js
```diff
@@ -690,6 +699,50 @@ $.widget( "ui.resizable", $.ui.mouse, {
...
+					height: parseFloat( element.css( "height" ) ),
+					width: parseFloat( element.css( "width" ) )
+				};
+		}
+
+		// Check if CSS inline styles are set and use those (usually from previous resizes)
+		var elWidth = parseFloat( ce.style.width );
+		var elHeight = parseFloat( ce.style.height );
+
+		var paddingBorder = this._getPaddingPlusBorderDimensions( element );
```
mgol, line 717: In this repo, we either use a single `var` (as I described in my comment above) or multiple ones in consecutive lines, not spread within the function. This is because the JS engine will hoist the declaration anyway and we want to make it explicit. You're free to assign the values here, but the declarations should be at the top.

Variables declared without initializers should be the first on the list, preferably all in one line if they fit.

ui/widgets/resizable.js
```diff
@@ -690,6 +699,50 @@ $.widget( "ui.resizable", $.ui.mouse, {
...
+	_getElementTheoreticalSize: function( element, extraSize, dimension ) {
+
+		// offsetWidth/offsetHeight is a rounded sum of content, padding, scroll gutter, and border
+		var size = Math.max( 0, Math.ceil(
+			element.get( 0 )[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ] -
+			extraSize[ dimension ] -
+			0.5
+
+		// If offsetWidth/offsetHeight is unknown, then we can't determine theoretical size.
+		// Use an explicit zero to avoid NaN (gh-3964)
```
mgol, line 740: `gh-3964` here is a reference to a jQuery issue, this is invalid in this repository. Please use the full link here:
```suggestion
		// Use an explicit zero to avoid NaN.
		// See https://github.com/jquery/jquery/issues/3964
```

tests/unit/resizable/core.js
```diff
@@ -244,4 +244,134 @@ QUnit.test( "nested resizable", function( assert ) {
 	outer.remove();
 } );
 
+QUnit.test( "Resizable with scrollbars and box-sizing: border-box", function( assert ) {
```
mgol, line 247: We have 4 tests here that are largely copy-pasted. I'd consider moving the test body to a generic function accepting the `borderBox` & `transforms` boolean parameters and then use them inside to apply the correct logic & messages.

See an example of a similar pattern in the jQuery repo:
https://github.com/jquery/jquery/blob/4b7ecbad24463c875f03ef4c7a7d307a091f93fd/test/unit/ajax.js#L1821-L1871

jQuery here uses an anonymous function inside the iterators to define the test body. I think I'd prefer to extract it into a parametrized function here to avoid the large indentation, but other than that it looks fine.

mgol (changes requested):
A few more comments.

tests/unit/resizable/core.js
```diff
@@ -246,41 +246,29 @@ QUnit.test( "nested resizable", function( assert ) {
...
-	elementContent.css( "width", "50px" );
-	testHelper.drag( handle, 10, 10 );
-	assert.equal( element.width(), 120, "element width (only vertical scrollbar)" );
-	assert.equal( element.height(), 120, "element height (only vertical scrollbar)" );
+	testResizableWithBoxSizing( assert, true, false );
 } );
 
 QUnit.test( "Resizable with scrollbars and box-sizing: content-box", function( assert ) {
 	assert.expect( 4 );
+	testResizableWithBoxSizing( assert, false, false );
```
mgol, line 254: See https://ariya.io/2011/08/hall-of-api-shame-boolean-trap; let's use named parameters for readability:
```suggestion
	testResizableWithBoxSizing( assert, {
		isBorderBox: false,
		applyScaleTransform: false
	} );
```
These are tests, we don't need to optimize for size or extreme performance.

tests/unit/resizable/core.js
```diff
@@ -246,41 +246,29 @@ QUnit.test( "nested resizable", function( assert ) {
...
-	var elementContent = $( "<div>" )
+QUnit.test( "Resizable with scrollbars, a transform and box-sizing: content-box", function( assert ) {
+	assert.expect( 4 );
+	testResizableWithBoxSizing( assert, false, true );
+} );
+
+function testResizableWithBoxSizing( assert, isBorderBox, applyScaleTransform ) {
+	var widthBefore, heightBefore,
+		cssBoxSizing = isBorderBox ? "border-box" : "content-box",
+		cssTrasform = applyScaleTransform ? "scale(1.5)" : "",
```
mgol, line 270: A typo:
```suggestion
		cssTransform = applyScaleTransform ? "scale(1.5)" : "",
```

mgol (changes requested):
Almost ready! Just two more comments from me.

Daniel-Garmig:
Changes done!
@mgol Thank you for all your suggestions and comments, they were great!

mgol (approved):
Thanks, looks good!

fnagel (approved):
+1 by reading and by running the unit tests (can confirm the new test will fail in Chrome if the fix is not present)

mgol:
Landed, thanks!

mgol:
jQuery UI 1.14.1 with this change has been released.

Issue #2278, opened 2024-08-08
1.14 needs `$.uiBackCompat = true;` to use `transfer` Effect


melloware:
To use the Transfer effect you have to have `$.uiBackCompat = true;` and not sure why.

You can see the error on the showcase: https://jqueryui.com/effect/

Pick Transfer and run.

![image](https://github.com/user-attachments/assets/600dcc8f-bda4-4ea7-b799-8dd84647c6b9)

mgol:
Thanks for the report. See https://api.jqueryui.com/transfer-effect/:
> This effect is deprecated, replaced by the [.transfer()](https://api.jqueryui.com/transfer/) method.

[Trac bug #14749](https://bugs.jqueryui.com/ticket/14749/) has more details:
> The transfer effect doesn't behave like a normal effect. As such, it should be converted a plugin instead of using the effects API.

We should just remove the effect from the demo, I think:
https://github.com/jquery/jquery-ui/blob/54f96eea31b21d9ecb00912261df3e5aaebf8cce/demos/effect/default.html#L78

And perhaps also its visual test:
https://github.com/jquery/jquery-ui/blob/54f96eea31b21d9ecb00912261df3e5aaebf8cce/tests/visual/effects/effects.js#L86-L92

melloware:
Thanks you i have updated to use `.transfer()` method!

mgol:
PR: #2291

PR #2291, opened 2024-09-09
Effect: Remove demos & visual tests of the transfer effect

mgol:
The transfer effect is deprecated and not present unless `$.uiBackCompat` is set to `true`. In previous UI versions, the default was including the backwards compat layer so these demos & tests worked; now they do not by default. Just remove them.

Fixes gh-2278


melloware (approved):

fnagel (approved):
+1 by reading

PR #2313, opened 2024-11-19
Widget: Increase performance for large sortable/draggable collections

mgol:
Fixes gh-2062

### DIFF
--- a/ui/widget.js
+++ b/ui/widget.js
@@ -301,10 +301,10 @@ $.Widget.prototype = {
 		this.uuid = widgetUuid++;
 		this.eventNamespace = "." + this.widgetName + this.uuid;
 
-		this.bindings = $();
+		this.bindings = new Set();
 		this.hoverable = $();
 		this.focusable = $();
-		this.classesElementLookup = {};
+		this.classesElementLookup = Object.create( null );
 
 		if ( element !== this ) {
 			$.data( element, this.widgetFullName, this );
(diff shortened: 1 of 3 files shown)


ui/widget.js
```diff
@@ -368,7 +368,7 @@ $.Widget.prototype = {
 			.removeAttr( "aria-disabled" );
 
 		// Clean up events and states
-		this.bindings.off( this.eventNamespace );
+		$( Array.from( this.bindings ) ).off( this.eventNamespace );
```
mgol, line 371: I'm not that worried about this since it's in `destroy`. That shouldn't happen too often.

ui/widget.js
```diff
@@ -450,16 +450,12 @@ $.Widget.prototype = {
...
 				continue;
 			}
 
-			// We are doing this to create a new jQuery object because the _removeClass() call
-			// on the next line is going to destroy the reference to the current elements being
-			// tracked. We need to save a copy of this collection so that we can add the new classes
-			// below.
-			elements = $( currentElements.get() );
-			this._removeClass( currentElements, classKey );
+			elements = $( Array.from( currentElements ) );
```
mgol, line 457: We re-wrap here, but the previous code was already creating a new jQuery object, so it shouldn't be worse than before.

Also, I don't understand the comment above; how would we lose the reference here? It's already in the `currentElement` object, that wouldn't just disappear...

melloware (approved):

PR #2320, opened 2024-12-18
Demos: fix easing demos

timmywil:
I stumbled across this during CSP testing. https://jqueryui.com/easing/

- Skip `_default`, which is a string and breaks the loop early
- Adjust text styling so the easing names fit.

## Old
image

## New
image


mgol (approved):
LGTM

PR #2324, opened 2025-01-08
Allow using a Function as the cancel option instead of a Selector. 

kghandi:
This is particularly useful when using Sortable on nested lists. 
I had a similar issue to that listed in issue 1996. Allowing cancel option to be a function allows me to override the default cancel implementation using a selector with something like:

$('.container', context).sortable({
    items: '>.widget',
    cancel: function (event) {
        return $(event.target).closest('.widget').is('.locked');
     }
});

### DIFF
--- a/ui/widgets/mouse.js
+++ b/ui/widgets/mouse.js
@@ -90,9 +90,11 @@ return $.widget( "ui.mouse", {
 
 		var that = this,
 			btnIsLeft = event.which === 1,
-			elIsCancel = typeof this.options.cancel === "string" ?
-				$( event.target ).closest( this.options.cancel ).length :
-				false;
+			elIsCancel = typeof this.options.cancel === "function" ?
+                                this.options.cancel.call(event.target, event) :
+                                ( typeof this.options.cancel === "string" ?
+				  $( event.target ).closest( this.options.cancel ).length :
+				  false);
 		if ( !btnIsLeft || elIsCancel || !this._mouseCapture( event ) ) {
 			return true;
 		}


mgol (changes requested):
Thanks for the contribution. This change makes sense to me, although before formally accepting I'd like @fnagel to also voice his opinion for possible wider context.

If Felix agrees this makes sense, there are still a number of changes we need before accepting this PR:
1. Make sure you follow our style guide: https://contribute.jquery.org/style-guide/js/. Currently, indentation is wrong, it uses spaces instead of tabs, spaces are missing inside of parens, etc.
2. Connected to the above - make sure our tests & linters succeed on the code before asking for another review. In this case, ESLint fails, for example. Run `npm install && npm run build && npm run lint && npm test` locally and see if everything works.
3. This change needs tests in all affected modules: draggable, sortable, etc. - perhaps whatever depends on `$.ui.mouse`.
4. This needs API updates so we'll need a PR to https://github.com/jquery/api.jqueryui.com/ as well before merging any new API.

fnagel:
Sounds reasonable to me and the changes are not that big, so I'm fine with integrating this when all requirements are met.

kghandi:
I have fixed the code style as requested. I have also created pull request for api documentation  https://github.com/jquery/api.jqueryui.com/pull/369

mgol (changes requested):
Thanks for the updates and the API PR. We still need to get this point addressed:

> This change needs tests in all affected modules: draggable, sortable, etc. - perhaps whatever depends on `$.ui.mouse`.

It's very important to have all new features covered by extensive tests.

PR #2325, opened 2025-01-14
Tests: Migrate test runner to jquery-test-runner

timmywil:
### Summary ###

I've moved all the code from `test/runner` into a package called [jquery-test-runner](https://github.com/jquery/jquery-test-runner). The changes to this repo are fairly simple. It uses the binary included in that package to run tests in browsers using selenium and browserstack.

Some things to note:

- I've tested jquery-test-runner with jquery, jquery-migrate, and jquery-ui and all their workflows have full green runs in my forks after migration.
- The `jtr` (alias for `jquery-test-runner`) binary includes several commands. The default command is for test runs (`jtr` or `jtr run`). jtr also includes `jtr serve`, which has a `-p` (`--port`) option (rather than an env var). See `npx jtr --help` for all commands and options.
- I've migrated yargs to commander in `jquery-test-runner`.
- I based the beginning infra of jquery-test-runner on the jquery repo's `test/runner` folder, which means the recent changes to allow Edge runs in IE-mode have been kept, `express` is not used for the test server, etc. jquery-migrate and jquery-ui will now be able to take advantage of those changes.
- `jquery-test-runner` can take a YAML config file, with support for all of the same options as the command line. I considered JSON and JS, but I liked how terse the YAML was. See [jtr-isolate.yml](https://github.com/jquery/jquery/blob/4618e612af11091df8879e6a980a93bc3edff15f/jtr-isolate.yml) for an example, which is used to create a separate run for each module in BrowserStack runs while still sharing browser instances.
- Upgraded the Node version used for testing to 22.x from 20.x in all workflows.

Ref https://github.com/jquery/jquery/pull/5604


mgol (approved):
LGTM

PR #2332, opened 2025-02-18
Build: Upgrade jtr to 0.2.5

timmywil:
The only other dependency that is upgradable is `grunt-eslint`, but it would require migrating to flat config.


mgol (approved):

mgol:
Landed, thanks.

BTW, @timmywil, our style guide mandates the PR title part after the `Component: ` intro should start with an uppercase letter.

PR #2335, opened 2025-03-02
Build: Switch from UglifyJS to SWC minify, make the minified file ES5

mgol:
More recent UglifyJS versions have started converting regular functions to arrow ones, making ES5 source file migrated to a ES2015+ minified one. We want to avoid that even in 1.14.x as long as we keep the source file in ES5.

Ref mishoo/UglifyJS#5967
Ref jquery/download.jqueryui.com#629


timmywil (approved):

fnagel (commented):
+1 by reading

PR #2336, opened 2025-03-02
Build: Update ESLint to v9, migrate to flat config, lint dist files

mgol:
Dist files linting is limited to checking if they're proper ES5.


mgol:
Note: this is ready to review but it will fail linting as long as #2335 doesn't land.

timmywil (approved):

fnagel (commented):
+1 by reading

PR #2337, opened 2025-03-14
Fixed datepicker positioning calculation

oojacoboo:
See #2057.  This still needs merging.

### DIFF
--- a/ui/widgets/datepicker.js
+++ b/ui/widgets/datepicker.js
@@ -816,7 +816,7 @@ $.extend( Datepicker.prototype, {
 		}
 		if ( !$.datepicker._pos ) { // position below input
 			$.datepicker._pos = $.datepicker._findPos( input );
-			$.datepicker._pos[ 1 ] += input.offsetHeight; // add the height
+			$.datepicker._pos[ 1 ] += inst.input.outerHeight(); // add the height
 		}
 
 		isFixed = false;


mgol (changes requested):
Thanks for the PR. We'll need tests for this change, though, as I indicated in the other PR. Have a look at existing test in https://github.com/jquery/jquery-ui/tree/main/tests/unit/datepicker to see how to write one.

oojacoboo:
@mgol these are unit tests.  Can you please explain what about this change needs to be added or updated with unit tests?  As for integration tests, which is actually the only way to add tests around this - does this lib even have those?  And if so, can you please direct me to one as an example where you can test the visibility of the datepicker's location within a browser window?

mgol:
We only have unit tests. While I agree some tests would better work as integration ones, most of the changes can be tested via unit tests.

I'd start with constructing a runnable example on a platform like JS Bin where the issue is obvious - this would be useful for people looking at the reasons for this patch in the future as well. Then we can think about a possible test case that would represent what we saw.

If you don't have any idea for how the unit test would look like, start with a test and maybe I'll be able to propose something.

oojacoboo:
@mgol How are you going to write a unit test for something that's dynamically calculated based on a browser window's scroll position?

mgol:
Tests are run on real browsers, the HTML files used for tests are also defined by us. You can scroll programmatically as long as you make sure there’s enough content to have something to scroll even on large screens.

mgol:
The existing spinner.html template probably doesn't have too much to scroll, but you can always dynamically attach a div with a large size to `#qunit-fixture` and then scroll.

If more invasive changes are needed, I added infrastructure for isolated iframe tests in https://github.com/jquery/jquery-ui/pull/2342, but it might not be necessary here.

oojacoboo:
Okay, I've added a test for the positioning with the more lazily calculated top position.  The issue occurs in more complex positioning scenarios.  I've added a test that creates a more complex positioning scenario to verify positioning.

tests/unit/datepicker/core.js
```diff
@@ -540,4 +540,83 @@ QUnit.test( "mouse", function( assert ) {
 		"Mouse click inline - next" );
 } );
 
+QUnit.test( "position", function( assert ) {
+	assert.expect( 4 );
+
+	var $container = $( "#qunit-fixture" ),
+		$modal = $container.children('div'),
```
mgol, line 547: That's a bit fragile to depend on it always being a single div. If you need a reference to this element, can you maybe start from the `inp` input and go to the parent?

tests/unit/datepicker/core.js
```diff
@@ -540,4 +540,83 @@ QUnit.test( "mouse", function( assert ) {
...
+	});
+
+	// Add an internal wrapper around all the children nodes
+	$modal.wrapInner(`<div id="relative-wrapper" style="position: relative"></div>`)
+	var $wrapper = $modal.children('#relative-wrapper');
+	$wrapper.prepend(`<div style="height: 300px;"></div>`);
+
+	// Wrap inputs with a container
+	$inputOne.wrap(`<div id="input-one-container" style="height: 200px;"></div>`);
+	$inputTwo.wrap(`<div id="input-two-container" style="height: 200px;"></div>`);
```
mgol, line 580: This is doing the wrapping after the widget is already initialized. I think we'd rather want to prepare the intended HTML structure before initializing the widget. If you modify the structure of an already running widget, various things may not work correctly and I don't think this issue requires any of that, does it?

Also, are all those modifications really needed? If it gets too heavy, it might be easier to write a separate iframe test where you can put all the style info in a style tag. But let's first verify why all those modifications are needed. Can you explain them all?

Plus, you're wrapping two different inputs with divs with the same ID; that's not compliant.

tests/unit/datepicker/core.js
```diff
@@ -540,4 +540,83 @@ QUnit.test( "mouse", function( assert ) {
...
+
+	assert.ok( $modal.is(":hidden"), "Modal is hidden" );
+
+	// Set the page scroll position way down the page
+	var done = assert.async();
+	$(document).ready(function() {
```
mgol, line 586: It shouldn't be needed to wrap any code in a test in document ready.

tests/unit/datepicker/core.js
```diff
@@ -540,4 +540,83 @@ QUnit.test( "mouse", function( assert ) {
...
+	$wrapper.prepend(`<div style="height: 300px;"></div>`);
+
+	// Wrap inputs with a container
+	$inputOne.wrap(`<div id="input-one-container" style="height: 200px;"></div>`);
+	$inputTwo.wrap(`<div id="input-two-container" style="height: 200px;"></div>`);
+
+	assert.ok( $modal.is(":hidden"), "Modal is hidden" );
+
+	// Set the page scroll position way down the page
+	var done = assert.async();
```
mgol, line 585: If the test is async, please put this declaration at the top of the test, as in other tests.

tests/unit/datepicker/core.js
```diff
@@ -540,4 +540,83 @@ QUnit.test( "mouse", function( assert ) {
...
+	$(document).ready(function() {
+		$(window).scrollTop(2000);
+
+		// Disable the scrollbar
+		$('body').css('overflow', 'hidden');
```
mgol, line 590: Is this necessary? Why? All such code modifying global elements needs to be reset manually at the end of the test or it may affect other ones.

tests/unit/datepicker/core.js
```diff
@@ -540,4 +540,83 @@ QUnit.test( "mouse", function( assert ) {
...
+
+		// Now show the wrapper with the input
+		$modal.show();
+
+		// Initialize datepickers
+		$modal.find('input').datepicker();
+
+		setTimeout(() => {
```
mgol, line 598: Why is the `setTimeout` needed, especially with such a long delay?

tests/unit/datepicker/core.js
```diff
@@ -540,4 +540,83 @@ QUnit.test( "mouse", function( assert ) {
...
+			$("#input-one-container").hide();
+			$("#input-two-container").css("padding-top", "100px");
+
+			var $inputTwo = $("#input-two-container").find("input");
+
+			$inputTwo.focus();
+			assert.ok( $("#ui-datepicker-div").is(":visible"), "Datepicker is visible" );
+			done();
```
mgol, line 606: Let's not call done before all the code is run, and we have some below.

mgol (changes requested):
Thanks for the updates. I added some comments. Please make sure lint & tests are passing for you locally before pushing changes. Right now, neither the lint nor the test you added is passing.

oojacoboo:
@mgol thanks for the feedback on this.  The reality here is that I am not familiar with qunit, or this lib's unique testing setup that comes with lots of pre-build assumptions.  Nor is there any documentation on testing (linting, etc).  The goal of the test is to create a scenario where the datepicker is initialized in a fixed position container, but with varying alterations to the DOM that affect it's calculated `top` value.

There appear to be a number of race conditions within tests as well, hence the timeout and `$(document).ready` calls.  Not to mention, this stock html template with default styling, etc.

I really don't have enough familiarity with all of this to generate a better test.  I can clean this one up.  But that's about it, and the time spent on this has already exceeded my goals by about 10x.

mgol:
I fully understand you may not have time to finish this PR. I don't have time either (see the project status at https://blog.jqueryui.com/2021/10/jquery-maintainers-update-and-transition-jquery-ui-as-part-of-overall-modernization-efforts/); beyond code reviews, most of my time on this project is spent on fixing regressions from 1.12.1 or newer; the rest is mostly left to the community.

This will just stay unfixed until someone has time to get this to the finish line. It's possible I'll have a closer look one day but I cannot promise this will happen.

juleslagarde:
This patch resolves an issue that only appeared on certain computers. It's strange that it's computer-dependent.

PR #2338, opened 2025-03-14
Spinner: Drop dependency on jQuery Mousewheel

mgol:
1. Listen to the native `wheel` event without depending on any wrapper plugins.
2. Keep listening to the `mousewheel` event for compatibility with projects using the jQuery Mousewheel plugin but route it to the `wheel` handler.


melloware (approved):

fnagel (approved):
+1 by reading

PR #2342, opened 2025-03-19
Spinner: Prevent double mousewheel & wheel event handling

mgol:
As of gh-2338, if one has loaded the jQuery MouseWheel plugin, the `mousewheel`
handler would fire the `wheel` one, but the `wheel` one would also run in
response to the native `wheel` event, resulting in double the distance handled
by the spinner. To prevent the issue, only fire the `wheel` handler from inside
the `mousewheel` on if the event was triggered by jQuery - jQuery will not care
that the underlying event is `wheel` and will only fire handlers for
`mousewheel`.

Also, add an iframe test using jQuery MouseWheel to not affect all the other
tests.

Ref gh-2338

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -247,6 +247,9 @@ grunt.initConfig( {
 
 				"requirejs/require.js": "requirejs/require.js",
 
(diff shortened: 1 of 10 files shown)


melloware (approved):

fnagel (approved):
Make sense and looks good by reading!

Issue #2344, opened 2025-03-22
v1.14.1: regression in tabs with non-standard IDs


tagliala:
Hello,

not sure that this is an issue with jQuery UI, but I think it is worth to report it

In Active Admin, we were trying to upgrade jQuery UI to 1.14.1 from 1.13.2, but one of our specs related to the tabs component was failing.

The failure is related to the way Active Admin creates the fragment name when it contains just an emoji, which is the same way GitHub creates fragments in markdown to HTML conversion

Example: `☺️` => `%EF%B8%8F`

### Reproducible test case:
Code taken from: https://jqueryui.com/tabs/
- ✅ v1.14.0 https://jsfiddle.net/tagliala/dkLa4v7e/8
- ❌ v1.14.1 https://jsfiddle.net/tagliala/dkLa4v7e/9

### Expected

<img width="525" alt="Image" src="https://github.com/user-attachments/assets/51233321-e9f5-4279-aa15-392d09d1b60f" />

### Actual

<img width="524" alt="Image" src="https://github.com/user-attachments/assets/f7e99a5b-c74f-4e35-93ff-01ad8554f085" />

Refs:
- https://github.com/activeadmin/activeadmin/pull/8670
- https://github.com/tagliala/rubyzip/wiki#%EF%B8%8F

mgol:
Thanks for the report and easy to run test cases!

This was caused by https://github.com/jquery/jquery-ui/pull/2307. I'll have a look at it for `1.14.2`.

mgol:
I added decoding in #2307 as it's needed for some complex characters like emojis to work - it's interesting that the linked issue claims it's emoji handling that's now broken... 🤔

mgol:
I wanted to see if we should or should not be decoding the anchor hash value in the end. I checked native browser behavior, without jQuery UI - one was testing whether `<a href="#🤗">` will scroll to `<div id="🤗">`, the second one was whether `<a href="#😅">` will scroll to `<div id="%F0%9F%98%85">`. Since `anchor.hash` is always encoded:
```js
anchor1.hash // '#%F0%9F%A4%97' for `<a href="#🤗">`
anchor2.hash // '#%F0%9F%98%85' for `<a href="#😅">`
```
we have the following:
1. If and only if we do decode, href from `<a href="#🤗">` will point to the id of `<div id="🤗">`
2. If and only if we do NOT decode, href from `<div id="🤗">` will point to the id of id="%F0%9F%98%85">`

[...]

Issue #2213, opened 2024-02-28
ui.tabs _isLocal fails if url contains username:password@


IAMJDA:
I had some really ugly problems with some jquery ui tab component and could pin it down to a wrong calculation in _isLocal.

decodeURIComponent(e) returns https://username:password@serverfqdn/path/file
decodeURIComponent(i) returns https://serverfqdn/path/file

These two are compared and the difference causes the return value to be false instead of true.
Therefore, the whole render logic breaks...

mgol:
Thanks for the report. Does the issue you describe exist when jQuery UI 1.12.1 is used or only with jQuery UI 1.13.0 or newer?

Also, please provide a runnable test case on a platform like JS Bin.

IAMJDA:
Sorry, I'm not a JavaScript dev. The function is depending on username + password being in the url, so a 1:1 example with any code plattform is unlikely to work.

The code in latest git seems to be the same. The culprit is that the anchor and location can have different values for the same URL.

mgol:
I understand but we cannot do much without a test case.

One thing you may be able to help with is saying whether jQuery UI 1.12.1 exhibits the same issue. You can find links to all releases on https://releases.jquery.com.

IAMJDA:
Please check https://jsbin.com/rinenupiba/edit?html,js,output

I faked _isLocal with user+password in location:
`			locationUrl = location.href.replace("://", "://test:test@").replace( rhash, "" );
`

Rendering of the page will change if you revert that line.

mgol:
It's not clear to me what the JS Bin is supposed to show. Can you use the stock version of jQuery UI without repeating its source in the JS pane and explain exactly what you expect to happen in this test case and what is happening instead?

mgol:
BTW, the `_isLocal` definition is identical between 1.12.1 & 1.13.2. Whatever the issue is, since it is already in 1.12, given limited team resources it's not likely to be fixed by the UI team; see the project status at https://blog.jqueryui.com/2021/10/jquery-maintainers-update-and-transition-jquery-ui-as-part-of-overall-modernization-efforts/. PRs are welcome if they're not too complex and contain tests.

IAMJDA:
> It's not clear to me what the JS Bin is supposed to show. Can you use the stock version of jQuery UI without repeating its source in the JS pane and explain exactly what you expect to happen in this test case and what is happening instead?

[...]

PR #2345, opened 2025-03-26
Tabs: Properly handle decoded/encoded anchor hashes & panel IDs, support URL-based credentials

mgol:
## 1\. Tabs: Support URL-based credentials

When credentials are provided directly in the URL, e.g.:

    https://username:password@www.example.com/

`location.href` strips out the auth part, but anchor links contain them, making
our `isLocal` computation broken. This fixes it by only looking at `origin`,
`pathname` & `search`.

Fixes gh-2213

## 2\. Tabs: Properly handle decoded/encoded anchor hashes & panel IDs

Prior to jQuery UI 1.14.1, hashes in anchor hrefs were used directly. In
gh-2307, that was changed - by decoding - to support more complex IDs, e.g.
containing emojis which are automatically encoded in `anchor.hash`.
Unfortunately, that broke cases where the panel ID is decoded as well.

It turns out the spec mandates checking both. In the "scrolling to a fragment"
section of the HTML spec[^1]. That uses a concept of document's indicated
part[^2]. Slightly below there's an algorithm to compute the indicated part[^3].
The interesting parts are steps 4 to 9:
4. Let potentialIndicatedElement be the result of finding a potential
   indicated element given document and fragment.
5. If potentialIndicatedElement is not null, then return
   potentialIndicatedElement.
6. Let fragmentBytes be the result of percent-decoding fragment.
7. Let decodedFragment be the result of running UTF-8 decode without BOM on
   fragmentBytes.
8. Set potentialIndicatedElement to the result of finding a potential indicated
   element given document and decodedFragment.
9. If potentialIndicatedElement is not null, then return
   potentialIndicatedElement.

First, in steps 4-5, the algorithm tries the hash as-is, without decoding. Then,
if one is not found, the same is attempted with a decoded hash.

This change replicates this logic by first trying the hash as-is and then
decoding it.

Fixes gh-2344
Ref gh-2307

[^1]: https://html.spec.whatwg.org/#scrolling-to-a-fragment
[^2]: https://html.spec.whatwg.org/#the-indicated-part-of-the-document
[^3]: https://html.spec.whatwg.org/#select-the-indicated-part

### DIFF
--- a/tests/unit/tabs/core.js
+++ b/tests/unit/tabs/core.js
@@ -747,4 +747,30 @@ QUnit.test( "extra listeners created when tabs are added/removed (trac-15136)", 
 		"No extra listeners after removing all the extra tabs" );
 } );
 
+QUnit.test( "URL-based auth with local tabs (gh-2213)", function( assert ) {
+	assert.expect( 1 );
+
+	var origAjax = $.ajax,
+		element = $( "#tabs1" ),
+		anchor = element.find( "a[href='#fragment-3']" ),
+		url = new URL( anchor.prop( "href" ) );
+
+	try {
+		$.ajax = function() {
+			throw new Error( "Unexpected AJAX call; all tabs are local!" );
+		};
+
+		anchor.attr( "href", url.protocol + "//username:password@" + url.host +
+			url.pathname + url.search + url.hash );
+
+		element.tabs();
+		anchor.trigger( "click" );
+
+		assert.strictEqual( element.tabs( "option", "active" ), 2,
+			"should set the active option" );
+	} finally {
+		$.ajax = origAjax;
+	}
+} );
+
 } );
--- a/tests/unit/tabs/core.js
+++ b/tests/unit/tabs/core.js
@@ -773,4 +773,55 @@ QUnit.test( "URL-based auth with local tabs (gh-2213)", function( assert ) {
 	}
 } );
 
+( function() {
+	function getVerifyTab( assert, element ) {
+		return function verifyTab( index ) {
+			assert.strictEqual(
+				element.tabs( "option", "active" ),
+				index,
+				"should set the active option to " + index );
+			assert.strictEqual(
+				element.find( "[role='tabpanel']:visible" ).text().trim(),
+				"Tab " + ( index + 1 ),
+				"should set the panel to 'Tab " + ( index + 1 ) + "'" );
+		};
+	}
+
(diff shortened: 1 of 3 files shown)


mgol:
## 1\. Tabs: Support URL-based credentials

When credentials are provided directly in the URL, e.g.:

    https://username:password@www.example.com/

`location.href` strips out the auth part, but anchor links contain them, making
our `isLocal` computation broken. This fixes it by only looking at `origin`,
`pathname` & `search`.

Fixes gh-2213

## 2\. Tabs: Properly handle decoded/encoded anchor hashes & panel IDs

Prior to jQuery UI 1.14.1, hashes in anchor hrefs were used directly. In
gh-2307, that was changed - by decoding - to support more complex IDs, e.g.
containing emojis which are automatically encoded in `anchor.hash`.
Unfortunately, that broke cases where the panel ID is decoded as well.

It turns out the spec mandates checking both. In the "scrolling to a fragment"
section of the HTML spec[^1]. That uses a concept of document's indicated
part[^2]. Slightly below there's an algorithm to compute the indicated part[^3].
The interesting parts are steps 4 to 9:
4. Let potentialIndicatedElement be the result of finding a potential
   indicated element given document and fragment.
5. If potentialIndicatedElement is not null, then return
   potentialIndicatedElement.
6. Let fragmentBytes be the result of percent-decoding fragment.
7. Let decodedFragment be the result of running UTF-8 decode without BOM on
   fragmentBytes.
8. Set potentialIndicatedElement to the result of finding a potential indicated
   element given document and decodedFragment.
9. If potentialIndicatedElement is not null, then return
   potentialIndicatedElement.

First, in steps 4-5, the algorithm tries the hash as-is, without decoding. Then,
if one is not found, the same is attempted with a decoded hash.

This change replicates this logic by first trying the hash as-is and then
decoding it.

Fixes gh-2344
Ref gh-2307

[^1]: https://html.spec.whatwg.org/#scrolling-to-a-fragment
[^2]: https://html.spec.whatwg.org/#the-indicated-part-of-the-document
[^3]: https://html.spec.whatwg.org/#select-the-indicated-part

fnagel (approved):
+1 by reading

Looks good to me, especially with the new tests and the manual testing in the original bug report (#2344).

mgol:
Landed in 53129e9cc7eb1c4f55b44a14adc91da23c7be85b & 89b0ecaaa3fc1f78e6b9f3d3b95de66f6bd22a2d.

Issue #2361, opened 2025-07-23
Slider: Add touch event support for better mobile interaction


nikunj8866:
Reference: https://core.trac.wordpress.org/ticket/63208

## Description
The slider currently lacks touch interaction support on mobile devices, making it difficult for users to navigate between steps using touch gestures. This limitation affects the feature's usability and accessibility on touchscreen devices.

## Screencast
https://screenrec.com/share/W1clQoqGSR

mgol:
Thanks for the report. We're avoiding adding new features to components at this stage of jQuery UI, but this one seems important and the patch is not that big, so I'll look into it.

PR #2362, opened 2025-07-23
Slider: Add touch event support for better mobile interaction

nikunj8866:
Issues: #2361


mgol:
Thanks for the PR. I'll look into it once we get jQuery 4.0.0-rc.1 out as that's our main focus at the moment.

hulyaafatih-rgb (approved):

nikunj8866:
Hi @mgol Thanks for the suggestion! The Pointer Events approach and the idea of a pointer.js module sound really good, but it's a bit out of scope for me. I'm not that experienced with this area. I opened PR mainly to address the specific issue that was fixed in WordPress.

I think this Pointer Events approach is a great idea for a dedicated implementation.

mgol:
@nikunj8866 Thanks for the response, I understand.

@joedolson you've asked what you can do to help moving this forward - that would be submitting a version of this PR with my feedback addressed. I can't say when I'd have time to look into it myself, our resources are limited here.

markvantilburg:
I also migrated Slider from Mouse to Pointer 
https://github.com/jquery/jquery-ui/pull/2373

Test fails since Pointer is not yet merged into main.

PR #2363, opened 2025-08-11
Build: Test on jQuery 4.0.0-rc.1

mgol:
When a stable 4.0.0 gets released, we'll remove the RC in favor of it.


fnagel (approved):
+1 by reading

timmywil (approved):

Issue #2354, opened 2025-06-10
The tabs feature requires nested interactive controls which are not allowed from an accessibility perspective


julie4754:
We run the AXE accessibility checker on our application.
This flags an issue on the elements we use  for the tabs feature.

See the following documentation on this issue.

https://dequeuniversity.com/rules/axe/4.10/nested-interactive?application=axeAPI

There needs to be a way to structure the elements for this tabs feature without nesting interactive controls.

Issue #2171, opened 2023-06-23
jquery-ui tab "Interactive controls must not be nested"


VishwaSangani:
[issue link](https://stackoverflow.com/questions/70857293/jquery-ui-tab-interactive-controls-must-not-be-nested)

The axe-core DevTools accessibility checker flags a "serious" problem with this code: "Interactive controls must not be nested", with the explanation "Interactive control elements must not have focusable descendants" and referencing guideline WCAG 4.1.2.

Even though I tried changing roles to 'none presentation' or totally removed the role but then it strats throwing issue for aria-selected and aria-expanded attributes as "Elements must used allowed aria attributes" 

How I do change the tabs widget code to make it accessible?

mgol:
Thanks for the report. Does the issue you describe exist when jQuery UI 1.12.1 is used or only with jQuery UI 1.13.0 or newer?

VishwaSangani:
I'm using jQuery UI 1.13.2 right now and facing this issue.

julie4754:
Accessibility is becoming a priority with many companies and organisations, is there any way to prioritize this?

mgol:
@julie4754 The best way to prioritize this is to submit a PR addressing this, with tests.

PR #2366, opened 2025-08-18
Fix(a11y/tabs) : move focusable element from <li> to <a>

XinaICBF:
issue: [2171](https://github.com/jquery/jquery-ui/issues/2171)
issue: [2354](https://github.com/jquery/jquery-ui/issues/2354)
This PR addresses accessibility issues in the Tabs widget.

- Moves focus from <li> elements to <a> elements.
- Resolves the "nested interactive controls" warning raised by accessibility testing tools.
- Aligns with [WAI-ARIA Authoring Practices for tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).

### DIFF
--- a/ui/widgets/tabs.js
+++ b/ui/widgets/tabs.js
@@ -176,8 +176,7 @@ $.widget( "ui.tabs", {
 	},
 
 	_tabKeydown: function( event ) {
-		var focusedAnchor = $( this.document[ 0 ].activeElement ).closest( "a" ),
-			focusedTab = focusedAnchor.closest( "li" ),
+		var focusedTab = $( this.document[ 0 ].activeElement ).closest( "li" ),
 			selectedIndex = this.tabs.index( focusedTab ),
 			goingForward = true;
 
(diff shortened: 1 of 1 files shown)


mgol (changes requested):
Thanks for the PR. However, the changes here are quite braking, as the `tabs` property is publicly available and I don't think we can change what it points to.

You can also see tests are failing.

Are you willing to address these issues?

XinaICBF:
Hi, the current jQuery UI Tabs structure uses <li role="tab"> with interactive elements like <a> inside, which triggers a "nested interactive controls" accessibility warning. A good example from[ Microsoft Teams docs](https://learn.microsoft.com/en-us/microsoftteams/platform/get-started/build-basic-tab-app?tabs=agentstoolkitcodespaces) shows an alternative: use <li role="presentation"> and move the interactive behavior and focus to the <a> element directly.

I’ll look into this approach over the weekend and see what can be done. Thanks!

PR #2369, opened 2025-09-12
Build: Commit package-lock.json, improve caching in CI

mgol:
Changes:
1. Commit package-lock.json.
2. Use cache settings built into `actions/setup-node`, drop `actions/cache` - it is simpler to maintain that way and it fixes Windows-specific issues.
3. Switch from `npm install` to `npm ci` in CI.
4. Update `actions/checkout` from v4.2.2 to v5.0.0.


fnagel (approved):
+1 by reading

PR #2371, opened 2025-09-24
Create pointer.js

markvantilburg:
A quick port of mouse.js to pointer events

### DIFF
--- a/ui/widgets/pointer.js
+++ b/ui/widgets/pointer.js
@@ -36,7 +36,7 @@ $( document ).on( "pointerup", function() {
 	pointerHandled = false;
 } );
 
-return $.widget( "ui.mouse", {
+return $.widget( "ui.pointer", {
 	version: "@VERSION",
 	options: {
 		cancel: "input, textarea, button, select, option",
--- a/ui/widgets/pointer.js
+++ b/ui/widgets/pointer.js
@@ -1,0 +1,209 @@
+/*!
+ * jQuery UI Pointer @VERSION
+ * https://jqueryui.com
+ *
+ * Copyright OpenJS Foundation and other contributors
+ * Released under the MIT license.
+ * https://jquery.org/license
+ */
+
+//>>label: Pointer
+//>>group: Widgets
+//>>description: Abstracts pointer-based interactions to assist in creating certain widgets.
+//>>docs: https://api.jqueryui.com/pointer/
+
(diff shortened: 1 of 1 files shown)


mgol (commented):
Thanks for the PR! Did you test the module with mouse & touch?

markvantilburg:
I tested it with a mouse (on desktop) and in chrome on my android, by replacing 'mouse' with 'pointer' with Slider and Sortable.

I tested with this file/js combination. More testing is needed for this feature.

[jquery-ui.min.1.14.1.js](https://github.com/user-attachments/files/22513914/jquery-ui.min.1.14.1.js)
[sort.html](https://github.com/user-attachments/files/22513915/sort.html)

markvantilburg:
@mgol is that safari code still needed?

mgol (changes requested):
I haven't tested this thoroughly, but the basic functionality works for the slider on both desktop & mobile with this PR + gh-2373; thanks for preparing these PRs!

That said, I think there are more changes to be made here:

1. Mouse & touch events have a different strategy for `*up` events - for mouse, they fire at the element above which the cursor was placed when it was relieved; for touch events, it fired at the element where the touch started. Pointer Events have a mechanism to opt into this touch behavior which is more convenient for drag operations; see MDN docs for [`setPointerCapture`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture) & [`releasePointerCapture`](https://developer.mozilla.org/en-US/docs/Web/API/Element/releasePointerCapture).
2. Pointer Events support multi-touch. Ideally, we'd allow to drag both slider handles for the range slider, but perhaps that'd be too big of a change. In that case, we should at least ensure the second & subsequent pointers are ignored.
3. We need to handle the [`pointercancel` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointercancel_event).

That's it for now. Would you like to continue the work on this PR?

markvantilburg:
Added pointercancel support, most of it should be implemented in the widgets

markvantilburg:
Only support the "first pointer" now
- Using isPrimary to Filter Pointer Events

markvantilburg:
@mgol would it be possible to create a complete build that uses this instead of mouse in all the widgets? Just case sensitive replace "mouse/Mouse" with "pointer/Pointer" so it can be tested on multiple devices.

PR #2379, opened 2026-01-15
Resizeable: avoid braking the size of an image

aleen42:
When an image has both `margin` and `padding`, there are two problems:

1. `margin` has set to the wrapper, but the image's one is not clear when initialising
2. While destroying the resizable widget, the image has resized while setting the width with `outerWidth` or the height with `outerHeight`.


mgol (changes requested):
Requires tests. Ideally, please also prepare a test case on JS Bin or similar.

Issue #2385, opened 2026-01-31
Datepicker created on focus does not open on initial focus with jQuery 4.0.0 (opens after blur/refocus)


tagliala:
Hello,

apologies if this issue does not belong here, I'm not a JavaScript developer

When creating a jQuery UI Datepicker lazily in a delegated focus handler (attach datepicker on focus the first time), the datepicker opens immediately on first focus with jQuery 3.7.1. With jQuery 4.0.0 the datepicker is created but does not open on that initial focus; it only opens after you blur and focus again.

Reproducible test case
- ❌  Not working (jQuery 4.0.0): https://jsfiddle.net/tagliala/vu1jg82p/3
- ✅  Working (jQuery 3.7.1): https://jsfiddle.net/tagliala/vu1jg82p/7/

### Minimal repro

HTML
```html
<input placeholder="To" value="" size="12" class="datepicker" maxlength="10" type="text" name="q[created_at_lteq]" id="q_created_at_lteq">
```

JS
```javascript
(($) => {

  $(document)
    .on('focus', 'input.datepicker:not(.hasDatepicker)', function() {
      const input = $(this);

      // Only create datepickers in compatible browsers
      if (input[0].type === 'date') { return; }

      const defaults = { dateFormat: 'yy-mm-dd' };
      const options  = input.data('datepicker-options');

      input.datepicker($.extend(defaults, options));
    });

})(jQuery);
```

### Steps to reproduce
1. Open the jQuery 4.0.0 fiddle: https://jsfiddle.net/tagliala/vu1jg82p/3
2. Click the input (initial focus)
3. Click outside (blur)
4. Click the input again

### Expected
- Datepicker should appear when the input is focused (on step 2) — every time the input is focused after creation.

### Actual
- On jQuery 4.0.0 the datepicker does NOT appear on the initial focus (step 2). It only appears after blur and refocus (step 4).
- On jQuery 3.7.1 the datepicker appears on the first focus as expected.

Can the jquery-ui team confirm whether this is:
  1) an intended behavior change in jQuery 4 that jquery-ui should adapt to, or
  2) a regression that jquery or jquery-ui should address so create-on-focus patterns still show the datepicker on the very first focus?

Thanks for any help you can provide

mgol:
Thanks for the report and a really good test case!

[...]

PR #2390, opened 2026-02-07
Datepicker: Support init inside a delegated focus handler in jQuery 4

mgol:
jQuery 4.0+ follows native focus events order, meaning that `focusin` is fired after `focus`. As delegated `focus` is implemented in jQuery via `focusin`, `focus` handlers attached during a delegated `focus` handler will not fire until the second time the field receives focus. This is what the `_attachments` method does. To account for that, show the datepicker if input is already focused. `_showDatepicker` checks if the datepicker is already open, so it's not a problem that it fires again as a `focus` handler in jQuery <4.

Note that the fact such an initialization worked inside of delegated focus handlers was a result of an implementation detail in jQuery. If a regular `focus` handler was used to initialize the datepicker, neither jQuery 4.0 nor 3.x would show the datepicker. This issue is now fixed as well.

Fixes gh-2385


fnagel (approved):
+1 by reading

PR #2391, opened 2026-02-09
Datepicker: Fix for years 0 - 999

### LINKED ISSUE #2388: Datepicker: Years 0000 - 0099 cannot be selected
Is there a chance to get any of thsese merged if I rebased it and did some cleanup? 
#1142, #1262, #1908 
https://bugs.jqueryui.com/ticket/7098/

What's the preferred way to configure this? A new config option, or maybe setting `shortYearCutoff: null`, `shortYearCutoff: 'none'` or some other special value?

MoonE:
Fixes #2388

This changes the `formatDate` function, when four-digit date format (`yy`) is used. A year in range 0-999 will be formatted as four digits with leading zeros. This fixes a bug where after selecting a year in that range, the datepicker reverted to the current date when opened again as `parseDate` expects exactly four digits.

And it introduces a `_createDate` function that constructs a Date object with the literal year if a four-digit year format is used.

Please have a look if the changes to Datepicker are acceptable. (Will try to add some more tests shortly).
Best to view each commit separately.

First commit changes the date formatting which fixes years 100 - 999.
Second commit introduces the `_createDate` and uses it where required, no functional changes.
Third commit adds the test for years 0 - 99. (I copied the test from #1262 with some modifications)
Fourth commit fixes Date creation with years 0 - 99.

Nothing changes with `y` format which still allows one or two digits and respects the `shortYearCutoff` config option.

Dates with year < 0 or >= 10,000 still won't work as of now. They can be selected but not parsed again from the input.

### DIFF
--- a/tests/unit/datepicker/options.js
+++ b/tests/unit/datepicker/options.js
@@ -1124,7 +1124,7 @@ QUnit.test( "Ticket #7244: date parser does not fail when too many numbers are p
 } );
 
 QUnit.test( "formatDate", function( assert ) {
-	assert.expect( 21 );
+	assert.expect( 24 );
 	testHelper.init( "#inp" );
 	var gmtDate, fr, settings;
 	assert.equal( $.datepicker.formatDate( "d m y", new Date( 2001, 2 - 1, 3 ) ),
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the PR. There are a lot of changes here, though, and while I understand they may be necessary to address the issue, I am not comfortable with merging such changes touching the widget logic at this stage of life of jQuery UI. The project is in maintenance mode and this issue has always been here, it'll likely just have to stay forever.

MoonE:
I hope you can reconsider. All changes done only apply to years that didn't work correctly before.

- f5ee24c35acb94ac17cf2a990f39169e8c53cba7 only changes output format for years 0-999 with format yy by padding it with zeros to four digits
- d795365d0a3c1319b1c551e6682927fe1d70f1cf is just adding a wrapper for the Date constructor where the year is a variable, quite a few lines, but very straight forward and no other change.
- 11430738ea9e9435841aa9f742b63a9e776ae8ee don't apply the short year offset if the date is parsed from a timestamp, or the year part is exactly four digits e. g. `0026`. Then if the year is in range 0-99 construct the date with a fixed leap-year `2000` then use `setFullYear` to set the actual year after.

PR #2394, opened 2026-03-11
Rename 'started' to '_mouseStarted' in mouse.js

markvantilburg:
It's used as : _mouseStarted in the rest of the class


lbesecker195 (approved):
Verified in `ui/widgets/mouse.js`: `this.started` is written once in `_mouseInit` and never read, while every other site reads and writes `_mouseStarted`. The init line was setting the wrong property. LGTM.

PR #2396, opened 2026-03-14
Fix year 100 - 999

MoonE:
This fixes part of #2388

### DIFF
--- a/tests/unit/datepicker/core.js
+++ b/tests/unit/datepicker/core.js
@@ -455,7 +455,7 @@ QUnit.test( "keystrokes", function( assert ) {
 } );
 
 QUnit.test( "mouse", function( assert ) {
-	assert.expect( 15 );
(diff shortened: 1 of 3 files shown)


mgol (commented):
Can you explain precisely what it fixes and what it does not?

MoonE:
Sorry for the delay, I seem to have missed your comment.

With `yy` format it expects exactly four digits for the year when parsing.
https://github.com/jquery/jquery-ui/blob/e803d4f67be9d7ae9a5a125c188dd003e3e3043a/ui/widgets/datepicker.js#L1237-L1240

Without this change selecting a date with year in range 100-999 works but is formatted as `999-01-01` which cannot be parsed when the datepicker is opened again.

Currently datepicker only works correctly for years 1000-9999.
This makes it work for years 0100-9999,
#2391 would make it work for years 0000-9999

PR #2416, opened 2026-09-04
Resizable: Fix resizing of border-box elements with a padding

### LINKED ISSUE #2382: border-box elements with a padding are wrongly resized, since 1.14.1
The width of the elements "jumps" by the value of the element's padding when `box-sizing` is set to `border-box`

Here is a jsfiddle showing the issue. Try resizing the element slowly and you'll notice a width/height jump.

https://jsfiddle.net/frgnt8ms/3/

The behavior was correct in version 1.14.0 (you can try changing the version in the fiddle).
The bug started happening in version 1.14.1.

charlesmass2:
On resize start, size & originalSize are computed from the computed CSS width & height. for border-box elements, that includes padding & border. Subtract padding & border when passing dimensions to width() and height()

Fixes https://github.com/jquery/jquery-ui/issues/2382


charlesmass2:
Hello @mgol @Daniel-Garmig just a quick ping if you find the time to look into this 🙂

